PR #2178, opened 2013-10-28
Added support for specifying packages to import in library.properties

dhowe:
Updated in response to pull-request #2172. I did notice the formatting issue you mention in an earlier commit, but thought I had fixed it (I don't see the problem when I diff) -- in any case, just let me know if its still there... 

-----------------------------

Added support for specifying packages to import in library.properties. Searches for 'imports' as a key in library.properties.

imports=mylib, mylib.util

If found, these imports are added when the user selects 'Import Library'.

import mylib.&#42;&#59;
import mylib.util.&#42;&#59;

If this line doesn't exist, falls back to current behavior and adds imports for all packages in the library's base jar.

This involved one substantive change: passing the name of the library to JavaEditor.handleImportLibrary(String) rather than the jar path. Once the library is found (by name), then its jar path is available via Library.getJarPath().


benfry:
Closing this one in favor of the other PR.

Issue #1939, opened 2013-07-04
Add "A file in sketch has changed, would you like to reload" pane in editor


whackashoe:
When working on a sketch I often open a file inside of it and edit it with an alternative editor. It would be nice if the editor could ask me if I'd like to update my sketch with the changes.

Possibly 3 answers: Update sketch, update file, No.

PR #2628, opened 2014-07-02
File changed detection, Issue #1939

pvrs12:
https://github.com/processing/processing/issues/1939

This monitors the open sketch for modifications to the file that it contains and offers the user the option to reload the sketch.

There are a few issues with this implementation, due to how java.nio works.
- The monitor detects, when files are modified, but not when they are deleted or created
- The monitor detects any files modified in the sketch directory, not just source
- The monitor can have a miss an update if a file was saved inside the editor and then a file was modified from outside the editor within 5 seconds (give or take)
- There is no way for the monitor to easily detect which file was modified, so the entire sketchbook is updated

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -2400,7 +2400,12 @@ public abstract class Editor extends JFrame implements RunnerListener {
 
         @Override
         public void windowLostFocus(WindowEvent arg0){
-          //do nothing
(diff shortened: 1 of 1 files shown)


Dealerpriest:
Was this merged into main? I can't seem to find any settings related to it when running processing. I'm running Processing 3.5.4

Issue #2986, opened 2014-11-25
Wrong width textbox in preferences.


gtitaev:
Menu File -> Preferences

Processing 3.0a4
![p3a4](https://cloud.githubusercontent.com/assets/6291442/5177509/7c98706e-7484-11e4-9f6c-2f9db920d0f0.png)

Processing 3.0a5
![p3a5](https://cloud.githubusercontent.com/assets/6291442/5177510/7c98f700-7484-11e4-93fc-9fd2fbf5302a.png)

benfry:
https://github.com/processing/processing/issues/67

Issue #67, opened 2013-02-10
redo the Preferences window using BoxLayout or GroupLayout


processing-bugs:
_Original author: b...@processing.org (June 07, 2010 00:15:49)_

This bug automatically added from:
http://dev.processing.org/bugs/show_bug.cgi?id=73

Comment from fry, 2005-07-15 16:20

redo panel to use proper Box layout etc
also needs to look good across all platforms

_Original issue: http://code.google.com/p/processing/issues/detail?id=28_

Akarshit:
![Preference Window](https://cloud.githubusercontent.com/assets/7762605/7116895/b5b04d94-e212-11e4-95f2-6386a2b27dfb.png)
@benfry how do you feel about this layout of Preference Window , this has been done using Group Layout , if you need any changes please comment i will include them and then issue the PR.   Meanwhile i am also changing layout of Find/Replace dialog to GroupLayout.

benfry:
Please keep the layout closer to the original. The 'Run sketches on display' line doesn't need a popup that extends the entire width of the window (neither do the others that take up the full horizontal space), and the three "More preferences" lines at the bottom should have less space between them.

Akarshit:
@benfry i have made the changes as you instructed, please look into the PR for more suggestions

benfry:
Incorporated for 3.0a6.

PR #2989, opened 2014-11-28
Fix Preference Window width for 3.0a5 #2986

Akarshit:


benfry:
Fix is outlined here: https://github.com/processing/processing/issues/67

Issue #2990, opened 2014-11-29
pressing PgDn in a code without scrollbar gives Exception


Akarshit:
In the code
void setup(){
}
pressing PgDn gives excetion

benfry:
Fixed for 3.0a6.

PR #2991, opened 2014-11-29
Fix for Exception when pressing PgDn in a code without scrollbar #2990

Akarshit:
Fix for Exception when pressing PgDn in a code without scrollbar #2990


benfry:
Think I found a simpler fix... the problem is just when `lineCount < visibleLines`, in which case Page Down should do nothing.

Issue #2998, opened 2014-12-04
Focus on window's buttons while closing


Akarshit:
The Save , Cancel , Don't Save Window's buttons are not binded to arrow keys.
The button could not be traversed using the arrow keys .

benfry:
On what platform? That dialog is handled by Java/Swing, so we're inheriting whatever behavior is implemented there.

Akarshit:
I tried it on linux and you are right that it is handled by java but there is a way to bind the arrow keys with the buttons of the dialog box as in #2999 .

benfry:
Let's keep things simple. The extra code, and debugging it across all our platforms, isn't worth it for the cool factor.

PR #2999, opened 2014-12-04
The arrow keys are binded to the buttons of  the Confirm Close dialog. #2998

Akarshit:
Fix for #2998

### DIFF
--- a/app/src/processing/app/Editor.java
+++ b/app/src/processing/app/Editor.java
@@ -2280,22 +2280,51 @@ public abstract class Editor extends JFrame implements RunnerListener {
     // With Java 7u40 on OS X, need to bring the window forward.
     toFront();
     
-    String prompt = "Save changes to " + sketch.getName() + "?  ";
+    
 
     if (!Base.isMacOS()) {
-      int result =
-        JOptionPane.showConfirmDialog(this, prompt, "Close",
-                                      JOptionPane.YES_NO_CANCEL_OPTION,
-                                      JOptionPane.QUESTION_MESSAGE);
+      Object[] options = new String[] {
+        Language.text("save.btn.save"), Language.text("save.btn.dont_save"), Language.text("prompt.cancel")
+      };
+      JOptionPane optionPane = new JOptionPane("<html> " +
+        "<head> <style type=\"text/css\">"+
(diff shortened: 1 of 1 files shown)


app/src/processing/app/Editor.java
```diff
@@ -2280,22 +2280,51 @@ protected boolean checkModified() {
...
+      dialog.setVisible(true);   
+
+
+      Object result = optionPane.getValue();                    
+        
+      if (result == options[0]) {
         return handleSave(true);
 
-      } else if (result == JOptionPane.NO_OPTION) {
+      } else if (result == options[1]) {
```
benfry: This is a very bad idea—the constants are used for a reason here, because placement is OS-specific. Why would we change this?

app/src/processing/app/Editor.java
```diff
@@ -2280,22 +2280,51 @@ protected boolean checkModified() {
...
+              JOptionPane.QUESTION_MESSAGE, 
+              JOptionPane.YES_NO_CANCEL_OPTION,
+              null,
+              options);
+              
 
-      if (result == JOptionPane.YES_OPTION) {
+      JDialog dialog = optionPane.createDialog("Confirm Dialog");
+
+      Set<AWTKeyStroke> forwardTraversalKeys =
```
benfry: If we use this approach, doesn't it mean implementing it in every single use of JOptionPane in the code? If so, it should be moved to a function/subclass that we can use everywhere, rather than repeating the two dozen lines of code in this patch all over the source.

But before we do that, why isn't this handled by Java, by default?

Akarshit: I am sorry, i should have added this with the original issue - 
Java does not proides the arrow bindings, it provides bindings with the `Tab` (which is by default)  
Though binding by `Tab` is enough i just thought it would be a "cool" feature as not many editors do that and i thought it can be useful.  
If you feel that this should be done i can extend JOptionPane and create a separate class for it.

app/src/processing/app/Editor.java
Akarshit: I will correct this once it's you agree to implement arrow key bindings.

PR #3025, opened 2014-12-31
greek translation and some fixes

mtsio:


benfry:
Thank you for the contribution! 

I need to figure out how we're going to handle fonts across languages, but I'd rather merge this now and sort it out later.

mtsio:
Actually it was one of my first contributions in github, so after much time had passed I thought that something went quite wrong :P

benfry:
Nah, it's just me having trouble finding free time to work on Processing. Thanks again.

Issue #3027, opened 2015-01-02
Use varargs in min() and max()


colouredmirrorball:
https://github.com/processing/processing/blob/master/core/src/processing/core/PApplet.java#L3724
I propose using eg max(int... a) in the method header instead of limiting to two or three arguments.

clankill3r:
It is not possible to have

```int max (int... a) {```` next to the current one cause it will make it ambiguous. 

I think replacing the existing one is a bad idea cause the current one is so simple which is fast.
Something like this could be done:

```
int max(int a, int b, int... c) {

  int max = max(a, b);

  if ( c != null) {
    for (int i = 0; i < c.length; i++) {
      max = (c[i] > max) ? c[i] : max;
    }
  }

  return max;
}
```

**Comment by colouredmirrorball** [comment]
Ah, yes, that is a much more elegant idea.

**Comment by datguy** [comment]
I was looking at the code tonight and found another solution:
```
  static public final int max(int a, int b, int... c) {
    int max = max(a, b);
    if ( c != null) {
      max = max(max,max(c));
    }
    return max;
  }
```
In addition to reusing the existing method to find the max value in the array, I love the line with 4 maxes.

colouredmirrorball:
Does that work? Because in general, c would be an array, and the current max method doesn't take arrays as input...

clankill3r:
No it won't work. And even if it did. I still think it's good to avoid the call to the max methodl to save overhead. Although I do not now when java compiles inline.

datguy:
It does work. There is a separate method that accepts an array, and max(c) makes use of that method.

datguy:
I tested the performance of a few variations (using min instead of max, but I expect max to be similar). Your implementation was slightly faster (5ms per 10M operations). Surprisingly, even though my implementation used the existing min(int[]) method, the existing method by itself was slower by 20ms per 10M.

GKFX:
Just out of curiosity, do any of these three changes to @clankill3r 's code help with speed?

````java
int max(int a, int b, int... c) {

  int max = a > b ? a : b; // manual max

  if (c != null) {
    for (int i : c) { //advanced for
      if (i > max) max = i; // don't run "max=max;" when i < max
    }
  }

  return max;
}
````

datguy:
[...]

PR #3033, opened 2015-01-08
Add min and max that use varargs

datguy:
Calculate min and max variables for multiple values without manually
creating an array. #3027


datguy:
Closed.  The improvements suggested by @GKFX in #3038 are faster.

PR #3038, opened 2015-01-10
Add varargs versions of PApplet.min/max

### LINKED ISSUE #3027: Use varargs in min() and max()

GKFX:
Calculate min and max variables for multiple values without manually
creating an array. Closes #3027.
Enhancement on #3033.

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -3749,7 +3749,10 @@ public class PApplet implements PConstants {
 
 
   /**
+   * Find the maximum value in an array.
+   * Throws an ArrayIndexOutOfBoundsException if the array is length 0.
    * @param list array of numbers to compare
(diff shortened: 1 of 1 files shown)


GKFX:
@benfry Although #3027 is locked, you didn't close it so it's now locked open. Please close it.
Closing this PR as it is no longer wanted.

benfry:
Yes, it's still open, because as I said, it's still under consideration for later in the 3.x process.

GKFX:
OK, sorry, I missed that bit.

benfry:
Closing this version because we'll either go all or nothing on the min/max replacement.

PR #3046, opened 2015-01-16
Open Recent files now dislays sketch after rename

### LINKED ISSUE #2984: Open Recent list doesn't get updated when renaming
If one renames the mother file (the pde named after the folder) using the Processing IDE, the project gets removed from the open recent list. I suppose this is a small problem with a small fix.

pjryan93:
Open Recent list is now updated when renaming within processing.  
Closes #2984.


benfry:
Thanks!

Issue #2852, opened 2014-09-10
"Your sketch has been modified externally" messages even w/ no changes


StanLepunK:
Hello, this popup appear often without reason, because there is no external modification.
I think it's when you work on few sketches, and you switch between them and you've write on it.
Some time there's a lot of windows popup open for the same sketch.
P 303 / OSX 10.9.4 / iMac 2012

![capture d ecran 2014-09-10 a 22 15 10](https://cloud.githubusercontent.com/assets/3332269/4224692/510f5cd4-3927-11e4-953c-8e0d4554e032.png)

benfry:
Think I have this fixed for the next release. For some reason, OS X is reporting the files as modified offset by exactly 1000 ms.

pvrs12:
I'm glad you were able to find that exceptionally strange bug. Sorry I wasn't able to keep with this without a mac

benfry:
Totally understand; appreciate your help in the first place.

distinctdan:
I'm seeing this bug on windows with version 3.0.1 on windows 7. When I run my sketch and then close it, processing constantly prompts me if I want to reload until I click "Yes". Clicking "No" only makes the popup immediately reappear. Also, after I click yes, one of my files (tabs) is gone. However, I can recover it by renaming the .tmp file for it. The bug is intermittent, but when it happens it's really annoying and kind of scary. I was afraid I had lost work the first time it happened.

distinctdan:
Update: I did just lose some work on a file. What I had after my last save is gone. This is unacceptable for an IDE. Is there some kind of config setting to turn this off?

ichbinscheisse:
Lost several files within one sketch after getting this message and "reloading" the sketch.
Any help if I mail the sketch to someone?

benfry:
There's a new issue here: https://github.com/processing/processing/issues/3965 that's probably what you're seeing. If not, please file a new bug. The problem reported here has been fixed (even though the symptoms may be similar.

Issue #2759, opened 2014-08-07
PDE erroneously detects changes in non-sketch files


kazimuth:
When exporting a sketch, new folders are created, but existing files aren't altered. The PDE pops up the "A file has been modified externally" message anyway.

This is also a problem in python mode, which compiles .py files into .class files in the sketch directory.

benfry:
The base issue here has been fixed. Anyone having problems with this, please file a new issue with details of what's going on.

PR #3048, opened 2015-01-18
File change detection rewrite

pvrs12:
I rewrote the original file change detection code that I had written so that it no longer uses the Java 7 FileWatcher tools. This should fix many of the issues that the old system had. 

The big changes between this and the previous version are
   * Change detection code has been moved from Editor.java to ChangeDetector.java
   * Detecting changes simply scans the directory for any files whose contents do not match the contents in memory or any files which have been created or deleted
   * The dialogs have been changed to be more clear, particularly emphasizing that reloading your sketch will clear any unsaved work
   * There is more support for choosing not to reload your files by marking the sketch files as modified so they will actually be written out when the sketch is saved

I have tested this on Archlinux, Windows 8, and OS X 10.10 (via a hackintosh VM) and I believe it should be working correctly now. Fixing #2852 and #2759 

The commit log for this is a huge mess because I still don't fully get git. If anyone would be willing to help me clean it up I would be grateful.


benfry:
Thanks for looking into it. This also needs to run on a thread (the file listing on window selection), otherwise anytime a user has an extremely full sketch folder (i.e. they've had `saveFrame()` running inside `draw()`), there will be significant lag/UI lockup when making the window active.

pvrs12:
Good point about threading, I'll take care of that. Is there a specific format that you like to use with processing for threaded tasks or just the standard java threading?

pvrs12:
This commit should switch it to using Java Threads to run the scan for changes

benfry:
Cool, thanks. Merged for the next alpha and we can try to shake out any bugs there. 

For the threading, just the regular Java threading is all we're looking for, though just mind any issues re: using the EDT. Update the GUI from the EDT using `EventQueue.invokeLater()` as necessary, since a new thread will be separate from the EDT.

pvrs12:
Do you want me to add the EventQueue.invokeLater() to the code? The only GUI elements to it are the dialogs, and some calls to header.rebuild()

benfry:
Yikes, this is totally broken on OS X. Even attempting a "save" brings up the "do you want to reload?" dialog before I even have a chance to save. Is this scanning for all files (not just `.java` and `.pde` or the valid extensions for the mode)? I wonder if OS X is writing/modifying the `.DS_Store` file.

benfry:
Yes, add `invokeLater()` as necessary. If you're not familiar with it, please read up on it first.

pvrs12:
I'm surprised that's happening, I checked it on an OS X VM and everything seemed good. 

It runs 2 scans, the first to see if there are any additions to the directory of files with the extensions specifed in the editor's current mode, and a second on the files and their contents that are actually in the sketch. 

I'll run through it some more on OS X to see if I can replicate your error. OS X seems to be the bane of my programming

benfry:
Ok, we may just have an issue with how the logic works on the change detection works in the current version. What should be happening is that the timestamp (`File.lastModified()`) of each file should be stored when it's loaded or saved in `SketchCode`. Then when the window gets focus, the current timestamp of the file should be compared to that. It's not perfect (if modified externally w/o activating/deactivating the frame, it won't pick it up), but it's far more reliable (and faster) than re-loading the contents of the file.

pvrs12:
That's a good point and a much better implementation that reading the whole file. Would it make sense to just grab the File.lastModified() time when focus is lost and compare it to when it is gained again?

Additionally, the current version scans the directory and compares the number of source files with the number of files in the sketch. This will cause it to also refresh if a new .pde file is added, or one is deleted. Does it make sense to keep this too?

benfry:
I think during load/save is best, because that's technically the contents of what the PDE knows to be the correct code. 

The edge case is a sketch that's open and running from a network-accessible location (a shared folder, or Dropbox, for that matter), in which case the file could be modified without the user de-focusing and re-focusing the window. For that case, we should probably do a check when saving, and if the stamp doesn't match, notify the user and rename the modified file to <filename>.conflict or something like that.

And yes, we'll want to keep the file scan. Thanks for your efforts on this.

PR #3050, opened 2015-01-18
Serial: Make it possible to open a port based on USB IDs

gohai:
The proposed format is (Vendor ID)#(Product ID or IDs)#(Serial), where everything after the initial vendor ID is optional. The necessary code to get to the USB descriptors is currently implemented only for OS X and Linux, Windows forthcoming.


benfry:
Couple thoughts: 

* If we add this, we should use the same syntax as the video library is using.

* I'm not nuts about the idea of having a bunch of constants defined that are tied to particular devices that have a shelf life of who knows what. It puts us in the position of deciding what goes in, and having the list be almost always out of date or at least a point of contention.

@shiffman @REAS thoughts?

gohai:
Thanks @benfry, looking into video library syntax tomorrow.

REAS:
I feel like I'm not knowledgable enough in this domain to have a strong opinion. I like the idea of making it easier to connect to devices, but I share @benfry's concerns about maintaining the list. I think it needs to be working with Windows when it's launched if it goes in.

gohai:
Agree wrt Windows. I'll need some more time to have this working, aiming for two weeks from now..

benfry:
Closing this in its current state. If we want to do something on these lines, we'll:

* Use the video syntax (as discussed above)
* Not include the constants... They belong in an Arduino library or something like that. I have no interest in making myself the target of the inevitable anger over what's included and what isn't.

PR #3051, opened 2015-01-20
Moved GUI operations to the EDT

pvrs12:
This PR moves all the GUI related operations to the main UI thread via EventQueue.invokeLater (most cases) and EventQueue.invokeAndWait (one case where the result from the user is needed to continue).


benfry:
Great, thanks for following up.

Issue #3052, opened 2015-01-20
Lists do not throw exceptions when non-existent elements are manipulated.


GKFX:
See #3044. The below code does not complain about increasing non-existent elements.
````java
FloatList list = new FloatList();
void setup(){
size(300,300);
list.add(0,000);
list.add(1,100);
list.add(2,200);
list.add(3,300);
}
void draw(){}
void mouseClicked(){
for(int i=0; i<3; i++){ println("list index "+i+" is "+list.get(i)); }
println("list length is "+list.size());
}
````

benfry:
Should be an easy fix for someone who wants to look at it (across IntList, FloatList, etc).

pvrs12:
I'd be willing to look at it. Is there a specific exception it should throw or should a new unchecked exception be made?

GKFX:
Follow http://docs.oracle.com/javase/6/docs/api/java/util/ArrayList.html#get(int) and throw an standard `IndexOutOfBoundsException`.

benfry:
Incorporated for 3.0a6.

PR #3053, opened 2015-01-21
Add exceptions for FloatList and IntList

pvrs12:
Added exceptions to the add, mult, sub, and div methods of FloatList and IntList to throw index out of bounds exceptions when they are accessed before creation per Issue #3052.


benfry:
Thanks for looking into it. This should be an `ArrayIndexOutOfBoundsException`, and also include a message explaining what's happened (the list size is x, trying to add to an element that doesn't exist). Also please keep the `if (index < count) {` style for spacing in line with the rest of the code.

pvrs12:
That should fix those problems.
The text on the error is 
> The list size is "+count+". Trying to "+method+" the element at "+index+" which does not exist.

And the user will see something like
> The list size is 3. Trying to add the element at 3 which does not exist.

Which I pulled out to a method that takes so it should be easy to change.

PR #3056, opened 2015-01-23
Formatted the Find/Replace dialog so that it displays the checkBox labels without ellipsizing.

Akarshit:
Formatted the Find/Replace dialog so that it displays the checkBox labels without ellipsizing.


.gitignore
```diff
@@ -2,4 +2,4 @@
 .AppleDouble
 ._*
 *~
-/build/shared/reference.zip
+/build/shared/reference.zip
```
omerjerk: This change is not required.

Akarshit:
@omerjerk thanks , i forgot about that.

benfry:
If we're going to change anything, this needs to be done with a layout manager. We can't add 10 here, subtract 20 there because while it may work on your system, it'll break (or look like hell) Linux or Mac OS X.

Issue #3057, opened 2015-01-24
Menu mnemonics (alt-f, etc) getting typed into the editor


Akarshit:
When pressing shotcuts for the menubar , the key also gets typed into the editor.
Steps to reproduce :  
1. Open processing
2. Press Alt + F
The file menu opens but "f" gets typed into the editor.

hamoid:
I don't remember ever having such shortcuts in Processing under Linux. Are those shortcuts OS dependent?
What I have seen lately is that sometimes I press CTRL+X and the console prints "cut".

Akarshit:
They were added recently and i also saw the that cut is getting printed in the console.  
But i am not able to build processing right now so i cant do anything about it.

GKFX:
@hamoid I have fixed the "cut" error. I haven't committed yet because the latest code won't build on my machine so I can't do good tests.

hamoid:
@GKFX great! I know there are many changes going on at the moment.
BTW. While fixing this, did you notice why it could happen that CTRL+C and CTRL+X suddenly would stop working? I can not reproduce it, but lately it has happened two or three times. I can solve it by closing and opening Processing. I suspected that the println("cut") was there to debug this issue I'm having at times.

GKFX:
No, I didn't. The `println("cut");` was added to debug moving from a cut ActionListener to a full CutAction, which allows menu dimming. When you say stop working, do you mean they went dim/inactive when they should have remained active? They are supposed to only gray out when there is no text selected.

hamoid:
It didn't occur to me to look at the menues to see if they were dimmed/inactive. I would just press CTRL+X for instance, but nothing would happen. The text would remain there, and the clipboard would keep it's old value. If it happens again I'll look at the menues.

hamoid:
I think I found it. I click on the Edit menu to open it, click again to close it without choosing anything. CTRL+X does not work. Click twice again on Edit, CTRL+X is back.

GKFX:
OK, I think I understand how to fix it. It sounds a little fiddly though so I'll wait until the cofde is more stable before I do any work.

benfry:
Fixed for 3.0a6.

PR #3058, opened 2015-01-24
MenuBar shortcut not being typed in the editor

Akarshit:
Fix for #3057 
In this i have ignored all the alt + characters.
Do let me know if this is the required behaviour. Or should i only ignore the keys which are shortcuts (eg. F,E,S,D etc. )

### DIFF
--- a/app/src/processing/app/syntax/DefaultInputHandler.java
+++ b/app/src/processing/app/syntax/DefaultInputHandler.java
@@ -249,6 +249,9 @@ public class DefaultInputHandler extends InputHandler
                 // http://code.google.com/p/processing/issues/detail?id=596
                 if ((modifiers & InputEvent.CTRL_MASK) != 0 && c == '/') return;
 
+                //Prevents the Menubar shortcuts from getting typed into the editor
+                if((modifiers & InputEvent.ALT_MASK) != 0) return;
+
                 if (c != KeyEvent.CHAR_UNDEFINED) // &&
                   //                (modifiers & KeyEvent.ALT_MASK) == 0)
                 {


benfry:
Thanks, though we need a better way to handle this—on OS X this would mean you can't type non-ASCII characters into the editor (alt is option on the Mac, which is how you get umlauts and that sort of thing), and on Windows, it would probably prevent Unicode alt-sequences as well.

Akarshit:
If you could recommend something i would be willing to implement it .   
One way is to only ignore the Alt + { 'F' , 'E' , 'S' , 'D' , 'T' , 'H' }  
This might prevent the problem on Windows also . If this looks promising to you then i will look further into this.

benfry:
I'd like to know the standard way this is handled in Java apps... It can't be unique to our project. I'm just not familiar with it myself, and we only just recently added support for the menu mnemonics.

GKFX:
Only ignoring a certain set of letters is a bad idea. It fails when you use eg. the German edition, because all the menu titles – and therefore all the shortcut keys – are different.
There are no menu mnemonics on OS X, so we can just not block any alt keys at all on Mac to solve that problem.

hamoid:
Are these any help?
http://www.codejava.net/java-se/swing/setting-shortcut-key-and-hotkey-for-menu-item-and-button-in-swing
http://docs.oracle.com/javase/tutorial/uiswing/components/menu.html

GKFX:
@hamoid Thanks for those links. I didn't see anything that might resolve this, but one thing worried me:
>If you use an Action for the menu item, the setMnemonic() doesn’t work. Instead, you must set mnemonic key via the Action.

On Ubuntu, this seems to be false. I could find no mention of it on the Javadoc either. However, if it is true, it seriously messes up the menu mnemonic system in a way that will be a pain to fix, so could I ask for the Windows users on here to check whether the menu mnemonics for `Edit`&rarr;`Cut` and `Edit`&rarr;`Copy` still work? (It's not a Mac issue.)

Akarshit:
I think the main problem is on Mac ( ability to use umlauts ) , can someone with mac tell how other java applications provide shortcuts.

hamoid:
What about these tips? They come directly from Apple:
https://developer.apple.com/library/mac/documentation/Java/Conceptual/Java14Development/07-NativePlatformIntegration/NativePlatformIntegration.html

GKFX:
@Akarshit
>Using mnemonics is discouraged in OS X, because mnemonics violate the principles of OS X Human Interface Guidelines.

I disabled menu mnemonics on OS X. We therefore do not need to block any keys at all on Mac. I've already said this!

Akarshit:
@GKFX Once @benfry agrees to this i would implement it.

GKFX:
@Akarshit OK, thanks. (Sorry for getting annoyed there; it was just a little frustrating  being ignored.)

benfry:
Folks, we're going in circles about incremental ways to avoid/hide the actual problem. We need to know why is this an issue for Processing, and not other Java apps. Processing isn't the only Java application that has a text area and menu mnemonics. To repeat from earlier: what's the *correct* way to handle this?

This discussion is also conflating menu *shortcuts* versus *mnemonics*. Mnemonics don't exist on OS X, but the solution is not to selectively avoid the check on OS X. I don't need links about how to implement menus. Processing has had menus and shortcuts for all 14 years of its existence.

As you can see in the code, there's a lot of spaghetti around the key checking. That's several years of duct tape. If we're going to bother with this, let's do it properly.

benfry:
Fixed separately for 3.0a6.

Issue #3068, opened 2015-01-29
Further build attempts fail when processing fails to start (when it shows one of the "We're off on the wrong foot" or "Terrible News" dialogs)


joelmoniz:
Once processing shows the "We're off on the wrong foot" or the "Terrible News" dialog, further ant builds fail with the following error:
  
```
BUILD FAILED
C:\Users\Joel\Documents\Code\p5\processing\build\build.xml:239: The following er
ror occurred while executing this line:
C:\Users\Joel\Documents\Code\p5\processing\build\build.xml:957: Unable to remove
 existing file C:\Users\Joel\Documents\Code\p5\processing\build\windows\work\jav
a\bin\awt.dll
```
  
The reason is most likely because pressing the "OK" button on the aforementioned dialog does not kill processing completely:  

![unkilled_p5](https://cloud.githubusercontent.com/assets/4526417/5957720/db5b680e-a7e2-11e4-97de-866697b4323f.png)

benfry:
https://github.com/processing/processing/issues/3063

joelmoniz:
Oops. This issue is different from the one mentioned, because it is not the fact that processing does not build that I meant to indicate, but the fact that when it doesn't build, rebuilding it requires one to go to the Task Manager (at least in Windows, not very sure about Linux/Mac OS) and manually kill the highlighted process there to build again.

benfry:
Sorry about the confusion. I appreciate the report and the fix.

Issue #3063, opened 2015-01-26
Build is currently broken on Ubuntu and Windows


GKFX:
After running `git checkout master; ant clean run` there are no error messages, but the splash screen appears for a moment and then nothing happens. I'm on Ubuntu, x64, Sun Java 1.8.0_25. Any thoughts? I've been compiling P5 for ages without this happening.

benfry:
Lots of things moving around at once. I've not had a chance to finish debugging on Linux and Windows.

GKFX:
@benfry I'm getting a `NullPointerException` [here](https://github.com/processing/processing/blob/master/java/src/processing/mode/java/JavaEditor.java#L2176)
````java
for (LineHighlight hl : breakpointedLines) {
````
Perhaps some null-checking would be good? If I add a check, it just reverts to the above behaviour, but at least it reduced the number of bugs.
![null pe](https://cloud.githubusercontent.com/assets/5357642/5943186/bf36fc8e-a714-11e4-8017-539b65defac2.png)

**Comment by benfry** [comment]
You can put a null check around that block to get it working. I've already done that in my version of the code. But mine is even more broken than what's on the head, so I've not pushed it yet.

**Comment by Akarshit** [comment]
I tried putting a null check around the statements , but still the error remains .  
But the strange thing is that even if i delete the 613 line in processing/app/src/processing/app/Base.java still the error is same , it's like the changes i am making are not being reflected.  
Can someone tell me what i am doing wrong.

**Comment by GKFX** [comment]
Normally `ant clean run` is the command I use to solve that error; failing that, I would delete the `build` directory entirely and check it out of git again. Also look into `git clean`. 

If that doesn't work, it's likely you're doing something daft like editing a different copy of the file to the one in your repository.

**Comment by pvrs12** [comment]
To get it to run for me I changed the code at line 1214 in Sketch.java to this

```
    try{
    editor.setCode(current);
//    editor.header.rebuild();
    editor.header.repaint();
    }catch(NullPointerException e){
      //e.printStackTrace();
    }
```
That seemed to get it to run for testing

benfry:
Fresh builds should be working again. Don't forget to `ant clean` before doing an `ant run`. 

For any build issues, please file a new issue.

PR #3069, opened 2015-01-29
Fatal errors now kill processing, enabling building it again without killing it manually without using the Task Manager

joelmoniz:
This commit resolves fatal errors, such as those indicated by the dialog box "We're off on the wrong foot", and not the usually non-fatal errors which sometimes might be fatal (such as those indicated by the dialog box "Terrible News", which though often non-fatal, might sometimes be fatal, as in #3063), thereby partly fixing the issue described in #3068.


benfry:
Ah, good catch—thank you.

PR #3070, opened 2015-01-30
File change detection v3

pvrs12:
This attempt at file change detection utilizes the time at which the files were modified through Processing's save/load interface vs the time the files were modifed on disk. A field was added to SketchCode.java to track this (lastModified) and it is updated when a sketch is saved, loaded, or saved as.

I had added some simple exception handling to Sketch.java to allow testing, but removed it for this push. If you want I can re-add that to allow others to successfully run Processing.
This is the simple change I made at line 1214
```
    try{
    editor.setCode(current);
//    editor.header.rebuild();
    editor.header.repaint();
    }catch(NullPointerException e){
      //e.printStackTrace();
    }
```
Hopefully this fixes the problems that Mac OS X was having with the spurious reloads and should run faster than reading through the content of each file individually.


benfry:
Giving it a shot, thanks for your continued work on this one.

shiffman:
FYI in current build (233) I'm getting the "Your sketch has been modified externally" messages continuously even with no changes again.  I noticed there's still an open issue #2852 so am not filing a new one.

benfry:
Yup, same here.

Issue #2843, opened 2014-09-07
Old version number shown in Mode Manager dialog.


jdf:
The Mode Manager, as of this writing, shows version 0368, but Python Mode is up to 0373. The 0368 version was replaced by 0369 nine days ago.

![screen shot 2014-09-07 at 2 31 16 pm](https://cloud.githubusercontent.com/assets/55948/4179234/49346db8-36bd-11e4-93e5-a92278bc0680.png)

![screen shot 2014-09-07 at 2 32 56 pm](https://cloud.githubusercontent.com/assets/55948/4179240/71aacc60-36bd-11e4-9af8-6cb43d1bba7e.png)

Akarshit:
@alingedleft From the comments it looks that this issue has been fixed. Can someone close this.

PR #3073, opened 2015-01-30
contributions.txt now gets deleted and recreated instead of overwritten

### LINKED ISSUE #2994: contributions.txt does not always get updated to the most recent one, even after the most recent one downloads
The newly downloaded contributions.txt does not always replace the old one. Thus, sometimes, older versions of contributions may be displayed (as was happening in [this](https://github.com/processing/processing/issues/2843) issue). This occurs even though the latest contributions.txt has downloaded successfully.

NB: This issue is not consistent, as usually, the newly downloaded contributions.txt replaces the old one.

joelmoniz:
The contributions listing is now downloaded into a temporary file first. Once the listing is downloaded, the old ```contributions.txt``` file is deleted, and the temporary file is then renamed to ```contributions.txt```. This is required because the local copy of the ```contributions.txt``` (strangely enough) does not get updated sometimes (although this error is not easily reproducible by locally modifying the file, except when the file is from the processing website), as found [here](https://github.com/processing/processing/issues/2843). This PR fixes #2994.

### DIFF
--- a/app/src/processing/app/contrib/ContributionListing.java
+++ b/app/src/processing/app/contrib/ContributionListing.java
@@ -23,6 +23,7 @@ package processing.app.contrib;
 
 import java.io.*;
 import java.net.*;
+import java.nio.file.Files;
 import java.util.*;
 import java.util.concurrent.locks.ReentrantLock;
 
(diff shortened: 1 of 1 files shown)


benfry:
Thanks, Joel. Does this also fix https://github.com/processing/processing/issues/2843?

joelmoniz:
This solves the bug highlighted in #2843 about the local copy of contributions.txt not getting updated, but I'm afraid I'm not quite sure about whether or not to mark the issue as resolved, since the issue also mentions the proposal of the contributions.txt automatically getting updated nightly.

benfry:
Got it, thanks.

Issue #3074, opened 2015-01-31
Opening and closing preferences window prompts user to save unmodified sketch


joelmoniz:
Steps to reproduce this error:  
  
1. Open a new sketch, or an already existing one 
2.  Open the preferences window  
3. Close the preferences window  
4. Close the sketch  
  
A dialog box prompting the user to save the current sketch is displayed (even though the sketch has not been modified).

benfry:
This might be https://github.com/processing/processing/issues/3057

Did you open preferences using a keyboard shortcut?

joelmoniz:
Oh yes, I did use a keyboard shortcut. However, nothing gets typed into the Editor when I use this shortcut, unlike in #3057. Further, this problem doesn't seem to occur when I use other key combinations.

benfry:
Though I suspect hitting alt or ctrl may mark the sketch as modified (on Windows), are you seeing that?

benfry:
https://github.com/processing/processing/issues/3057

benfry:
Fixed for 3.0a6.

joelmoniz:
Thank you :)

PR #3076, opened 2015-02-01
Sketch doesn't get modified when preference window is opened through sho...

Akarshit:
Sketch doesn't get modified when preference window is opened through shortcut. #3074


benfry:
Now fixed separately.

Issue #3028, opened 2015-01-03
textAlign(RIGHT) including spaces at the end of each line


lilalinda:
When using the example code below -- inspired by the `text()` API entry -- the rendering shows a slight offset from the right.  Without digging deeper into the code, I would assume these to be the white spaces where the lines are allowed to wrap.  When setting alignment to `RIGHT`, in my opinion, the spaces should move to the following line or swallowed entirely to allow for perfect alignment.

```
void setup() {
  size(100,100);
  smooth();
  noLoop();
}

void draw() {
  String s = "The quick brown fox jumped over the lazy dog. X";
  fill(0);
  textAlign(RIGHT);
  text(s, 10, 10, 70, 80);  // Text wraps within text box
}
```
The orange line shows the unwanted space, which should align better with the last character X:

![sketch_150103a](https://cloud.githubusercontent.com/assets/1231235/5603725/4dffc02e-9359-11e4-899a-730461342124.png)

spasmsaps:
IMHO, it would be nice to have control in how text is wrapped. textWrapMode(int MODE)
http://www.w3.org/TR/css3-text/#white-space-property

spasmsaps:
related #478

benfry:
@lilalinda Huh, I can't imagine how that's never been seen/caught before.

Akarshit:
I was looking into this issue but i am facing some problems , i asked it on the forum but haven't received any response yet . I am really stuck and if someone would be kind enough to take a look at   http://forum.processing.org/two/discussion/9245/how-to-make-changes-to-papplet-class-in-processing-core-papplet-java#latest  
I would be really grateful.
I know this might not be the right place to reference the forum discussion , but i am facing similar problems in changing other files .

benfry:
@Akarshit added notes to that forum post, thanks for looking into it.

Akarshit:
@benfry I tested few texts with textAlign(LEFT) and they were working fine.

Could you paste here a string which gives incorrect output.

benfry:
Fixed since 3.0a6.

PR #3078, opened 2015-02-02
textAlign Right print correctly

Akarshit:
Fix for #3028   
The spaces are not carried to the next line instead they are ommited .  
I am not sure if this is the required behaviour. Should the spaces before the first word in the line be printed ?


benfry:
This would fix the `textAlign(RIGHT)` case, but break `textAlign(LEFT)`.

Akarshit:
@benfry I tested few texts with textAlign(LEFT) and they were working fine.  
Could you paste here a string which gives incorrect output.

benfry:
We'll give it a try.

Issue #2944, opened 2014-11-14
sketchbook window doesn't update when sketches are added, renamed, etc


benfry:
In the current source, new sketch names are not added to the sketchbook and they're not renamed after changes are made. This is an error (and a regression from earlier releases.)

Separately, there have been requests to reload the menu when the sketchbook folder is modified externally (i.e. from the Finder or Windows Explorer). That's a separate feature request: https://github.com/processing/processing/issues/1831

Akarshit:
If @JayDusara is not working on improving  #3178 . I would like to work on the suggestion given there and fix issues related to the updation of SketchBook

PR #3081, opened 2015-02-03
Sketchbook updating when renaming, saving and deleting from within Processing. 

Akarshit:
Fix for #2944


benfry:
I think this is covered by the fixes I made in https://github.com/processing/processing/commit/03a5c78ca799c67da5829a524529e0e1041a7350

Issue #1046, opened 2013-02-10
exported Linux sketches must be run from the sketch folder


processing-bugs:
_Original author: jwmatt...@gmail.com (March 07, 2012 08:34:54)_

In Processing 2.0a4, Ubuntu 11.10:

Sketches with loadFont() export incorrectly. If the application is launched from outside the application.linux folder, a NullPointerException is thrown.

1. Create a simple sketch with a vlw file in data folder:
    //start sketch
    PFont font;
    font = loadFont(&quot;FreeSans-24.vlw&quot;);
    text(&quot;test&quot;,50,50);
    //end sketch

2. Export application
3. Open terminal, try:
     cd ~; ./export_test/application.linux/export_test

java.lang.NullPointerException
	at java.io.DataInputStream.readInt(DataInputStream.java:370)
	at processing.core.PFont.&lt;init&gt;(Unknown Source)
	at processing.core.PApplet.loadFont(Unknown Source)
	at export_test.setup(export_test.java:20)
	at processing.core.PApplet.handleDraw(Unknown Source)
	at processing.core.PApplet.run(Unknown Source)
	at java.lang.Thread.run(Thread.java:662)
Exception in thread &quot;Animation Thread&quot; java.lang.RuntimeException: Could not load font FreeSans-24.vlw. Make sure that the font has been copied to the data folder of your sketch.
	at processing.core.PApplet.die(Unknown Source)
	at processing.core.PApplet.die(Unknown Source)
	at processing.core.PApplet.loadFont(Unknown Source)
	at export_test.setup(export_test.java:20)
	at processing.core.PApplet.handleDraw(Unknown Source)
	at processing.core.PApplet.run(Unknown Source)
	at java.lang.Thread.run(Thread.java:662)

NOTES: No error occurs when launched from within the application.linux folder.
File is confirmed to be in data folder.

The bug fix listed here does not fix the problem: http://code.google.com/p/processing/issues/detail?id=945&amp;q=application%20export&amp;colspec=Stars%20ID%20Type%20Status%20Priority%20Owner%20Summary

_Original issue: http://code.google.com/p/processing/issues/detail?id=1008_

benfry:
Fixed via #3083

PR #3083, opened 2015-02-04
Allow exported linux applications to be launched from a different folder

kfeuz:
This addresses Issue #1046 which deals with launching exported linux applications from a different folder.  The problem was that the $APPDIR folder was not added to the classpath so it was not finding the data folder when the script was launched from a different location.


benfry:
So awesome to have this fixed. Thanks for looking into it and submitting the PR.

Issue #1400, opened 2013-02-10
Recreating Client instance will cause an out-of-memory error


processing-bugs:
_Original author: hya...@gmail.com (October 31, 2012 08:32:17)_

<b>If possible, please attach the smallest possible sketch that reproduces</b>
your problem. First make a tiny example, then use Tools -&gt; Archive Sketch
<b>to create a zip file, and then attach it to this bug report.</b>

<b>What steps will reproduce the problem?</b>
1. run the sketch ClientServerTest
2. open process monitor application such as Activity Monitor (MacO X) to check the real memory size continuously increasing
3. wait for a few minutes; application will stop with out-of-memory error

<b>What is the expected output? What do you see instead?</b>

no real memory size continuous increase, no out-of-memory error.

<b>What version of the product are you using? On what operating system?</b>

processing 1.5.1, 2.0b5 on MacOS X

<b>Please provide any additional information below.</b>

dispose() method defined in Client.java seems not unregistering the &quot;dispose&quot; this object.
Adding the following fragments of code at the beginning of the dispose() would fix this issue.
N.B. 1.5.1 PApplet does not have unregisterMethod(), so the following code is for 2.0+ only.

        // If this Client object was created with the Client(PApplet parent, String host, int port) constructor,
        // unregister this from the parent.
	if (host != null)
            parent.unregisterMethod(&quot;dispose&quot;, this);

<b>Note: Please do not link to locations on other sites, such as your own</b>
<b>image site, video site, blog, etc. Include the relevant information here</b>
<b>with the report.</b>

_Original issue: http://code.google.com/p/processing/issues/detail?id=1362_

benfry:
Fixed by #3088

PR #3088, opened 2015-02-05
Prevent memory leak of Client by unregistering parent dispose method

kfeuz:
This addresses issue #1400 dealing with a memory leak in the processing.net.Client which occurs when the parent "dispose" method is not unregistered if the client is stopped.  Rather than trying to track the state of several different variables, I've just declared an explicit boolean variable to track whether or not the dispose method needs to be unregistered.  Now when spawning and closing Clients the memory usage stays consistent.


benfry:
Thanks for the fix.

Issue #3089, opened 2015-02-05
processing.net.Server only cleans up internal Clients when trying to write


kfeuz:
This can lead to memory issues if the server only reads data and never/seldom attempts to write to the clients.  I see a few ways to attempt to resolve this.  We could check for dead clients in the available() method or the run() loop.  The other approach would be to add some sort of callback to the Server when the internal client dies.  Is there a preference for one technique over the other?  I would be happy to submit a PR once this is decided.

gohai:
My first time looking really at the network code, but I see the issue you describe.. I'd suggest adding the check to available(), since this should work for all practical cases, and we're already iterating over the array there.

kfeuz:
In looking at the issue a little closer, checking for dead clients in available does seem like the best choice.  It allows us to have a chance to retrieve the unread data from dead clients before they get discarded.  I have submitted a pull request.

benfry:
Fix incorporated for 3.0a6.

PR #3097, opened 2015-02-06
Server check for dead clients while checking for available messages

kfeuz:
As discussed in issue #3089 the processing.net.Server only cleans up dead clients when trying to write to them.  This patch has the server also check for dead clients while it is looking for available data to read.  The check is done in such a way so that a client that has died with unread data can still be read.  Additionally, the order of the clients should remain consistent and no client should be skipped.  I have included comments explaining the code so that in the future the non-obvious statements can be easily understood.


benfry:
Thanks for looking into it.

Issue #2530, opened 2014-05-20
IllegalArgumentException thrown when clicking in a second sketch window


joelmoniz:
The following stack trace is printed when one switches from an open sketch to another:

```
Exception in thread "AWT-EventQueue-0" java.lang.IllegalArgumentException: Bounds out of range: 9,9 [0]
	at processing.app.syntax.JEditTextArea.select(JEditTextArea.java:1227)
	at processing.app.syntax.JEditTextArea.setCaretPosition(JEditTextArea.java:1181)
	at processing.app.syntax.JEditTextArea$MouseHandler.doSingleClick(JEditTextArea.java:2456)
	at processing.app.syntax.JEditTextArea$MouseHandler.mousePressed(JEditTextArea.java:2415)
	at java.awt.AWTEventMulticaster.mousePressed(Unknown Source)
	at java.awt.Component.processMouseEvent(Unknown Source)
	at javax.swing.JComponent.processMouseEvent(Unknown Source)
	at java.awt.Component.processEvent(Unknown Source)
	at java.awt.Container.processEvent(Unknown Source)
	at java.awt.Component.dispatchEventImpl(Unknown Source)
	at java.awt.Container.dispatchEventImpl(Unknown Source)
	at java.awt.Component.dispatchEvent(Unknown Source)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Unknown Source)
	at java.awt.LightweightDispatcher.processMouseEvent(Unknown Source)
	at java.awt.LightweightDispatcher.dispatchEvent(Unknown Source)
	at java.awt.Container.dispatchEventImpl(Unknown Source)
	at java.awt.Window.dispatchEventImpl(Unknown Source)
	at java.awt.Component.dispatchEvent(Unknown Source)
	at java.awt.EventQueue.dispatchEventImpl(Unknown Source)
	at java.awt.EventQueue.access$200(Unknown Source)
	at java.awt.EventQueue$3.run(Unknown Source)
	at java.awt.EventQueue$3.run(Unknown Source)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$1.doIntersectionPrivilege(Unknown Source)
	at java.security.ProtectionDomain$1.doIntersectionPrivilege(Unknown Source)
	at java.awt.EventQueue$4.run(Unknown Source)
	at java.awt.EventQueue$4.run(Unknown Source)
	at java.security.AccessController.doPrivileged(Native Method)
[...]

PR #3101, opened 2015-02-07
IllegalArgumentException not getting thrown when clicking in a second sketch window

Akarshit:
Fix for #2530


benfry:
Awesome, thank you!

PR #3131, opened 2015-02-26
Formatted the Find/Replace dialog so that it displays the checkBox labels without ellipsizing.

Akarshit:
Sorry for #3056 
Realized that this might be the more approprite way to do this.


Akarshit:
Better alternative would be to use GroupLayout as in #3190

Issue #2953, opened 2014-11-15
Ready to add contributed example packages?


shiffman:
(moving this discussion from e-mail to github).

@prisonerjohn and @joelmoniz we're ready to start adding example packages to the contributions manager.  Looks like this was implemented with https://github.com/processing/processing/pull/2795 and we just need now for an example category in contributions.txt.  Here is my first attempt at making an examples package:

http://shiffman.net/p5/example-packages/natureofcode.txt
http://shiffman.net/p5/example-packages/natureofcode.zip

Also, the window title for the examples should say: "Examples Manager" rather than "Examples-Package Manager".

![screen shot 2014-11-15 at 11 53 02 am](https://cloud.githubusercontent.com/assets/191758/5058810/0b2beb86-6cbe-11e4-8c13-a23d35184e3c.png)

shiffman:
What's the status of this?  I'm ready to work on some example packages as well as put together some documentation (see: https://github.com/processing/processing-docs/issues/173).

joelmoniz:
Examples work well now. But it seems the [contribs.txt](http://download.processing.org/contribs.txt) doesn't contain The Nature of Code examples package any more, although things worked perfectly when it was there some time ago.

shiffman:
I think the issue was that if we add it to `contribs.txt` then previous alphas break.  My question would be does this include a5.  If it breaks a5 then let's wait until a6 is released and add it in.  @prisonerjohn can you shed any light?   Let's make sure we double-check that it works before the next release as it would be great to start getting these packages in there.

prisonerjohn:
Yes, adding any example packages would break previous versions, including a5.

The issue, as detailed in [my comment above](https://github.com/processing/processing/issues/2953#issuecomment-63483130), is that the Contribution Manager will fail if it receives a contribution type other than "library", "tool", "mode", or "examples-package" for any version of Processing released to date. However, ["examples-package" was changed to "examples"](https://github.com/processing/processing/commit/31a186ace3318d20fb88730a58cc95519548b9ab) soon after the a5 release, so adding an "examples" contribution would break all releases out there.

We should definitely wait until a6 is released before adding any "examples" contributions.

Also, I strongly suggest we change the `return` statement [here](https://github.com/processing/processing/blob/master/app/src/processing/app/contrib/ContributionListing.java#L509) to something like a `continue` in order to avoid this problem with future releases.

benfry:
[...]

PR #3132, opened 2015-02-27
Contrib Manager does not stop parsing contribs.txt if an error exists

joelmoniz:
The implementation is a slightly modified version of @prisonerjohn 's suggestion [here](https://github.com/processing/processing/issues/2953#issuecomment-71921649). It has been tested to run well with both the current contribs.txt as well as a the contribs.txt [here](https://github.com/joelmoniz/G4P_Examples/blob/master/contributions.txt) which contains entries for the now  renamed ```examples-package``` type.


benfry:
Great, thanks!

Issue #3136, opened 2015-03-01
Cut and Copy shortcuts broken on HEAD


aengelke:
When issuing Ctrl-C or Ctrl-X on a selection in the editor, it does not copy or cut. The menu items work fine.

benfry:
Fixed by @akarshit.

PR #3138, opened 2015-03-05
Cut and Copy are now working .

Akarshit:
Fix for #3136 and the discussion done in #3057 
The problem was that when Edit menu was opened without selecting any text the cut and copy options were disabled but when the menu was closed these options remained disabled and after that when cut/copy command was executed a doClick() was done on these items which were left disabled thus these command were not working.


benfry:
Awesome, thanks for tracking this down.

PR #3141, opened 2015-03-06
Added a confirmation Dialog to About Window

coolutpal:
On clicking About Window it didn't prompt for conformation of cancellation rather the window just diposed off . Added it just as an enhancement .


benfry:
The current implementation is frequently used in software... A confirmation dialog is to prevent the user from performing a destructive action. The suggested implementation would be an annoyance.

Issue #3120, opened 2015-02-22
colorMode(RGB) and colorMode(HSB) should restore default ranges


teo1978:
Example code:
  colorMode(RGB);
  // now the ranges are 255,255,255 in RGB
  colorMode(HSB, 360, 100, 100);
  // now the ranges are 360 for H, 100 for S and B
  colorMode(RGB);

expected: the ranges should be now 255, 255, 255 for RGB. 
observed: the ranges are now 360 for R, 100 for G, 100 for B

The obvious behavior that anyone would expect is:
- if I specify ranges, use those ranges
- if I don't specify ranges, use default ranges.
It makes no sense to maintain current ranges when explicitly calling colorMode() without specifying ranges, especially when switching mode 

(remembering ranges in stateful manner per-mode, i.e. having current ranges for RGB and current ranges for HSB, could arguably make some sense, as long as it was documented, but I don't think it would be the most expected or desirable behavior).

Akarshit:
I would be willing to implement this once someone confirms that this is the required behaviour.

benfry:
The reference doesn't specify default ranges for either (outside of the default when you start being `colorMode(RGB, 255)`), so I think implementing them will cause more problems than it solves. There isn't an obvious solution here, and the various solutions each come with their own downsides, including breaking existing code.

A better option is just to specify the ranges in your code. Using `colorMode(HSB, 360, 100, 100)` explicitly is better self-documentation for anyone looking at the code to understand how it works.

teo1978:
> There isn't an obvious solution here

How not? I see two possible obvious solutions: either remembering ranges per-mode, or always restore defaults when switching modes, whichever you prefer. I don't see any other reasonable behavior.

> the various solutions each come with their own downsides, including breaking existing code.

Could you name any other downide besides breaking existing code? I've seen changes in Processing break existing code pretty often, when there was a good reason for it, and it seems to be the case here.

>A better option is just to specify the ranges in your code. Using colorMode(HSB, 360, 100, 100) explicitly is better self-documentation for anyone looking at the code to understand how it works.

Then the change would only break poorly-written existing code. Indeed, if that is the philosophy, then why not require always the three parameters?

PR #3142, opened 2015-03-07
Defalut value restored when switching modes in colorMode()

Akarshit:
Probable fix for #3120 .


benfry:
Opting not to do this... We don't specify that a default for HSB anywhere, and changing this behavior would also break existing sketches.

Issue #2784, opened 2014-08-13
setStroke() method for PShape in JAVA2D throws error, but not in P2D/P3D


REAS:
```
PShape zig;

void setup() {
  size(100, 100);
  //size(100, 100, P2D);
  zig = createShape();
  zig.beginShape();
  zig.fill(0);
  zig.vertex(10, 0);
  zig.vertex(100, 30);
  zig.vertex(90, 70);
  zig.vertex(100, 70);
  zig.vertex(10, 90);
  zig.vertex(50, 40);
  zig.endShape(CLOSE);
  frameRate(4);
}

void draw() {
  background(204);
  color strokeVal = color(random(255));
  color fillVal = color(random(255));
  zig.setStroke(strokeVal);
  zig.setFill(fillVal);
  shape(zig, 0, 0);
}
```

Akarshit:
@benfry @galsasson @shiffman I hope nobody was working on this , i have issued a PR with some suggestions please look into it. Further suggestions are welcome.

PR #3143, opened 2015-03-07
Stroke can be set in PShape

Akarshit:
Fix for #2784  
A small modification could be to make the `is3D` variable global . But i was not sure if it would be too much of a change that it might break some other thing.  
But doing so would definitely increase the speed.

### DIFF
--- a/core/src/processing/core/PShape.java
+++ b/core/src/processing/core/PShape.java
@@ -661,11 +661,13 @@ public class PShape implements PConstants {
 
   public void vertex(float x, float y) {
     if (vertices == null) {
-      vertices = new float[10][2];
+      vertices = new float[10][PGraphics.VERTEX_FIELD_COUNT];
     } else if (vertices.length == vertexCount) {
       vertices = (float[][]) PApplet.expand(vertices);
     }
-    vertices[vertexCount++] = new float[] { x, y };
+    vertices[vertexCount][X] = x;
+    vertices[vertexCount][Y] = y;
+    vertexCount++;
 
     if (vertexCodes == null) {
       vertexCodes = new int[10];
@@ -1714,10 +1716,18 @@ public class PShape implements PConstants {
     if (vertices == null) return;
 
     boolean insideContour = false;
+    boolean is3D = false;
     g.beginShape();
 
+    for (int i = 0; i < vertices.length; i++) {
+      if(vertices[i][Z] != 0.0){
+        is3D = true;
+        break;
+      }
+    }
+
     if (vertexCodeCount == 0) {  // each point is a simple vertex
-      if (vertices[0].length == 2) {  // drawing 2D vertices
+      if (!is3D) {  // drawing 2D vertices
         for (int i = 0; i < vertexCount; i++) {
           g.vertex(vertices[i][X], vertices[i][Y]);
         }
@@ -1730,7 +1740,7 @@ public class PShape implements PConstants {
     } else {  // coded set of vertices
       int index = 0;
 
-      if (vertices[0].length == 2) {  // drawing a 2D path
+      if (!is3D) {  // drawing a 2D path
         for (int j = 0; j < vertexCodeCount; j++) {
           switch (vertexCodes[j]) {
 


core/src/processing/core/PShape.java
```diff
@@ -661,11 +661,13 @@ protected void endContourImpl() {
 
   public void vertex(float x, float y) {
     if (vertices == null) {
-      vertices = new float[10][2];
+      vertices = new float[10][PGraphics.VERTEX_FIELD_COUNT];
```
benfry: VERTEX_FIELD_COUNT is not necessary; this should be 2 or 3.

benfry:
I don't know what this code does to fix #2784, did you submit everything?

Akarshit:
The problem was that the vertices array was of 2 columns but various properties about the shape (eg. stroke) are set from the 3rd column , like stroke is stored in the 17th column.  
So i made the vertices array of required dimensions and for checking if the shape is 2D or 3D there is the for loop.

benfry:
Hm, I see (I wasn't aware that the recent PShape refactoring had been implemented that way)... but it also means we need to vet the code to make sure that default values are filled in for those fields on other vertices.

Akarshit:
I think you are right and will look into this.  
But i have a small problem when i try to run a P2D code in eclipse it's giving thw following error,  can you give the solution for this.
`Exception in thread "main" java.lang.UnsatisfiedLinkError: no lwjgl64 in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1865)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
	at java.lang.System.loadLibrary(System.java:1119)
	at org.lwjgl.Sys$1.run(Sys.java:72)
	at java.security.AccessController.doPrivileged(Native Method)
	at org.lwjgl.Sys.doLoadLibrary(Sys.java:66)
	at org.lwjgl.Sys.loadLibrary(Sys.java:87)
	at org.lwjgl.Sys.<clinit>(Sys.java:117)
	at org.lwjgl.opengl.Display.<clinit>(Display.java:135)
	at processing.opengl.PSurfaceLWJGL.initFrame(PSurfaceLWJGL.java:113)
	at processing.core.PApplet.initSurface(PApplet.java:9481)
	at processing.core.PApplet.runSketch(PApplet.java:9427)
	at processing.core.PApplet.main(PApplet.java:9285)
	at processing.core.PApplet.main(PApplet.java:9269)
	at com.akarshit.wal.Main.main(Main.java:7)
`

benfry:
You need to add `-Djava.library.path=/path/to/folder/containing/lwjgl64.dylib` to your run options. Welcome to Eclipse! (and Java coding without the PDE)

Akarshit:
@benfry Thanks for the tip, it got the things working.  
In PShape i think 
`strokeColor = stroke;
    if (vertices != null) {
    for (int i = 0; i < vertices.length; i++) {  
    setStroke(i, stroke);  
    }  
    }
`
at https://github.com/processing/processing/blob/master/core/src/processing/core/PShape.java#L2457 is unusable.  
Similarly for other functions corresponding code is unusable as the co-ordinates are copied to a PGraphicsJava2D object later and we cannot set the values for individual edges in Java2D.
But i am not sure as i could not understand the implementation fully, perhaps the author of this could shed more light on this.  
Some things i noticed while debugging in PShape are (not sure if they are bugs or intentional) -
*  `/**
When drawing POLYGON shapes, the second param is an array of length
 VERTEX_FIELD_COUNT. When drawing PATH shapes, the second param has only
 two variables.
/  `
`protected float[][] vertices;`
But this has not been implemented here.    

* `public void setTint(int fill) {  

if (openShape) {  
PGraphics.showWarning(INSIDE_BEGIN_END_ERROR, "setTint()");  
return;  
}  
tintColor = fill;  
if (vertices != null) {  
for (int i = 0; i < vertices.length; i++) {   
setFill(i, fill);  
}  
}  
}
`
  
This might be a copy paste error(judging by the variable name `fill`) and also the `setTint(int index, int tint)` is not getting called anywhere.

benfry:
(Just did some formatting changes to your last entry to make the code readable.) 

I'm gonna close this PR since it's not something to be merged. Perhaps you can file the other entries as separate items so that someone can follow up?

PR #3155, opened 2015-03-16
Automatically install libraries that a user imports, if not already installed

### LINKED ISSUE #2566: Check if library is installed before running sketch; if not, prompt to install
When sharing sketches with others, the sketches can't be run unless all required libraries are installed.

Add logic that, when a sketch is about to be run, checks to see if the required libraries are installed. This could be based on looking at any import statements in the code, and then taking a best guess about what's needed.  For example, if my code includes: `import peasy.*;` we can assume I would need the PeasyCam library installed.

If the libraries are installed, run per usual.  If they are missing, introduce a prompt to facilitate download and installation through the contribution manager.  (For example, "This sketch may require ___ and ___. Download and install them now?")

Request posted per @benfry in our conversation today.

joelmoniz:
This fixes #2566. It basically checks for whether the imported libraries are already installed or not. If one or more of them aren't, and if any/all of them are available, a dialog menu is displayed to the user listing the available, not installed libraries. Selecting ```Yes``` installs the libraries and runs the sketch, while selecting ```No``` simply runs the sketch directly (which eventually displays an error due to the missing library).

Note: This adds a new field, ```imports```, to the ```library.properties``` file, where the most commonly used headers of the library are listed.

### DIFF
--- a/app/src/processing/app/contrib/ContributionManager.java
+++ b/app/src/processing/app/contrib/ContributionManager.java
@@ -334,8 +334,8 @@ public class ContributionManager {
             String statusMsg = base.getActiveEditor().getStatusMessage();
             if (statusMsg.contains("has been installed"))
               base.getActiveEditor().statusNotice(statusMsg + " "
-                                                    + "Now downloading "
-                                                    + ad.name);
+                                                    + "Downloading "
+                                                    + ad.name + "...");
             else
               base.getActiveEditor().statusNotice("Downloading " + ad.name
                                                     + "...");
@@ -377,7 +377,7 @@ public class ContributionManager {
     base.getActiveEditor().statusEmpty();
     System.out.println("The following libraries have been installed:");
     for (String l : installedLibList) {
-      System.out.println(l);
+      System.out.println("  • " + l);
     }
   }
 
--- a/app/src/processing/app/contrib/ContributionManager.java
+++ b/app/src/processing/app/contrib/ContributionManager.java
@@ -197,7 +197,10 @@ public class ContributionManager {
 
   /**
    * Non-blocking call to download and install a contribution in a new thread.
-   *
+   * Used when information about the progress of the download and install
+   * procedure is not of importance, such as if a contribution has to be 
+   * installed at startup time.
+   * 
    * @param url
    *          Direct link to the contribution.
    * @param ad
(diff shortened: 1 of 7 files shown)


Manindra29:
@joelmoniz Did you check with @prisonerjohn about adding new fields to library.properties? This modification shouldn't break compatibility with existing libraries.

joelmoniz:
Oops. No, I haven't checked with him. @prisonerjohn , could you please
check up on the changes I've made? As far as I can tell, they won't break
existing libraries in any way, but merely provide the functionality to
automatically install the library if imported by the user when the new
field is present.

NB: This field is required only for AvailableContributions, not
LocalContributions, since it is needed only to install the library, and not
once it is installed

benfry:
Looks like we need an update for this one so that I can merge it against the current repo.

app/src/processing/app/contrib/ContributionManager.java
```diff
@@ -284,13 +287,118 @@ public boolean accept(File dir, String file) {
...
+            base.getActiveEditor()
+              .statusNotice(Language
+                              .interpolate("contrib.import.progress.install",
+                                           ad.name));
+            LocalContribution contribution = ad.install(base, contribZip,
+                                                        false, null);
+
+            if (contribution != null) {
+              contribListing.replaceContribution(ad, contribution);
+              if (base.getActiveEditor() != null)
```
benfry: Please always use { and } for all blocks, no single statements, please. Too prone to errors.

app/src/processing/app/contrib/ContributionManager.java
```diff
@@ -284,13 +287,118 @@ public boolean accept(File dir, String file) {
...
+
+          try {
+            // Use the console to let the user know what's happening
+            // The slightly complex if-else is required to let the user know when
+            // one install is completed and the next download has begun without 
+            // interfereing with occur status messages that may arise in the meanwhile
+            String statusMsg = base.getActiveEditor().getStatusMessage();
+            if (isPrevDone)
+              base
+                .getActiveEditor()
```
benfry: We don't use this type of indenting elsewhere in the source code. Just get things on one line, or if they don't fit in 80 chars, break up the statements.

joelmoniz:
Whoops. Sorry about the formatting, lack of enclosing curly braces and the merge conflict. I shall update things and push up another commit right away.

joelmoniz:
@benfry Sorry about that. The brackets and indentation ought to be fixed now, and things should be auto-mergeable.

benfry:
Thank you; much appreciated!

Issue #3129, opened 2015-02-26
Sketch window dimensions off in Java2D


shiffman:
Only tested on Mac OSX retina, 10.10.2

JAVA2D:
![screen shot 2015-02-26 at 11 52 08 am](https://cloud.githubusercontent.com/assets/191758/6391440/2db80752-bdae-11e4-91a4-07f2f1a765ca.png)

P2D:
![screen shot 2015-02-26 at 11 52 22 am](https://cloud.githubusercontent.com/assets/191758/6391443/33a89c6c-bdae-11e4-9195-dc1382fed19f.png)

Code:

```
void setup() {
  size(200, 200);
  //size(200, 200, P2D);
}

void draw() {
  background(0);
  fill(255);
  ellipse(100, 100, 100, 100);
}
```

Akarshit:
Not occuring in linux. Something specific to Mac .

shiffman:
I neglected to mention that I'm testing in the current build 0233.

benfry:
I've been seeing this too. More fixing to do; thanks for filing.

benfry:
Fixed for 3.0a6 with help from #3129.

PR #3162, opened 2015-03-29
Fix for #3129 (Incorrect Java 2D window dimensions)

williamahartman:
This is a fix for issue #3129.

The call to getInsets() was returning all 0s because the method requires the JFrame it is called on to have been previously shown. A call to frame.show() was added. 

This change made a small section of Linux-specific code (lines 577-588 in setVisible() ) unnecessary, so they were removed. 

These changes broke the call to setUndecorated() for present mode. A call to dispose() was added to release native resources and allow the call to setUndecorated(). The call to frame.setVisible() a few lines later re-adds the native resources.

The patch was tested with Java 1.8.0_31 on Windows 10 and Java 1.8.0_40 on Fedora Linux.

### DIFF
--- a/core/src/processing/core/PSurfaceAWT.java
+++ b/core/src/processing/core/PSurfaceAWT.java
@@ -573,19 +573,6 @@ public class PSurfaceAWT implements PSurface {
 
   public void setVisible(boolean visible) {
     frame.setVisible(visible);
-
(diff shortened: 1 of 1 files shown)


benfry:
Thank you! I've incorporated this with a few edits (i.e. using `addNotify()` instead of `show()`, since it's a matter of getting the peer). Let me know if I've missed anything.

PR #3170, opened 2015-04-01
Fix multi-touch horizontal scrolling on OSX

### LINKED ISSUE #180: Horizontal two finger scroll doesn't work in OS X
_Original author: b...@processing.org (June 07, 2010 01:10:23)_

This bug automatically added from:
http://dev.processing.org/bugs/show_bug.cgi?id=1046

Comment from ewjordan, 2008-11-23 16:22

Very low priority RFE: horizontal two finger scrolling doesn't work in the
PDE the way it does in most other applications on OS X.

Somehow this is probably possible in Java (Eclipse manages to get it to
work), but it might not be easy, and a quick Google didn't pop up any
information (doesn't help that Apple's support for Java is pretty much
non-existent these days).

Maybe one for the distant future if someone happens to stumble across a way
to enable it...

_Original issue: http://code.google.com/p/processing/issues/detail?id=141_

Plastix:
Horizontal scroll works but it is a little wonky because of the multiple scroll events from Apple's inertia scroll.

This PR also edits the native cocoa wrapper to disable inertia scrolling (See https://developer.apple.com/library/mac/releasenotes/DriversKernelHardware/RN-MagicMouse/). According to Apple that should work, but I'm not sure it does anything.

 Fixes #180

### DIFF
--- a/app/src/processing/app/syntax/JEditTextArea.java
+++ b/app/src/processing/app/syntax/JEditTextArea.java
@@ -140,10 +140,12 @@ public class JEditTextArea extends JComponent
 //    focusedComponent = this;
 
     addMouseWheelListener(new MouseWheelListener() {
+
+      @Override
       public void mouseWheelMoved(MouseWheelEvent e) {
         if (scrollBarsInitialized) {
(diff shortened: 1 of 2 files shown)


benfry:
Thanks, we'll give it a shot.

Issue #2945, opened 2014-11-14
improve start time by populating sketchbook/libraries on threads


benfry:

benfry:
@joelmoniz it's for the methods that rebuild the sketchbook menu. It's very slow to walk the sketchbook hierarchy, looking for sketches, libraries, tools, etc. If you have dozens of sketches, or a even a single sketch that's used `saveFrame()` to create 1000s of files, the I/O will be extremely slow. (Even worse if your machine doesn't have an SSD.)

joelmoniz:
Ah, thank you. I'm almost done working on this, and should be able to file PR by tomorrow. Could you please clarify one small point for me? Does the sketchbook menu refer to File>Sketchbook, or to Sketch? I did some profiling, and the former does not seem to affect the start time at all, since the first time the sketchbook is actually walked searching for sketches is [here](https://github.com/processing/processing/blame/master/app/src/processing/app/Editor.java#L770), when the user actually clicks on the File>Sketchbook menu item. Tool, etc. do however, and I'll submit a PR that improves this.

benfry:
Sorry, I misspoke. It's for the methods that walk the sketchbook... These are for 1) getting the list of sketches (to put in the sketchbook window, in the past this was a menu, which is what I was referring to in the previous comment), 2) getting the modes, libraries, and tools. If you have a large sketchbook, or a lot of libraries installed, all or some of these can take a very long time to load. 

If it's something that's needed immediately, that can't be done on a thread, but depending on the situation, this can happen in the background. The fix isn't terribly difficult, it just requires looking at the code very closely to make sure that we're doing it in the right place and taking all the scenarios into account.

joelmoniz:
Ah, thank you. As far as I could see, the walking is done just-in-time, when the user actually opens the Sketchbook window. I've profiled the code too, and though I had several large sketches (1+ GB each), this doesn't affect start time at all (the method call doesn't even show up on the profiler). I'll look into this a little more and make sure that this is indeed so. 
  
[...]

PR #3178, opened 2015-04-05
Fix for #2944 sketchbook updates properly without restarting on all occasions

JayDusara:
Sketchbook updates automatically when new sketch is saved or when renamed. 

Also if the main sketch is deleted from outside the processing window, the user can press SPACE to refresh and reflect changes.

Also, if the working directory is changed in Preferences window, the tree is updated and shows the sketches in the new directory.


app/src/processing/app/contrib/ProgressMonitor.java
```diff
@@ -0,0 +1,183 @@
...
+//   * This is called when the current task is finished. This should always be
+//   * called when a task is finished, whether or not an error occurred or the
+//   * task was cancelled.
+//   */
+//  public void finished();
+//}
+
+
+//abstract class AbstractProgressMonitor implements ProgressMonitor {
+abstract class ProgressMonitor {
```
benfry: These classes were removed from the source, please don't add them back again.

app/src/processing/app/syntax/JEditTextArea.java
```diff
@@ -163,12 +162,14 @@ public void mouseWheelMoved(MouseWheelEvent e) {
 
             // inertia scrolling on OS X will fire several shift-wheel events
             // that are negative values.. this makes the scrolling area jump.
-            boolean isHorizontal = Base.isMacOS() && e.isShiftDown();
```
benfry: This is undoing recent fixes in the source.

app/src/processing/app/syntax/JEditTextArea.java
```diff
@@ -189,6 +190,7 @@ protected TextAreaPainter createPainter(final TextAreaDefaults defaults) {
   /**
    * Inline Input Method Support for Japanese.
    */
+  @Override
```
benfry: Please don't add annotations like this, especially for classes that aren't affected by this PR.

core/todo.txt
```diff
@@ -73,8 +73,16 @@ X play with improvements to full screen here
 
 
 showstoppers
+<<<<<<< HEAD
```
benfry: And please take care not to mess with the todo files... There should never be any changes to those in a PR.

app/src/processing/app/syntax/JEditTextArea.java
JayDusara: I will take care of the annotations problem...i guess i missed one or two...i will check it out

app/src/processing/app/syntax/JEditTextArea.java
JayDusara: there must be some problem while merging...cuz i had to resolve conflicts manually...i  will correct it

JayDusara:
Corrected the changes as needed. Have a look if everything is okay. 

I had one more doubt in mind. While working on the sketchbook refresh issue, i discovered this. In preferences if you set your sketchbook folder to some folder having large number of files (for examples sake say i set it to D: (d-drive)), then it take a lot of time for the PDE to load. That is because the addSketches method is called recursively on the directory to check if a subfolder of this folder contains sketch. Can this be called an issue?

benfry:
Yes, so now you understand why we don't reload the entire sketchbook whenever a file is saved or renamed. A related issue: https://github.com/processing/processing/issues/2945

On new/rename/save as, only the portion of the sketchbook tree that has changed (the one file) should be updated. Re-indexing the full sketchbook is way too slow (and overkill) for that situation.

JayDusara:
How about this:
Similar to what eclipse does...whenever we define some directory as eclipse workspace, it creates a .metadata folder from which it picks up what all projects were added to the IDE and other settings. Going on similar lines,
1. when the user is saving a program in this directory for the first time, we add the full path to the index which is stored in a separate file, that file is read only and hidden.

2. Henceforth whenever the user saves any program in this directory, it is added to the index

3. Whenever the PDE is started it can directly pick up the index from this index file. For each entry in the index file, we can check if the entry is still valid...i.e. the file has not been moved/deleted etc. (can this checking run on a separate thread so it will not slowdown the PDE loading)

4. If file is not found for some reason, ask the user if he wants to search it in the directory or knows the location in case he/she moved it.

5. Also, we keep a copy of the index file in a master folder or perform some kind of logging to retrieve entries in case index file itself is affected in that directory

6. If somehow the index file in the working directory is deleted, try and retrieve it from the master file.

7. Finally if all of above fails and user still has files in that directory, and if he/she says yes when prompted to rebuild the indexes, then call those recursive methods

(Besides, there will be other things to take care of)

Has this been already thought of before? or is there some point i am missing regarding the working of the code?

benfry:
That's much too complicated, and we have always gone out of our way to avoid adding extra metadata or properties files to the sketch folder. If people want Eclipse, they should use Eclipse.

There are three things here:

1. On new/rename/save as, the entry for that sketch needs to update in the sketchbook menu. That's a regression, and needs to be fixed. https://github.com/processing/processing/issues/2944
2. When the sketchbook is (or even files in the current sketch are)  modified outside Processing, it should automatically reload. This is handled in other pull requests, but the support is buggy so it's not turned on. https://github.com/processing/processing/issues/2759
3. Third is the performance of sketchbook loading, which should be carefully moved to an independent thread: https://github.com/processing/processing/issues/2945

#1 is a very high priority. #2 is nice to have. #3 is also a lower priority.

benfry:
Closing this one since it's out of date, it mixes several of the issues at hand (see my last comment), and also includes quite a bit of code that's not related (language, core, and even todo.txt edits).

Issue #3172, opened 2015-04-03
Install/Remove button in Mode/Library/Tool managers not working in the current source


REAS:
OS X 10.8.5, running from source. 

The install button is there and it highlights on rollover, but there is no action when it's pressed.

benfry:
@joelmoniz any thoughts on what's happening here?

Plastix:
I am replicating this on OSX 10.10 and running latest commit from master. Will investigate.

Plastix:
I figured out that you have to click the install/remove button (so it is selected) and press space to actually send the click event. No idea why...

joelmoniz:
@benfry
Terribly sorry. I'm afraid I've not yet been able to look into this, since
I've had loads of work at school this last week. Will definitely be able to
look into this tomorrow.

On a side note, based on @Plastix 's tests, might this be related to #3079?
@Plastix could you please let me know if double clicking the install/remove
button enables installation/removal?

Thanks!

Plastix:
@joelmoniz No matter how many times I click the buttons it doesn't install/remove the library. The only way I've found is to click the button to select it and press space. Code is all over the place in `ContributionPanel.java` and it is hard to tell what is going wrong.

Here is a gif: 
![test](https://cloud.githubusercontent.com/assets/1526881/6994385/8f79a712-dae5-11e4-9935-3fff8c6b8918.gif)

joelmoniz:
Oh wow. Thanks a bunch. Probably related to #3079, but definitely not a
duplicate, since the behavior on Windows isn't like this at all, although
I'm fairly certain that the same bug in the Contributions Manager is at the
root of both these issues. I shall definitely have a look at this asap.

Thanks again.

PR #3183, opened 2015-04-06
Contribution Manager buttons are no longer broken

joelmoniz:
This aims to provide a temporary fix to #3172. Right now, I've modified the mouse listener added to each ContributionPanel to pass on any clicks it receives to either the button dealing with installs, removes and undos, or the one dealing with updates (depending on where the user clicked and whether or not the button was visible).  
  
NB: Although it is bug-free, I think this is a temporary fix, since the Contribution Manager ran fine without this up until 2 months ago. I, however, am afraid that I have no clue about what changed, since, as far as I can tell, even the git logs don't seem to show any changes between 2-6 moths ago.

### DIFF
--- a/app/src/processing/app/contrib/ContributionPanel.java
+++ b/app/src/processing/app/contrib/ContributionPanel.java
@@ -237,6 +237,24 @@ class ContributionPanel extends JPanel {
 
     setExpandListener(this, new MouseAdapter() {
       public void mousePressed(MouseEvent e) {
+        // TODO: Fix Contribution Manager buttons
+        // The following is a temporary solution that calls all the ActionListeners 
+        // associated with a button when a user clicks on it
+        if (ContributionPanel.this.installRemoveButton.contains(e.getPoint())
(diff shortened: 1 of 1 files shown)


benfry:
Hm, this can't be a good route. Let's figure out what's actually broken. 

If nothing changed in there, that sounds suspiciously like a threading problem: https://docs.oracle.com/javase/tutorial/uiswing/concurrency/

joelmoniz:
Ah yes. I was afraid this was a little hack-y, and I'd intended this to only be a temporary fix till the real problem was figured out. The application doesn't seem unresponsive though, since the panel gets highlighted. The only problem is that the button has to be selected first before clicking on it (which ends up in the button effectively requiring a double click). Could you please let me know why you suspect a threading problem? Thanks.

benfry:
Because the UI is unresponsive, and because you said no changes were made to the code since the time it was working properly. Since the time you cite, we've moved to a different version of Java, which is the sort of change that will reveal problems like that.

PR #3186, opened 2015-04-08
Fix ColorChooser cursor

federicobond:
Previously, the whole window would show a crosshairs cursor when you hovered on it. Now only the slider and range components do.


benfry:
Good call; thanks.

PR #3190, opened 2015-04-14
Find/Replace dialog uses GroupLayout

Akarshit:
findReplaceDialog
The old code has not been cleaned yet and there is also no documentation, i will do these once the design is agreed upon.  
This produces a dialog similar to the already present one but some changes that could be made are 
* Enclose the checkboxes in a border under Options heading
* Move 2 buttons to the next line
* Make sizes of all the buttons equal


benfry:
Cool, thank you!

PR #3191, opened 2015-04-14
Initial Stage for Preference Window with GroupLayout

Akarshit:
Fix for #67 
newprefwindow
This is very similar to the current Preference Window.
The old code has not been cleaned, i will do it once design is finalized ( and also add comments ). 
Some suggestion i think should be done are 
* Size of Ok and Cancel should be increased
* Window should be made resizable
* "Background color when Presenting" should be changed to something more meaningful


benfry:
Cool, thanks! That was an oldie.

Issue #3148, opened 2015-03-12
Newline after a very long line moves the visible area right. 


Akarshit:
Steps to reproduce ->  
Type a very long line in PDE which is more than the width of PDE . 
Press enter and start typing some text .
The text gets typed in PDE but is not visible .

Akarshit:
This is most probably because somewhere offset for Gutter is getting added twice .

benfry:
Incorporated for 3.0a7.

PR #3196, opened 2015-04-15
Now the cursor moves when the line is changed after a long line

Akarshit:
Fix for #3148 

Though this fixes the current issue there is a lot of  buggy areas in horizontal offset .
It would be nice if the author can document these function in detail so that they can be made bug free.

### DIFF
--- a/app/src/processing/app/syntax/JEditTextArea.java
+++ b/app/src/processing/app/syntax/JEditTextArea.java
@@ -563,11 +563,11 @@ public class JEditTextArea extends JComponent
     int width = painter.getFontMetrics().charWidth('w');
 
     if(x < 0) {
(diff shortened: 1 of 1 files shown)


benfry:
The author is not part of the project, it's old/abandoned code: http://syntax.jedit.org/

Akarshit:
So do you have any plan/idea of redoing this with something better or atleast well maintained.

benfry:
Of course, but it's been 14 years, and nobody's had a chance to execute on the plan.

Akarshit:
@benfry Ok i will open an issues for this, but i won't be able to work on this right now as exams are starting. Hopefully someone would like to implement this.

PR #3197, opened 2015-04-17
improve internationalization and localization in greek

mtsio:


benfry:
Thanks, also modified the `translateCategory()` to something more general with less code.

mtsio:
yeah, much better now

PR #3201, opened 2015-04-19
Graphics overlap fix in preferences

jgall:
the check box to "show warnings" in preferences is obscured by other
objects in the frame, making it difficult to click. This is just a quick fix for it, adding a tiny bit
of padding in between it and the setting to its left.


Akarshit:
@jgail 
The layout of PreferenceFrame has changed, it now uses group layout. You are using the old code, this bug does not occur in the latest code. If any other bug related to preference window comes to light please inform me.

PR #3202, opened 2015-04-21
Add option to save JSON in compact form

federicobond:
This puts the third parameter in `saveJSONObject(json, filename, options)` to some use. Main use case is storing big amounts of JSON data without wasting space on disk.

### DIFF
--- a/core/src/processing/data/JSONObject.java
+++ b/core/src/processing/data/JSONObject.java
@@ -1554,12 +1554,19 @@ public class JSONObject {
 
 
   public boolean save(File file, String options) {
-    return write(PApplet.createWriter(file));
+    return write(PApplet.createWriter(file), options);
   }
 
-
   public boolean write(PrintWriter output) {
-    output.print(format(2));
+    return write(output, null);
+  }
+
+  public boolean write(PrintWriter output, String options) {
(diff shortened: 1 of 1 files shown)


benfry:
This should probably be handled with an option that does things like `indent=2` or `indent=4`. Question is whether this particular case should be `indent=none` or `indent=compact` (since it's different from `indent=0`). Or if we should just add `compact` as you've done here. 

I wish there were a standard term for this—"compact" is colloquial (and not a word we use elsewhere) but could work, "minified" is more known but strictly speaking, not what we're doing here, etc...

federicobond:
Yes, those are some good points. I had done a bit of research and `compact` seemed to [come](http://jsonformatter.curiousconcept.com/) [up](http://www.freeformatter.com/json-formatter.html) frequently in tools to refer to this particular "preset" for outputting JSON.

I thought about adding an `indent` option with arguments, but I couldn't come up with much use cases for it. As a user, the choice is usually between optimizing for human-readability (pretty-printing) and optimizing for storage efficiency (compact). Anyway, If the need arises, we are not closing the doors to adding an explicit `indent` option with argument.

benfry:
Works for me, thanks for giving it some thought.

PR #3205, opened 2015-04-21
Improvements to SVG. Fixes #2992.

GKFX:
Proper support for percentages, partial support for viewbox, full
support for named colors, allow gradients to specify color-stops using
style or direct attributes on &lt;stop&gt; (fix #2992), nicer error message on loading an
HTML document as SVG (since I assume it can happen to novices
downloading SVGs from the web).

~~Also unrelated edit to PApplet which disabled a doc comment which was
coming up in the wrong place.~~

PR #3207, opened 2015-04-23
PApplet.java: disable a doc comment

GKFX:
It's coming up in the wrong place on the Javadoc because its method is commented out.
Already mentioned at #3205 but you wanted it in a separate commit.


benfry:
Thanks!

PR #3211, opened 2015-04-24
Removed extra edges in sphere tesselation (#3193)

### LINKED ISSUE #3193: Sphere overlapping itself?
I think the sphere() command may be producing a mesh which may be overlapping itself (but I'm not sure if that's what's causing the artifact I'm seeing). If you use a semi-transparent stroke, you'll notice a bright band compared to the rest of the sphere, probably because of lines drawn on top of each other. I suspect that the loop which draws the sphere goes one incremental step too far in the "theta" angle.

Image - note the bright lines along one band:

sphere_too_far

Code:

```Java
import peasy.*;
PeasyCam cam;

void setup() {
  size(800,600,P3D);
  cam = new PeasyCam(this,0,0,0,500);
  background(255);
}

void draw() {
  background(0xFFFFFFFF);
  drawSphere();
}

void keyPressed() {
  saveFrame("sphere_too_far.jpg");
}

void drawSphere() {
  sphereDetail(100,100);
  stroke(0x40FFFFFF);
  fill(0xff000000);
  sphere(400);
}
```

JakubValtar:
Tested in 3.0a5, latest does not run for me.

Fixes #3193.

### DIFF
--- a/core/src/processing/opengl/PGraphicsOpenGL.java
+++ b/core/src/processing/opengl/PGraphicsOpenGL.java
@@ -8863,9 +8863,6 @@ public class PGraphicsOpenGL extends PGraphics {
         indices[indCount + 1] = vert1 - detailU;
         indices[indCount + 2] = vert1 - 1;
         indCount += 3;
-
-        addEdge(vert1 - detailU, vert1 - 1, true, true);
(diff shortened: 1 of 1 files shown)


benfry:
Thanks!

Issue #3213, opened 2015-04-25
Find/Replace layout regressions in 3.0a6


benfry:
Our carefully created Find/Replace window from previous releases:
![find-replace](https://cloud.githubusercontent.com/assets/1623101/7332017/a17c8adc-eaf9-11e4-82ba-497d543ab690.png)

Has taken a step backwards since https://github.com/processing/processing/pull/3190
![find-replace-new](https://cloud.githubusercontent.com/assets/1623101/7332019/c294439a-eaf9-11e4-96be-7cce3231752a.png)

benfry:
Fixed the bounds, but we're still definitely in the "ugly Java UI" category, which is why I didn't use a layout manager in the past and ran it manually. 

![find-mac](https://cloud.githubusercontent.com/assets/1623101/7333758/3e770a66-eb49-11e4-9f37-cb606ac3200c.png)

This should be a straightforward project to fix up, please help!

benfry:
Main issues are: 
* Need a little border on all sides
* 'Find:' should be right-aligned
* The 'Ignore Case' button should align left to the left edge of the text fields (the other checkboxes can stay adjacent—they shouldn't necessarily be spaced out like they were in the original).

Akarshit:
Sorry, i was busy with some college project for my exams.

This is not a issue that is happening on Ubuntu all the time, some time it does and some time the size is ok ( that's why it passed through my testing ). I will try to fix this today.

benfry:
Fixed for 3.0a7.

PR #3216, opened 2015-04-26
Find/Replace Window now looks like original.

Akarshit:
Fix for #3213


benfry:
Great, thanks for looking into these.

Issue #3212, opened 2015-04-25
Preferences window elements not sized correctly in 3.0a6


benfry:
Regressions introduced by https://github.com/processing/processing/pull/3191

![preferences-cropped](https://cloud.githubusercontent.com/assets/1623101/7332010/41c8fbd4-eaf9-11e4-929b-5a30bb7860e4.png)

On OS X, the window contents are not sized correctly. 

In addition, the "Editor font size" menu is enormous and runs the width of the window.

benfry:
Fixed the bounds, but we're still definitely in the "ugly Java UI" category, which is why I didn't use a layout manager in the code in the first place. 

![prefs-mac](https://cloud.githubusercontent.com/assets/1623101/7333762/8d761ad0-eb49-11e4-9aec-32a218542675.png)

This should be a straightforward project to fix up, please help!

benfry:
And for reference, this is what it looked like before the PR:

![screen shot 2015-04-25 at 12 52 51 pm](https://cloud.githubusercontent.com/assets/1623101/7333784/1c3286be-eb4a-11e4-9eb3-d1e5b24c0519.png)

Looks like we also lost the "MB" designation after the memory text field.

benfry:
Hm, on further consideration, I wonder if it's time to have separate tabs for modes. Because of the PDE X integration, half of those items are specific to Java Mode.

Akarshit:
Sorry, i was busy with some college project for my exams.  
This is not a issue that is happening on Ubuntu all the time, some time it does and some time the size is ok ( that's why it passed through my testing ). I will try to fix this today.

benfry:
Thanks for looking into this. I've merged the changes but it looks like we still have some issues w/ the font drop-downs:

![font-boxes](https://cloud.githubusercontent.com/assets/1623101/7337339/9c63e83c-ebf4-11e4-8d7f-c38888adac32.png)

Let's just do setResizable(false) as well... Resizing the window doesn't do much for us.

Akarshit:
I will hard-code these values for now.

benfry:
They can't just use their default width and have extra empty space to their right (ala `BoxLayout` or `FlowLayout`)?

Akarshit:
Actually they are working fine on Ubuntu, but i will try to do something to make them compatible with Mac. The main problem is i don't have access to a Mac to test.

benfry:
[...]

PR #3217, opened 2015-04-26
Changed the Pref Window a little to fix some bugs.

Akarshit:
Fix for #3212 

Can someone please chech if the size of the fields of Editor Font Size and Console Font Size are correct on Mac as this seems fine on Ubuntu.
If the window needs to be split in to tabs please inform me and i will implement it.


benfry:
I'll give a little more thought to how the prefs windows are handled before we break things up. 

Font menus are off on OS X, they should just be the default width and then we can have open space to their right.

Akarshit:
What do you mean by Font Menu are off ?
A screenshot would be a great help.

Issue #3208, opened 2015-04-23
Window size not passing into Tweak Mode in 0233


REAS:
The Display Window is showing up at 100 x 100 pixels when running in Tweak mode, regardless of the dimensions set in size()

benfry:
I suspect that Tweak is doing its own parsing of the `size()` command and not doing it properly. Went looking for it but didn't find anything immediately.

shiffman:
@galsasson any ideas?

galsasson:
Hi, can you please describe the issue again?
I'm not sure I understand.

REAS:
When you start a sketch with Tweak, it's not actually the mode anymore as mentioned above, the dimensions from size() aren't passed through and the window shows up as 100 x 100 pixels. This feature is there in the current source and will be in 3.0a6.

galsasson:
thanks, I will take a look.

galsasson:
Yes, this looks like the same issue, I could reproduce it only when passing a third argument to size.
TweakMode adds two function calls to the beginning of setup (before any user code), so it might be the case. 
Also, tweak treats width and height as tweak-able values (and as a result the modified sketch is actually using variables to provide these numbers) and reading the 'size' function reference now it says that numbers should be passed to this function and not variables.
@benfry, Is this a real?

Is there currently a hook to execute code before setup?
If not, I will need to add some code so that tweak will ignore the numbers in setup (which don't really make sense to tweak anyway), then I can move these two functions to the end of setup.

benfry:
The reference has *always* said that size() goes first and that you shouldn't use variables. It's more strict in 3.0a6 due to changes we need to make in order to get better performance.

galsasson:
I created a pull request for this issue (#3227) with some additional improvements.

benfry:
Incorporated for 3.0a8.

REAS:
Wonderful. I confirm it's working.

eric-souther:
Looks like this was fixed in 3.0a8., however no matter what I change with size or fullScreen now when I run tweak mode in 3.0a10 on OS X 10.10.3 I get this error "IllegalStateException: fullScreen() cannot be used here, see https://processing.org/reference/fullScreen_.html" All I have in setup is fullScreen(P3D), and in draw a background and ellipse.

PR #3227, opened 2015-04-28
TweakMode fix to issue #3208

galsasson:
This pull request fixes issue #3208 as well as:
- tweak init functions were moved to the end of setup.
- parser now skips comment blocks
- parser skips global assignments


benfry:
Thanks for the fixes.

Issue #3214, opened 2015-04-25
Sketchbook Window doesn't update after changing sketchbook location in the Preferences


benfry:

Akarshit:
@benfry Is every sketchbook supposed to contain Modes, Tools, Libraries ... folders or is it more analogous with Workspace in eclipse ?

PR #3230, opened 2015-04-30
Sketchbook updating now when the path is changed in the preference without restarting Processing

Akarshit:
Fix for #3214


benfry:
This isn't correct. It would only reset the frame next time it's opened for the current mode. It wouldn't reset a window that was already open, and if editor windows were present for say, Python and Java, then only one would be reset properly. 

For problems like this you need to dig a little further. Changes occur in `applyFrame()`, which calls a method inside `Base` to handle the sketchbook location change. That's where a change like this needs to be made.

Issue #1968, opened 2013-07-12
Tab menu is not accessible when many tabs are open


knupel:
Hello just little comment about the tab GUI design. In my sketch I've 50 tabs and more in the future. And with the new gestion of the window the dropdown is accessible only if the window is really big. May be is not clear so I give two screenshots for examples
Dropdow is not accessible
![dropdown is not here](https://f.cloud.github.com/assets/3332269/787559/6e54b18e-eadd-11e2-9218-2e49dd445af2.png)
Dropdown tab accessible
![the dropdown is here](https://f.cloud.github.com/assets/3332269/787560/71dc1798-eadd-11e2-9480-6e0d33cdbbcb.png)

I hope my comment make sens, because sometime I don't want open my sketch on all my display !!

StanLepunK:
I'm back with this issue. I know is not a priority, but with Processing 306, the tab is bigger, so the limit of displaying tab is lower. Maybe with this new version of the GUI it's time to do something around that ?
Integrate the tabmanager tool in Processing can be good ? or change something in the GUI ; make the right button be always visile, or add arrow right and left arrow to navigate in the tabmanager.
![tab](https://cloud.githubusercontent.com/assets/3332269/7337186/8746ec66-ec20-11e4-9120-d01702638a59.jpg)

benfry:
Processing is simply not intended to be used this way, not built for it, etc. If you have more then 3-5 tabs, you're better off with Eclipse or Netbeans or something else. 

It's great that it works for a dozen tabs, but 20 and 30 in this example is an edge case that will make the code messier and provide little benefit for something we don't recommend anyway.

StanLepunK:
I understand the purpose to keep the simplicity, I love that in Processing that's a reason I want stay in Processing. 
One of big problem, for the artiste-coder like me, to code in Processing in Ecclipse or Netbean, it's complex and hard approche of the GUI, plus it's just a nightmare to build mac application with Eclipse or Netbean.
It's very hard to keep the "dropdown  arrow button" of tab manager visible all the time whatever the number of tab ?

pvrs12:
I know this is low priority, but what about if the arrow were moved to be located on the right side of the window and persist through the addition of tabs. I played with the code that renders the tabs a tiny bit (really just moving the menu-arrow) and came up with this
![Right Menu 1 tab](https://cloud.githubusercontent.com/assets/6956401/7440297/ff7ff04e-f079-11e4-832c-515d90e321ed.png)
[...]

PR #3236, opened 2015-05-02
Potential Tab menu fix

pvrs12:
This is a possible fix for #1968 and other duplicates of this issue. It causes the tab menu to lock to the right of the window when there are too many tabs to display and it doesn't render tabs which would interfere with the menu. Resulting in something that looks like this if there are too many tabs to render.
Too many tabs 
The only potential issue I see is that there is no indication that more tabs are hidden off screen. It might make sense to add another 'tab' which indicates more tabs are hidden (an ellipses or something). 

Hope this fixes this issue


pvrs12:
Just wanted to bump this

benfry:
Please don't bump. I either have the time to evaluate the fix or I don't.

Issue #2911, opened 2014-10-26
File / Examples / Libraries / net shouldn't contain CarnivoreClient


x14c3:
In fresh install of Processing (both 2 and 3) in File / Examples / Libraries / net there is CarnivoreClient example. It shouldn't be there. It need extra library Carnivore (http://r-s-g.org/carnivore/).

benfry:
Moved here: https://github.com/processing/processing-docs/issues/238

PR #3257, opened 2015-05-08
removing Carnivore client as per #2911 and…

shiffman:
Re: https://github.com/processing/processing-docs/issues/238 it seems to me that core library examples are not in the processing-docs repo but actually here so this is where Carnivore client should be removed?  Doing this as a pull request in case I'm wrong.


benfry:
This is @REAS's call. I'm not sure what his plan was.

shiffman:
Thanks, sorry if i'm making extra noise, I had an hour waiting for a home repair and thought "let me fix some stuff if i can!"

benfry:
Nope, not at all. I just wasn't aware why he hadn't done it already, so I wanted to leave it to him.

benfry:
Gonna go ahead and merge so we can clear things out of here. We can roll it back if need be.

Issue #2643, opened 2014-07-21
Processing doesn't work with proxies


jimaobian:
Operation system: Mac OS X 10.9.4
Processing Version: 2.2.1

When using the SOCKS Proxy in Network configuration, Processing cannot start the Application normally.
Once the SOCKS Proxy is Off, Processing can run smoothly

![screen shot](https://cloud.githubusercontent.com/assets/3905677/3640963/dd5a2e34-10ab-11e4-92e3-cadd7c2473ba.png)

Akarshit:
The socks proxy and port that we put into preference.txt is not getting used anywhere(in Preference.java) , i will do PR to fix this and hope that this bug is solved but someone will have to check as i don't have access to mac.

benfry:
It should be getting set in `PApplet` (which is used by the PDE as well).

Akarshit:
@benfry i was not able to find the code setting up socks proxy for the PDE, though i made a PR that sets socks proxy from preference.txt . This got the things working on my system and hopefully would work on others too.

benfry:
Thanks, incorporated for 3.0a8.

PR #3260, opened 2015-05-09
Added socks support in Processing

Akarshit:
Fix for #2643 
This should most probably fix the issue.
@benfry the field to be added to support socks proxy is `socksProxy.host` and `socksProxy.port` in the preference.txt . This should be added somewhere in the documentation and i am not sure where.


benfry:
Thanks. Changed to `proxy.socks.host` and `proxy.socks.port` to fit with other prefs (and help w/ alphabetical ordering).

Issue #3222, opened 2015-04-27
The message "Your sketch has been modified externally..." appears erroneously


Corrado972:
Hello,
while a sketch is running, if I modify and save the sketch the message "Your sketch han been modified externally. Would you like to reload the sketch?" appears and at the beginning of the file the comment " //<>// " is automatically inserted.

This same message also appers if I open and close the dialog "About processing".

Processing 3.0a7-windows64
Windows 7 Pro 64bit

glasscanvas:
Same problem  on Macintosh

benfry:
I can't reproduce this on any machines. Are you using the debugger? What's the exact sequence that causes the problem?

Corrado972:
I start Processing, 
load the sketch, 
modify at least one character 
then save the sketch.

I switch to another application then switch to Processing again: the message appear and in the console window there are rows like (Gold5EsecInfo is the name of the sketch):
Gold5EsecInfo.pde 35
Gold5EsecInfo.pde 9
Gold5EsecInfo.pde 9

If I click "yes", the comment " //<>// " appear on the sketch.

Hope this help

glasscanvas:
Ben,

This happened on the latest alpha 3.07a and Mac Yosemite. I did have the debugger ON.  I had also recently added the GUI builder but the error happened after I had already saved.

Unrelated: I have sketch with a lot of tabs which I mostly work on my iMac 27. But when I work with it on a smaller screen (11" Mac book air) there is no way to reach all the tabs. It might be nice to have a down arrow appear or scroll arrows.  

I saw this reported already, but it is really hard to see the breakpoints in the 3.07a IDE.

Thanks for developing Processing. I can get lots of results without so much development overhead. I mostly use it to process images and create a new image based on the content of the image source. I mostly create splines depending on image gradients, color, etc. and save the result as either bitmaps or SVGs.

Will processing have native GUI elements such as sliders, boxes to enter parameters? I am familiar with the GUI library though.

Thank you very much,
I hope you don't mind the other feedback 

Carlos Caicedo

benfry:
@Corrado972 If you leave the debugger off (and perhaps restart Processing?), does it still cause trouble?

Corrado972:
Ben,
the debugger is off because the menù "Debug" show "Enable Debugger".

Am I right? O there is a different debugger that I can disable? I'm not an expert of Processing.

Thank you

pvrs12:
[...]

PR #3262, opened 2015-05-11
Added buffer to file detection time

pvrs12:
Adds a buffer of 1000 ms to the file change detection to fix a bug where the file modified time which processing recorded was different from that recorded by java. As discussed in #3222


benfry:
Needs to be <= 1000 ms (not <) to fix the OS X case as well, but I'll make that edit myself.

Issue #3248, opened 2015-05-07
Library Manager scrolling issues


shiffman:
I'm noting something that came up in a GSoC discussion with @alignedleft and @prisonerjohn for @Akarshit to possibly work on.

When you first run the library manager, it displays the libraries in your "libraries" folder and then a bit later loads the full list from contributions.txt.  This causes the scrollbar to go a bit haywire and jump around.  It's awkward and we need a more smooth entry into the manager.

benfry:
Thanks for filing this. It's been on the todo list but didn't have a bug assigned. Drives me nuts.

Akarshit:
OK so i had a severe headache trying to fix this :sweat_smile:   
@alignedleft @benfry you can read this for a better understanding of what i did and suggest if more has to be done.
The problem as far as i could understand was that the contribution that were being downloaded were getting added to the list one by one and after each one is added the list was made again from the scratch, that's why the "scrollbar was going haywire" . 
Another problem was that the contribution were being added on different threads, it was working fine in modes and tools as their number is small but in libraries it was breaking as the no of libraries is around 130. I think some thread was executing later that was making the scroll visible to the user and this was looking very ugly.

__What i did -__  

* I made the contributionAdded function not to use threading.
* This made the addition to the list very slow and the time lag was irritating so i modified the code so that the list is only updated once all the contribution have been added. This reduced the time approximately equal to the threading time.
* I made the list invisible while the above is being done, because if the list is visible during this process the layout has to be made and it would be a time consuming task. As such the list before downloading is not useful anyways because it gets changed once the contributions.txt gets dowloaded. The list is made visible once the building of the list is complete.

alignedleft:
@Akarshit Thanks for this. I haven't tested your PR yet, but your description above make sense.

Since we are overhauling the UI of this piece, I think it will make sense to add a loading indicator (such as a spinner) or some other visual feedback to indicate the data is being loaded.

I'm drafting some UI concepts now and will check back with you. Thanks!

Akarshit:
@benfry this can be closed now.

PR #3328, opened 2015-05-28
Fix for the scrolling issue

Akarshit:
Fix for #3248


Akarshit:
Sorry i mixed two PR's, if the progress bar one isn't getting accepted i will modify this one.

benfry:
Can you update this (with just the scrolling changes) and I can apply it?

Akarshit:
@benfry the other branch that is in this PR has already been merged so it won't be a problem. But i would recommend not to merge this as major changes in CM are on the way and then depending on the performance I will fix the issue.

benfry:
Ok, but as much as possible, do those Manager fixes incrementally. If you give me a massive PR, I can't review it, or it takes far more time to do it.

Issue #3154, opened 2015-03-16
Add internationalization support for individual contributions


joelmoniz:
Implementation

This could be done by having something like a `lang` or `languages` folder within a contribution (library/mode/example/tool) which contains files like `properties.de`, `properties.fr`, `properties.es` and so on. Each of these files would need to contain only a sub-set of fields present in the `library.properties` file (for example). These fields would primarily include `name`, `url` (in case the author has his/her webpage available in several languages), `sentence` and `paragraph`. No changes will need to be made to the `contributions.txt`, and this feature won't break a Contribution Manager that does not support internationalization.  

I shall get cracking on this soon (once some of the more pressing bugs in the Contributions Manager are resolved), although I'm not quite sure about its priority.

benfry:
Another earlier implementation: https://github.com/processing/processing/pull/2833

joelmoniz:
Ah thank you. I was under the impression that I had seen something along these lines somewhere, but was unable to find an issue like this. I never realized it was a PR instead that made this idea seem familiar.

federicobond:
Hi there! I experienced problems for weeks trying to run the latest Processing code (it would crash and burn on startup) but I kept trying to build it and somehow I managed to fix it today, so I would be happy to take a look at this and other issues. Let me know if you have already looked into it. I will try to rebase and submit another pull request.

federicobond:
I believe the basic idea still stands: mantain a map of bundles associated with internationalization domains. The default domain is `processing.app` but contributions should be able to register their own via something like `Language.registerDomain` or `Language.registerContribution`.

Unfortunately, the new system makes this a little more difficult, because I cannot use the contribution classloader to load the bundle from the contribution classpath, and the `LanguageBundle` is too coupled to the core language files.

Akarshit:
@alignedleft The implementation i found most easy to implement is to have the <type>.properties file along with language specific files like <type>_<language>.properties. This way the current contributions won't have to be modified at all.  But the problem is that this would only let the contribs that are downloaded to display in the specified language rest of them would be in English as they are taking there data from contributions.txt which is in English.  I have made a PR for the method i described should i summit it ?

benfry:
The most important point here is to be able to use `Language.text()` to handle messages and dialog boxes in things like Tools and Modes. 

[...]

PR #3337, opened 2015-05-31
added i18n to individual contribution

Akarshit:
This PR is according to my comment in #3154


benfry:
See notes in the original report.

PR #3344, opened 2015-06-02
focus is on the list and hint is displayed

Akarshit:


app/src/processing/app/contrib/ContributionManagerDialog.java
```diff
@@ -177,6 +177,7 @@ public void actionPerformed(ActionEvent arg0) {
       contributionListPanel.grabFocus();
```
benfry: Shouldn't this one be removed if we're adding the other down below?

Akarshit: oh yes i removed that, must have forgot to commit.

PR #3345, opened 2015-06-02
Update Monte Media library for MovieMaker

GKFX:
Not many changes had to be made; only mildly controversial thing is a sub-class of `JOptionPane` to enable line-wrapping,  to handle large error messages (which otherwise exceeded the width of my screen).


benfry:
What do we gain with this additional 8,000 lines of code and 50 files? This is a massive change that seems out of proportion with the idea of having a simple movie exporter, and might be better handled as a contributed Tool.

GKFX:
You asked for this to be done in MovieMaker.java, lines 37-39, so I did it. I didn't check the line count before submitting, because I wrongly assumed that the the old and new libraries would be roughly the same size. I've already trimmed out all the files that weren't necessary to make it run (beware that it's possible to build it with less, but then it won't do anything), so it can't be made any smaller without taking stuff out line by line. Making this a separate tool seems like overkill. What do you want to do?

benfry:
Again, what do we gain with this? It's a huge change, and therefore that much more to maintain. Any time we're adding code we have to weigh against what we're going to maintain and what's worthwhile. All of `processing.app` is 88 files. This is 50 just on its own.

GKFX:
You asked for the thing, so I'm trying to guess your reasons for it. I imagine you wanted it to get upstream bug fixes, and to make it easier to bring in further new code from upstream in the future, so that they can do the maintenance. 
If it's too big, then don't accept it; I don't mind either way.

GKFX:
By the way, if you decide against using the new library, please could you consider pulling the l10n commit on its own?

benfry:
It's an offhand comment in the code about how we could update it for the future. I don't need insight into why I made the comment—I want to know if by adding this enormous commit we're actually gaining anything: major current bugs fixed, ongoing support for the future, anything. As I said above, we always have to weigh what we gain versus the overall impact. So with no useful impact, several dozen files and thousands of lines of code don't make sense. It's just bloating the project for what is a very minor feature, and giving me more to maintain in the future. Since you don't have any reasons why this commit helps us, I'll close this one. 

I appreciate the effort, and if you'd like to maintain it, please make a Tool contribution for it so the community can make use of it.

GKFX:
Okay, that's fine. I don't think that many people would want to download a contributed version of this on the grounds that it might contain unspecified bug fixes, so I won't make one. I'll leave the code online so that if some situation *does* appear which forces us to upgrade, it can be done without any fuss, and I'll put the i18n commit in a new pull request for you.

PR #3346, opened 2015-06-02
Update README.md

GKFX:
Code style guidelines are done!


benfry:
Ha! Good catch. Re-adding via strikethrough so we can link to it.

PR #3370, opened 2015-06-10
Mode problems dialog now displays vertically

### LINKED ISSUE #3369: "Mode Problems" dialog does not display properly when an incompatible mode is selected
When an incompatible mode is selected, the error trace on the dialog that appears displays horizontally rather than vertically:

image

joelmoniz:
This resolves #3369, which was occurring because the `message` passed to the `showBadnessTrace()` function had new lines in it that weren't replaced with <br/> tags.


benfry:
Thanks!

PR #3392, opened 2015-06-16
Add missing internationalization in app/Sketch.java

joelmoniz:
This PR adds in a couple of key-value pairs into PDE.properties file from the app/Sketch.java that were hard-coded so far. This should complete the internationalization of app/Sketch.java.

### DIFF
--- a/app/src/processing/app/Sketch.java
+++ b/app/src/processing/app/Sketch.java
@@ -450,17 +450,17 @@ public class Sketch {
     }
 
     if (newName.startsWith(".")) {
-      Base.showWarning("Problem with rename",
-                       "The name cannot start with a period.");
+      Base.showWarning(Language.text("name.messages.problem_renaming"),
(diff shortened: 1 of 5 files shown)


benfry:
Don't all those quotes need to be escaped, the way they are in the rest of the file?

joelmoniz:
I had actually escaped them initially, but then removed the escapes when I realized that most of the file actually contained quotes that weren't escaped. I tested this out as well, and it seems that strings read in from a text file don't need anything to be escaped. I checked this out again just now, and surprisingly, strings with escaped characters when read from a file print the escaping slash as well, so something like `The file \"%s\" already exists.` actually shows up as `The file \"sketch_150618a\" already exists.`, while `The file "%s" already exists.` shows as expected: `The file "sketch_150618a" already exists.`  

NB: I've also updated the strings that had escaping backslashes in them and removed them.

benfry:
Ok, thanks for checking.

PR #3404, opened 2015-06-19
Contributed library's import statements now gets added correctly

### LINKED ISSUE #3403: Trying to import a contributed library adds in a blank import statement
When trying to add import statements corresponding to a contributed library via `Sketch>Import Library...>(contributed library)`, only an import statement of the form `import .*;` gets added.

joelmoniz:
Adding in import statements corresponding to a contributed library via `Sketch>Import Library...>(contributed library)` now adds in the correct import statement(s). The error was because of the relatively new `specifiedImports` field in the `.properties` file, which is blank for most library contributions at present. The import statements were getting read from here, and null was being returned only if the field was null. However, the field was being parsed and stored as an empty string if the field was not present/was blank, which caused an import statement of the form `import (blank string).*` to be added.
  
This resolves #3403.

### DIFF
--- a/app/src/processing/app/contrib/LocalContribution.java
+++ b/app/src/processing/app/contrib/LocalContribution.java
@@ -543,10 +543,12 @@ public abstract class LocalContribution extends Contribution {
   }
   */
   /**
-   * Returns the imports (package-names) for a library, as specified in its library.properties
+   * Returns the imports (package-names) for a library, as specified in its 
+   * library.properties
    * (e.g., imports=libname.*,libname.support.*) 
    * 
-   * @return String[] packageNames (without wildcards) or null if none are specified
+   * @return String[] packageNames (without wildcards) or null if 
+   * none are specified
    */
   public String[] getSpecifiedImports() {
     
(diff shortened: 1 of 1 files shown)


benfry:
We should fix this at the source, not add another band-aid for the symptom. Why is a blank entry ever being added to the list in the first place?

joelmoniz:
Ah, sorry about that. This happens because of the `specifiedImports` field in the library .properties file. It is a relatively new field, and is left blank in most of the contributed libraries, as a result of which a blank entry gets stored. I've now updated the commit to ensure that the blank entry be detected and treated as missing (stored as a null) during the initial parsing of the .properties file itself, instead of handling this check in the getSpecifiedImports() method like I had done till now.

benfry:
It doesn't look like you've updated the PR.

joelmoniz:
The change was originally in `getSpecifiedImports()`, but is now in the correct location `parseImports()`. The fix itself is the same, however, its just that it is now added at the correct place.

benfry:
Fixed through another route, plus other major cleanups in there.

PR #3422, opened 2015-06-23
merged all CM into 1

Akarshit:
This one is huge, don't worry from next time it will be small PRs

* A new class ContributionTab contains the stuff that is visible in a tab.
* ContributionManagerDialog contains the tabbedPane which has ContributionTabs
* ContribtionListing object is present in every ContributionTab object but it will be removed later to include a single object in ContributionManagerDialog so we can get rid of synchronization issues.
* Added shortcuts to the tabs like Alt+1, Alt+2 on linux and windows etc.

waiting for @alignedleft to give this PR a go then @benfry can review and merge it.


app/src/processing/app/contrib/ContributionManagerDialog.java
```diff
@@ -485,129 +511,149 @@ public void setProgress(int value) {
...
+      return updatesContributionTab;
     }
-    filterField.applyFilter();
+    
   }
 
-
-//  private JPanel getPlaceholder() {
-//    return contributionListPanel.statusPlaceholder;
+//
```
benfry: Just use /* block comments */ for large sections like this (that don't already have block comments), so it doesn't mark every single line as changed/rewritten.

benfry:
This needs to be updated with to the latest from the HEAD so that it can be merged.

PR #3432, opened 2015-06-25
StatusPanel created for CM

Akarshit:
@alignedleft @benfry 
I am doing this PR mainly because if i don't do it at this state then the next one will be a huge one. This is kind of incomplete but is the best i could manage without breaking the existing code. I don't expect this to get merged but please review if I am in the right direction. 

How its supposed to look
file1

How it's looking now 
screenshot from 2015-06-26 01 37 03
screenshot from 2015-06-26 01 37 27

Things left
* display the progressbar that hides behind the label showing version.
* do something(or don't do anything) when a contribution that is not compatible is clicked

Doubts
* Suppose i have clicked on a library and its status is showing in the status panel, now when i change the tab(say mode) but haven't clicked on any of the though previously i selected the android mode, should the status update to android mode?
* In the update tab, as according to wireframes there is only a update all button, so what happen when a user simply clicks on a list item?
* In update tab, wouldn't it make more sense to let the update and remove button stay as such and replace the install button with `Update all` button ?
* When the user clicks on `Update all` whose progress bar to show ?
file3
* Are the status messages like "Downloading contributions file...", "Done" to be completely removed because of the spinner ?
* As in a conversation with @joelmoniz there is a undo button which becomes active when a contribution is removed unsuccessfully, where is it(if it is) supposed to be placed.
* @joelmoniz also mention there is a provision for restart button if a mode demands that, what to do with it ?
* When internet is not working and the message having the `Try Again` button is closed, should it appear again anywhere? If so how ?
file2
* If a contribution is clicked that is not compatible should the status be updated ?
* Is the icon supposed to be shown in the status ?
* I have kept the hyperlinks, I suppose that's okay
* And now the main one can someone give me an idea on how to make that list, i mean which layout to use ?


app/src/processing/app/contrib/ContributionListPanel.java
```diff
@@ -97,7 +99,7 @@ private void updatePanelOrdering() {
 
       add(entry.getValue(), c);
     }
-
+/*
```
Akarshit: Why was a status getting added to the end of list ?

app/src/processing/app/contrib/ContributionManagerDialog.java
```diff
@@ -57,13 +58,16 @@
...
+  
+  ContributionListing contributionListing = ContributionListing.getInstance(); 
 
   public ContributionManagerDialog() {
-    toolsContributionTab = new ContributionTab(ContributionType.TOOL);
-    librariesContributionTab = new ContributionTab(ContributionType.LIBRARY);
-    modesContributionTab = new ContributionTab(ContributionType.MODE);
-    examplesContributionTab = new ContributionTab(ContributionType.EXAMPLES);
-    updatesContributionTab = new ContributionTab(null);
+    statusPanel = new StatusPanel(450,this);
```
Akarshit: giving size to statusPanel won't be needed in future

app/src/processing/app/contrib/StatusPanel.java
```diff
@@ -21,34 +21,197 @@
...
+      }
+    });
+    installProgressBar = new JProgressBar();
+    updateLabel = new JLabel("  ");
+    updateButton = new JButton("Update");
+    updateButton.addActionListener(new ActionListener() {
+      
+      @Override
+      public void actionPerformed(ActionEvent e) {
+        StatusPanel.this.contributionManagerDialog.getActiveTab().contributionListPanel.getSelectedPanel().update();
```
Akarshit: is this chain of references supposed to be shortened ?

benfry: Yes, keep things to 80 chars. If it needs two lines, you can do an assignment on one line and then use that variable on the next, rather than the huge chaining stuff. More importantly, if the chain is that long, there's something wrong in the logic, or there needs to be a more sensible set of accessors.

benfry:
If it hasn't broken anything, I'd rather incorporate the PR so that people can see status and we can keep the PRs smaller. At other stages (right before a release) I don't like to do that sort of thing, but we're a few days from a release. When do you think you'll have it in a usable state? I'll probably want to get alpha 11 out on Thursday or Friday (2nd or 3rd of July).

Akarshit:
@benfry I don't think anything is broken right now though there is two of everything, so this can be merged. But i would be very thankful if someone can answer my doubts as in the previous comment(specially the last one).

I would try to make some more change and get them stable before Friday.

app/src/processing/app/contrib/ContributionListPanel.java
benfry: Don't know—I didn't write it. But if it's not needed we can take it out.

benfry:
I'll leave the other questions to @alignedleft, this is his baby.

Akarshit:
@benfry No problem, can shed some light on which layout to use for the main list. It would help me to get building quickly.

benfry:
Hm, maybe just use a `JTable` and some custom cell renderers? 

I'm also looking at using HTML for UIs (via JavaFX, see the "Welcome" class), but that's not ready for prime time.

Akarshit:
@benfry I had a brief look at JavaFX, but I think the decision depends on you should I use `JTable` or JavaFX ?

alignedleft:
Thanks, @Akarshit. This is amazing work.

I am trying to run the current build (as of timestamp on this post), but opening the app just hangs for me. Any suggestions? I would love to preview this for myself.

alignedleft:
> Suppose i have clicked on a library and its status is showing in the status panel, now when i change the tab(say mode) but haven't clicked on any of the though previously i selected the android mode, should the status update to android mode?

Good question. Because when the mode changes, all the items in the CM window would be affected (as it only shows items for the current mode), correct?

Maybe the CM should be a dialog box, i.e. always on top of everything else.

*Or*… say I open the CM window, then go to an editor window and try to switch modes. Should we check to see if the CM window is open, then, if so, check to see if any updates/downloads are in progress? We could warn "switching modes will interrupt this update".  

None of these are great solutions — I just want to get the ideas out there.

> How to show errors that occur while downloading/installing/removing/updating a contribution ( in normal case and in Update all case ?

For example, if the download fails part way through?

> In the update tab, as according to wireframes there is only a update all button, so what happen when a user simply clicks on a list item?

The Updates tab is just for quick updating, not for exploring details. Clicking a list item does nothing. To see details or update items individually, the user has to switch to one of the other tabs.

> In update tab, wouldn't it make more sense to let the update and remove button stay as such and replace the install button with Update all button ?

I see what you mean, but it's confusing, because, in the first four tabs, the lower panel represents information specific to a given contribution.  An "update all" button pertains to more than one contribution, so it would be confusing to have it in the lower panel.

> When the user clicks on Update all whose progress bar to show ?

Good catch. We will have to handle this somehow. Could we have individual progress bars for each updating item?  For example, the area where the version numbers are shown could be replaced by a progress bar.

> Are the status messages like "Downloading contributions file...", "Done" to be completely removed because of the spinner ?

Yes.

> As in a conversation with @joelmoniz there is a undo button which becomes active when a contribution is removed unsuccessfully, where is it(if it is) supposed to be placed.

I don't think we need this undo button. To "undo" the user needs to reinstall the contribution.

> @joelmoniz also mention there is a provision for restart button if a mode demands that, what to do with it ?

Can you find an example of this? Maybe when the mode is downloaded, we could prompt with a dialog to "quit and restart processing".

> When internet is not working and the message having the Try Again button is closed, should it appear again anywhere? If so how ?

For example, if my network connection *was* working, but then is interrupted? Yes, I suppose the "Try again" area should reappear.

> If a contribution is clicked that is not compatible should the status be updated ?

Not sure I understand.

> Is the icon supposed to be shown in the status ?

Yes.  I can get you artwork for each of these. The wireframes just show concepts, not final images.

> I have kept the hyperlinks, I suppose that's okay

I think it's good to keep the links, but let's incorporate this into the design. For example, instead of being redundant (as in the screenshots above) and repeating the name of the library and author, we could include a link below the description to "Library website" or just show the URL, like "www.something.com".

> And now the main one can someone give me an idea on how to make that list, i mean which layout to use ?

Not sure I can help; this sounds like a Java dev question.

benfry:
@Akarshit `JTable` all the way, you probably don't even need a custom cell renderer. Don't worry about the JavaFX stuff at all, it's not clear if we'll be using that or not.

joelmoniz:
@alignedleft 
A quick note about when the restart button was originally used. Initially, the CM required the user to restart the sketch whenever a mode or tool was removed to have the uninstalled contribution removed completely. This was subsequently modified in [this](https://github.com/processing/processing/pull/2705) PR, after which removals happen successfully most of the time. It was only on the off-chance that the class loader that loaded the contribution could not be closed and a restart required to remove the contribution that this button was displayed (this was just for convenience, the user was free to restart later, and this button was subsequently hidden altogether I think).

benfry:
Is that the code that calls Java on the command line? We should pull that out. There's no way that's going to run consistently across all the platforms we support.

Issue #3429, opened 2015-06-25
Error during installation of any contribution


Akarshit:
The file is getting downloaded but not getting installed, i have checked the permissions of the folder also. Is anyone else getting similar error ?

benfry:
I'm seeing that as well.

benfry:
@Akarshit Can you look into this one? It's much higher priority than the status panel, and I have very limited (and expensive) internet connectivity so I can't test downloading things.

PR #3438, opened 2015-06-26
Fixed the NPE

Akarshit:
Fix for #3429

### DIFF
--- a/app/src/processing/app/contrib/AvailableContribution.java
+++ b/app/src/processing/app/contrib/AvailableContribution.java
@@ -270,7 +270,10 @@ public class AvailableContribution extends Contribution {
         specifiedImport = sbImport.toString();
       }
       */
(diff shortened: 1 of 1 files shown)


app/src/processing/app/contrib/AvailableContribution.java
```diff
@@ -270,7 +270,10 @@ public boolean writePropertiesFile(File propFile) {
         specifiedImport = sbImport.toString();
       }
       */
-      String importItem = importsList.join(",");
+      String importItem = "";
+      if(importsList != null){
```
Akarshit: is this supposed to come null in most of the cases ?
Is there a contribution which needs to be tested to see if `importsList` is working fine ?

benfry:
Thanks for checking into this, my goof. I've added a cleaner fix that should do the trick.

PR #3444, opened 2015-07-01
Fixes java mode loading

### LINKED ISSUE #3443: modes extending Java mode should load their parent's JAR
Modes such as REPL and Android extend JavaMode.jar, and thus their loader is unable to resolve this dependency unless JavaMode.jar is added explicitly to the [loader's creation](https://github.com/processing/processing/blob/master/app/src/processing/app/contrib/LocalContribution.java#L164).

Manindra29:
If the NoClassDefFoundError is caught for JavaMode, reinits the urlclassloader. (fixes #3443)

### DIFF
--- a/app/src/processing/app/contrib/LocalContribution.java
+++ b/app/src/processing/app/contrib/LocalContribution.java
@@ -176,6 +176,9 @@ public abstract class LocalContribution extends Contribution {
   }
   
   public String initLoader(String className, URL[] extraUrls) throws Exception {
+    
+    // initLoader(String) can call this with a null URL array
+    
     File modeDirectory = new File(folder, getTypeName());
     if (modeDirectory.exists()) {
       Base.log("checking mode folder regarding " + className);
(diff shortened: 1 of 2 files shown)


benfry:
This is a really hacky approach and only gets it working with Java Mode... 

We already have the idea of the imports entry in the .properties file, a better option would likely be to include the import for Java Mode in there, and then use that to figure out which classes need to be loaded. So you'd include:
```
imports=processing.mode.java
```
in your `mode.properties` file, and we'd iterate through the modes to find the right classes, then read those JARs from the new mode's ClassLoader. 

On a quick glance, it also looks like your approach is stealing the `ClassLoader` of `JavaMode` and polluting it with the subclassing Mode's classes, which is also a bad idea, because it would mean other Modes would be capable of breaking the default mode, and therefore completely hosing Processing.

Manindra29:
@benfry I wasn't aware of the imports field in mode.properties. I've updated the PR as you suggested. Now dependencies are loaded by classname. If mode.properties has:
````
imports=processing.mode.java.JavaMode
````

The installed modes are searched for a class name matching `processing.mode.java.JavaMode`. If such a Mode is found, all of its jars are added to the urlclassloader. This method feels much cleaner.

codeanticode:
Tested the android mode, it works now!

benfry:
Thanks for fixing.

PR #3454, opened 2015-07-05
CM list converted into table

Akarshit:
@alignedleft 
I thought it would be nice to do a PR before b11 is released
Things left to do
* Handle status tab (@benfry do you think i should make a separate layout for update panel or use many ifs to reuse the current one)
* Color-coding
* Ellipses in the contribution name. (@benfry can you please help me in this, see the line note)
* Remove threading(not needed any more)
* Fix that weird bug when python mode is selected from the list.

### DIFF
--- a/app/src/processing/app/contrib/ContributionListPanel.java
+++ b/app/src/processing/app/contrib/ContributionListPanel.java
@@ -40,7 +40,6 @@ import javafx.collections.transformation.SortedList;
 import javax.swing.BorderFactory;
 import javax.swing.GroupLayout;
 import javax.swing.Icon;
-import javax.swing.ImageIcon;
 import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.JScrollPane;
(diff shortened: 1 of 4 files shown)


app/src/processing/app/contrib/ContributionListPanel.java
```diff
@@ -78,51 +108,182 @@ public ContributionListPanel(ContributionTab contributionTab,
...
+          }
+        }
+        label.setIcon(icon);
+        if (isSelected) {
+          label.setBackground(Color.BLUE);
+        }
+        label.setOpaque(true);
+//        return table.getDefaultRenderer(Icon.class).getTableCellRendererComponent(table, icon, isSelected, false, row, column);
+      } else if (column == 1) {
+        JTextPane name = new JTextPane();
```
Akarshit: I can get the ellipses when this column is rendered as a string but then i wont get the contribution name in bold. To get the contribution name in bold i have to use html tags but with html the ellipses do away. So i am using JTextPane to render the name, but still am not getting the ellipses is there a easy way of getting those ellipses without loops and all.

benfry: Can't you call `setFont()` on the `JLabel`?

How does the ellipsis implementation work in `JLabel`?

benfry:
I'd vote for a separate layout, rather than a bunch of `if` statements... The CM code is messy enough as it is.

app/src/processing/app/contrib/ContributionListPanel.java
Akarshit: I don't have to set the font for the whole label just a part of it.

I will open a separate issue for this and look into it later.

PR #3461, opened 2015-07-11
Added a comment in LargePage

tiwawan:
Environment: Ubuntu 14.10
Version: 3.0 alpha 0238

I tried an example of PDF library named "LargePage," and I had difficulty in finding the generated PDF file.
The comment in LargePage says "The file is saved to the sketch folder," but I think this is unfriendly because we usually open examples from "File->Examples..." so we don't care about where the sketch folder is.
Fortunately, we can open the sketch folder by "Sketch->Show Sketch Folder." I added a comment about this.


benfry:
@REAS Your call on this one.

Issue #3440, opened 2015-06-27
CM doubts and status


Akarshit:
@alignedleft  
I am opening this because I think there are more people involved in CM than us so they can also point out if we missed something.

Things left
* ~~display the progressbar that hides behind the label showing version.~~
* ~~Make the error Panel.~~
* do something(or don't do anything) when a contribution that is not compatible is clicked.
* convert the list to table
* handle the special case of update tab

Doubts
* ~~I opened the CM and then clicked on the BootStrap then the statusPanel shows information about BootStrap like this 
![screenshot from 2015-06-28 00 06 08](https://cloud.githubusercontent.com/assets/7762605/8393361/05e4e7f0-1d2a-11e5-8d65-32b2bd41ff5e.png)  
Then I change to `Modes` tab and the satusPanel is still showing the information about BootStrap(as it should ? I think) like
![screenshot from 2015-06-28 00 07 08](https://cloud.githubusercontent.com/assets/7762605/8393383/550fda56-1d2a-11e5-9b2a-398dd9fe1f0c.png)
Now when i select a mode it shows information about the in the statusPanel like 
![screenshot from 2015-06-28 00 07 33](https://cloud.githubusercontent.com/assets/7762605/8393388/6c7d5786-1d2a-11e5-9497-fecdf6669fde.png)
Now when i switch back to 'Tools` tab then the statusPanel is still showing the information about the Android Mode like 
![screenshot from 2015-06-28 00 07 41](https://cloud.githubusercontent.com/assets/7762605/8393391/b678f34a-1d2a-11e5-9918-b1d7cf6d70ed.png)
 So should it show information about Bootstrap or the mode ?~~
_Selection should be displayed in statusPanel_
* What if the user changes modes while CM is open?  
_Good question. Because when the mode changes, all the items in the CM window would be affected (as it only shows items for the current mode), correct?  
Maybe the CM should be a dialog box, i.e. always on top of everything else.  
Or… say I open the CM window, then go to an editor window and try to switch modes. Should we check to see if the CM window is open, then, if so, check to see if any updates/downloads are in progress? We could warn "switching modes will interrupt this update".  
None of these are great solutions — I just want to get the ideas out there._
* How to show errors that occur while downloading/installing/removing/updating a contribution ( in normal case and in `Update all` case ? These error can be of various types like __No net connection__, __poor format of contribution__, __file system error__, etc.
[...]

PR #3465, opened 2015-07-15
Made update classes into one

### LINKED ISSUE #3464: Contribution Manager looking really rough
@Akarshit what's going on here? This is a huge step backwards from the previous version. I'm trying to get alpha 11 out and it can't go out looking like this... cc @alignedleft 

(Also FYI, I moved the icon files around to keep them in line w/ how we handle all other assets. The `icons` directory is for the app's icons.) 

screen shot 2015-07-15 at 10 02 52 am

Akarshit:
Tried to fix #3464 @benfry can you please check that this has been solved and see if there is any other significant bug.

P.S The bug was occurring because i left that ellipses problem for later.


benfry:
No, still looks like this: 

screen shot 2015-07-15 at 8 58 42 pm

I'm not sure if you're trying to follow @alignedleft's sketch too literally (with the weird hyper-plain borders and whatnot) but even the serif type, bad proportions, etc are all wrong.

Akarshit:
@benfry Can you please change the constant 4 [here](https://github.com/processing/processing/blob/master/app/src/processing/app/contrib/ContributionListPanel.java#L283) to a lesser value please. It should fix this. If it doesn't i will try something else

Do you want me to get rid of those border and let the lookAndFeel decide about the looks ?
I thought the the CM had to be made exactly like that with all the borders and everything because @alignedleft asked
> I realize our Java window customization possibilities are limited. How realistic is it to create tabs
> and other elements that look like the wireframes
in #3440.

PR #3496, opened 2015-07-20
Remove duplicate curve vertex

### LINKED ISSUE #2937: curveTightness causes glitches in P2D
Compare the ouptut of the following sketch between P2D and the default renderer:

```java
int counter, frames=180;
float xc1, yc1, xc2, yc2, xc3, yc3;
float theta;
 
void setup() {
  size(500, 500, P2D);
}
 
void draw() {
  background(#4156BC);
  stroke(255);
  strokeWeight(40);
  fill(255, 150);
  float t = map(sin(theta), -1, 1, 1, -1.3);
  float d = width/3;
  curveTightness(t);
  beginShape();
  counter = 0;
  for (int angle = 0; angle < 360; angle +=  60) {
    float xc = cos(radians(angle)) * d + width/2;
    float yc = sin(radians(angle)) * d + height/2;
    curveVertex(xc, yc);
    switch(counter) {
    case 0:
      xc1 = xc;
      yc1 = yc;
      break;
    case 1:
      xc2 = xc;
      yc2 = yc;
      break;
    case 2:
      xc3 = xc;
      yc3 = yc;
      break;
    }
    counter++;
  }
  curveVertex(xc1, yc1);
  curveVertex(xc2, yc2);
  curveVertex(xc3, yc3);
  endShape();
 
  theta += TWO_PI/frames;
  //if (frameCount<=frames) saveFrame("image-###.gif");
}
```

(from http://www.openprocessing.org/sketch/171016)

JakubValtar:
Fixes #2937

### DIFF
--- a/core/src/processing/opengl/PGraphicsOpenGL.java
+++ b/core/src/processing/opengl/PGraphicsOpenGL.java
@@ -13083,7 +13083,7 @@ public class PGraphicsOpenGL extends PGraphics {
       }
       if (stroke) addStrokeVertex(x, y, z, strokeColor, strokeWeight);
 
-      for (int j = 0; j < pg.curveDetail; j++) {
+      for (int j = 0; j < pg.curveDetail - 1; j++) {
         x += xplot1; xplot1 += xplot2; xplot2 += xplot3;
         y += yplot1; yplot1 += yplot2; yplot2 += yplot3;
         z += zplot1; zplot1 += zplot2; zplot2 += zplot3;


benfry:
So this is the sort of thing you need to special case so that you can close curves properly. It's not enough to just ignore the last vertex—you need it there as an indication that the curve should close—but the GL renderer needs to take it into account (internally) and not double up the point to cause the problems seen in #2937.

JakubValtar:
There is currently no way to make a loop which does not have stroke caps, just stroke joints. If you call `endShape(CLOSE)`, the loop is indeed closed, but you get straight line between last vertex and the first vertex.

JakubValtar:
If the first and the last vertex happen to be at the same position, stroke starts to glitch.

benfry:
That's what I'm saying: when the first and last are the same position, the renderer needs to special case that situation to have it properly close and not glitch. Internally it'll need to skip the last point to avoid the glitches, but from a user's point of view (and the API's point of view), the line should behave as if it's closed.

JakubValtar:
However, when you draw just vertices (straight lines) you don't draw the first one again at the end, `endShape(CLOSE)` will connect them. When you draw curves though, you want to draw the last one because there are multiple ways how the curve can go. Is this inconsistency something we can live with, or are we going to change how the simple lines work too?

Also, @codeanticode pointed out that when you want to smoothly close `curve` you need to repeat first three curve vertices again at the end and that we could provide some means to close it easily since we already have all the curve vertices. In theory you just need to supply a bunch of ordered vertices which the curve should go through and we can automatically make a curve loop out of that.

Issue #3492, opened 2015-07-20
mode properties are not properly downloaded by the CM


codeanticode:
When trying to install the latest release of the Android mode using the CM in 3.0 alpha 11, I get the following message

"This item is not compatible with this version of Processing"

after downloading, so the mode does not get installed. f you download the mode package manually you can see that the properties file reads:

```text
name = Android Mode
authorList = [The Processing Foundation](http://android.processing.org/)
url = https://github.com/processing/processing-android
sentence = Create projects with Processing for Android devices
paragraph = This version of the Android Mode is for Processing 3.0a11+
imports=processing.mode.java.JavaMode
version = 239
prettyVersion = 3alpha
minRevision = 237
maxRevision = 0
```

However, the mode files left in a temporary subfolder after the failed installation contains a slightly different properties file:

```text
name=Android Mode
category=Unknown
authors=[The Processing Foundation](http://android.processing.org/)
url=https://github.com/processing/processing-android
sentence=Create projects with Processing for Android devices
paragraph=This version of the Android Mode is for Processing 3.0a11+
version=239
prettyVersion=3alpha
lastUpdated=0
minRevision=237
maxRevision=0
```

In particular, the imports field is missing.

alignedleft:
@Akarshit can you look at this?

Akarshit:
@alignedleft looking into it

Akarshit:
The properties file doesn't contain the imports because of [this](https://github.com/processing/processing/blob/master/app/src/processing/app/contrib/AvailableContribution.java#L352). It looks like this is by design so mabe @benfry can help me with this.  
Regarding why installation is not working i am still working on it.

Akarshit:
I am unable to figure out exactly why this is happening. But [this](https://github.com/processing/processing/blob/master/app/src/processing/app/contrib/ModeContribution.java#L87) line is throwing Exception. But the loader is pointing to the jar file inside the temporary folder that is getting created and the className is the correct name of the class in binary format. @benfry if you could help me with this or point me in the right direction it would be really great.

codeanticode:
@Akarshit we faced a similar problem when the PDE tried to load the mode (once it was already installed). See [this PR](https://github.com/processing/processing/pull/3444) for the details.

kreidler:
I do have the same issue on W7Pro 32. Not only Android mode gets "This item is not compatible with this version of Processing", Any new mode fails to be installed.

[...]

PR #3499, opened 2015-07-21
imports added to mode

Akarshit:
Fix for #3492


app/src/processing/app/contrib/AvailableContribution.java
```diff
@@ -348,7 +348,7 @@ public boolean writePropertiesFile(File propFile) {
         writer.println("lastUpdated=" + lastUpdated);
         writer.println("minRevision=" + minRev);
         writer.println("maxRevision=" + maxRev);
-        if (getType() == ContributionType.LIBRARY && importsList != null) {
+        if ((getType() == ContributionType.LIBRARY || getType() == ContributionType.MODE) && importsList != null) {
```
Akarshit: Why is imports getting added to only library and modes, do example need imports ?

PR #3506, opened 2015-07-27
Assortment of patches from my arm-3.0 branch

gohai:


gohai:
Rebased to current master so that it applies cleanly

gohai:
@benfry How is this looking? (I hope <target if="" does what I think it does)

gohai:
@benfry Something like this? I hope `<target if=""` does what I think it does..

build/build.xml
```diff
@@ -24,6 +24,21 @@
...
+  <!-- detetect Raspberry Pi and friends -->
+  <condition property="linux-arm32" value="linux-arm32">
+    <and>
+      <equals arg1="${platform}" arg2="linux" />
+      <equals arg1="${os.arch}" arg2="arm" />
+      <equals arg1="${sun.arch.data.model}" arg2="32" />
+    </and>
+  </condition>
+
+  <target name="check-linux-arm32" if="linux-arm32">
```
benfry: I think you still need an `antcall` or another target that depends on this one so that it's called properly. (Or am I missing that elsewhere in this PR?)

build/build.xml
```diff
@@ -219,10 +216,22 @@
 
   <target name="jre-download" depends="jre-check, downloader-setup"
           unless="jre.tgz.downloaded">
+    <!-- we normally download a JRE, except on arm -->
+    <property name="jre.downloadJdk" value="false" />
```
benfry: Every variable in build.xml uses dots, not camel case. Let's keep it that way.

core/build.xml
```diff
@@ -39,13 +44,16 @@
 	   encoding="UTF-8"
 	   includeAntRuntime="false"
 	   debug="true"
-	   srcdir="src" destdir="bin"
```
benfry: Why is this removed?

build/build.xml
gohai: Made linux-build depend on his target below

core/build.xml
```diff
@@ -39,13 +44,16 @@
...
+	   destdir="bin"
 	   classpath="apple.jar; 
                       library/jogl-all.jar; 
                       library/gluegen-rt.jar"
 	   nowarn="true">
       <!-- kinda gross, but if not using the JDT, this just ignored anyway -->
       <compilerclasspath path="${jdt.jar}; ../java/mode/jdtCompilerAdapter.jar" />
+      <src path="src"/>
+      <include name="processing/**"/>
+      <exclude name="processing/javafx/**" if="fx.missing"/>
```
benfry: Let's call it `fx.unavailable`. And `if` can be used here? (Didn't know that...)

build/build.xml
gohai: Will change!

core/build.xml
gohai: Will rename. Didn't know that either - but it works even on my prehistoric ant on the Pi, which doesn't support the regular `<if>` elements.

PR #3508, opened 2015-07-27
Prevent flickering #3473

### LINKED ISSUE #3473: OpenGL sketch flickers when draw() is missing or empty
Looks like double buffering issue.

```java
size(200, 200, P2D);
background(0);
fill(255);
rect(10, 10, 80, 80);
```

```java
void setup() {
  size(200, 200, P2D);
  background(0);
  fill(255);
  rect(10, 10, 80, 80);
}
```

```java
void setup() {
  size(200, 200, P2D);
  background(0);
  fill(255);
  rect(10, 10, 80, 80);
}

void draw() { }
```

JakubValtar:
Makes sure both textures have the same content after setup().

Fixes #3473

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -2424,6 +2424,12 @@ public class PApplet implements PConstants {
 
     if (frameCount != 0) {
       handleMethods("post");
+    } else {
+      if ((P2D.equals(renderer) || P3D.equals(renderer))) {
+        // Makes sure both textures have the same content
+        // after setup and there is no flickering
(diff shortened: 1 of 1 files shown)


benfry:
This should be handled in `PGraphicsOpenGL` instead, otherwise it won't work for subclasses, and we should keep as much renderer-specific code out of `PApplet` as possible (witness the current disentangling of `PGraphicsJava2D` in 3.x).

JakubValtar:
Yes, that's a better idea. I'll fix that and send another pull request.

PR #3514, opened 2015-07-29
Small bug fix in  createInputRaw(String filename) . 

poqudrof:
It could no longer load data from Jars. 

I had some Class cast exception, the URLs were  JarURLConnection instead of HttpURLConnection. I figured out a solution by checking the type of the  URLConnection. 
Cheers, 

Jeremy.


benfry:
Thanks for looking into it. For future PRs, please mind the [style guidelines](https://github.com/processing/processing/wiki/Style-Guidelines) so I can accept the PR directly.

Issue #3483, opened 2015-07-19
P2D, P3D on Linux causes IllegalStateException: size() cannot be used here


hamoid:
```
void setup() {
  size(600, 600, P2D);
}
```

On 3.0a10 and 3.0a11 I get this error:

```
Display 1 does not exist, using the default display instead.
java.lang.RuntimeException: com.jogamp.opengl.GLException: Caught IllegalStateException: size() cannot be used here, see https://processing.org/reference/size_.html on thread main-Display-.x11_:0-1-EDT-1
	at jogamp.newt.DefaultEDTUtil.invokeImpl(DefaultEDTUtil.java:249)
	at jogamp.newt.DefaultEDTUtil.invoke(DefaultEDTUtil.java:163)
	at jogamp.newt.DisplayImpl.runOnEDTIfAvail(DisplayImpl.java:427)
	at jogamp.newt.WindowImpl.runOnEDTIfAvail(WindowImpl.java:2133)
	at jogamp.newt.WindowImpl.setVisible(WindowImpl.java:1024)
	at jogamp.newt.WindowImpl.setVisible(WindowImpl.java:1029)
	at com.jogamp.newt.opengl.GLWindow.setVisible(GLWindow.java:499)
	at processing.opengl.PSurfaceJOGL.setVisible(PSurfaceJOGL.java:356)
	at processing.core.PApplet.showSurface(PApplet.java:10179)
	at processing.core.PApplet.runSketch(PApplet.java:10163)
	at processing.core.PApplet.main(PApplet.java:9899)
```

I'm running 64bit ArchLinux with recent Intel graphics. Maybe someone could try on Ubuntu?

On 2.2.1 and 3.0a5 it works fine.

monkstone:
Interesting same code on my 64bit ArchLinux box NVIDIA graphics runs OK still get the 1st message
```
Display 1 does not exist, using the default display instead.
```

benfry:
The problem here is that the window manager is messing with the window size after we've opened the window, so `size()` thinks you're trying to call it again. We've had to solve this many different ways in the past, and the method changed a little in alpha 11, so someone needs to rewrite that code a bit.

hamoid:
For what it's worth, I'm using the i3 tiling window manager. I configured it so Processing sketches are "floating windows", so i3 does not force a size on them. I hope it still works on alpha 11 and newer :)

JakubValtar:
[...]

Issue #3412, opened 2015-06-20
Error on size() when using FX2D due to stage inset issues


aengelke:
This sketch:

```
void setup() {
  size(400, 300, FX2D);
}
void draw() {
  background(255, 0, 0);
}
```

This error when running (a window appears): `size() cannot be used here ...`

benfry:
FX2D isn't finished yet or announced and is broken in many ways. 

In this case, the window's insets aren't properly being handled when sizing the FX Stage.

benfry:
Working properly in the current beta.

PR #3517, opened 2015-07-29
Remove size() from setup() when copying to settings()

JakubValtar:
Possibly solves: #3412 #3483


benfry:
Can you follow up on those to see if this fixes things properly?

JakubValtar:
Will do.

JakubValtar:
@benfry I think for #3412 we should first check if the window is sized properly, because this could just mask the issue in FX2D.

PR #3523, opened 2015-07-31
Remove alpha filler

JakubValtar:
No longer needed because primary surface is now ARGB

https://github.com/processing/processing/commit/f3b72c709cc1db96edf06176cd39eee2c233127b


benfry:
Are you absolutely certain we can do this? I've had to fix and re-fix this so many times that I've lost track.

JakubValtar:
It's not needed as long as the surface has 4 bands (colors + alpha). Maybe we can leave it here as a safety measure? I wonder how this affects the performance (probably not at all).

benfry:
Fingers crossed...

PR #3526, opened 2015-08-02
Second bunch of arm patches

gohai:
Hi @benfry! Here are more patches for you to look at - the most controversial one is probably the one that introduces the armv6hf target for the exporter. This could probably be cleaned up, but I thought I do it pretty mechanically first to not break anything for the other architectures. The difficulty is that we have to give up on the notion of bits (64 vs 32) and instead introduce the idea of a "variant". I though about calling it "arch"[itecture], but I decided against it, since the JVM returns "x86" also for 64-bit Intel Processors. The other thing that makes this a bit more ugly than it needs to is that we want to keep the directory name "linux64" and "linux32", but instead of "linuxarmv6hf" I went for "linux-armv6hf" - so a few more special cases.

@codeanticode Does the GL2ES2 part and the custom paths for our renderers look okay to you?

Thanks!


codeanticode:
@gohai yes, they look good to me. Limiting the number of lights to 2 is a temporary solution until we find a better way to handle the limitations of current drivers, right?

I was testing to switch to GL2ES2 on desktop, but then the GLSL version gets bumped and I need to make some additional changes to the shader code. Was this a problem on the Rpi?

gohai:
@codeanticode Yes, temporary measure - although I think the old binary driver will probably never be updated to fix this. When the new one is clearly better we could just drop support for the former, but until then it would help us trying to get it up on both, and keep our options open.

The shaders worked as-is for me - but I only tried on the Pi. I was under the impression that it wouldn't make a difference on desktop systems where a GL2 profile will be selected, but perhaps I am wrong. If it's a lot of work on your side perhaps we could only select it on arm & Linux for now to reduce fallout?

Thanks!

benfry:
Ouch, messy indeed... Why not make a new platform called armv6hf and never distinguish between 32- and 64-bit? Technically it's Linux, but there's no need for it to be identified as "Linux" for the way that we use the platform code.

gohai:
I considered that, but came to the conclusion that there are a couple of compelling reasons to keep it under the LINUX umbrella:
* it doesn't get *worse* then what is there now - as long as we load the native libraries from our own subdirectory we can just pass as regular Linux the rest of the times, no special cases to add in random other parts of the codebase!
* passing as regular Linux gives us better test coverage and we don't have to update libraries that make use of PApplet.platform
* there will be more ARM micro-architectures coming along that we'd want to support (perhaps armv7hf and of course [armv8](https://jdk8.java.net/download.html) - the 64-bit architecture one, that will soon also be on everyone's desk)
* the same reasons for keeping 64-bit x86 under LINUX would also apply here

The fixes that I need for arm weren't isolated to its platform jar, unfortunately, so I though it might be cleaner to try to move this all at once to a current git version - rather than trying to do a franken build. I've been told that the official 2.3.2 should be fairly soon, and I would push an update that gets us to a stable release as soon as that happens.

benfry:
Hm, not particularly compelling over keeping ARM as its own island. But I'll sleep on it.

app/src/processing/app/Base.java
```diff
@@ -98,6 +98,14 @@
     }
   }
 
+  static String nativeArch;
+  static {
+    nativeArch = System.getProperty("os.arch");
+    if (nativeArch == null) {
+      nativeArch = "";
```
benfry: Yuck, no! `null` means something here, we should be testing for that rather than doing an empty string compare.

benfry:
And we'll wait for the official JOGL 2.3.2 binaries. Every time we update binaries it permanently adds size to the repo, so it needs to be done sparingly.

PR #3527, opened 2015-08-03
Feature: depth reading

JakubValtar:
Adds new hint ENABLE_DEPTH_READING, which will cause multisampled PGraphics to also downsample their depth and stencil buffers (apart from color buffer).

### DIFF
--- a/core/src/processing/core/PConstants.java
+++ b/core/src/processing/core/PConstants.java
@@ -514,5 +514,8 @@ public interface PConstants {
   static final int ENABLE_STROKE_PURE         =  9;
   static final int DISABLE_STROKE_PURE        = -9;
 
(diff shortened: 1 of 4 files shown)


benfry:
Isn't `DEPTH_READING` a misnomer here when it also applies to stencil buffers?

JakubValtar:
Suggestions?

benfry:
Something that makes more sense. :-)

benfry:
Changed to `ENABLE_BUFFER_READING`, smart guy.

JakubValtar:
Sounds good (in spite of the PGL function being called `readPixels`).

benfry:
No second-guessing if you don't offer anything useful.

PR #3530, opened 2015-08-04
Second bunch of arm patches (v2)

gohai:
This is https://github.com/processing/processing/pull/3526 but without the JOGL update, as to not grow the git repo too much.

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -98,6 +98,14 @@ public class Base {
     }
   }
 
+  static String nativeArch;
+  static {
+    nativeArch = System.getProperty("os.arch");
+    if (nativeArch == null) {
+      nativeArch = "";
+    }
+  }
+
   static private boolean commandLine;
 
   // A single instance of the preferences window
@@ -1538,6 +1546,10 @@ public class Base {
    */
   static public int getNativeBits() {
     return nativeBits;
+  }
+
+  static public String getNativeArch() {
+    return nativeArch;
   }
 
 
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -101,9 +101,6 @@ public class Base {
   static String nativeArch;
   static {
     nativeArch = System.getProperty("os.arch");
-    if (nativeArch == null) {
-      nativeArch = "";
-    }
   }
 
   static private boolean commandLine;
@@ -1549,6 +1546,9 @@ public class Base {
   }
 
   static public String getNativeArch() {
+    if (nativeArch == null) {
+      throw new RuntimeException("os.arch is not set");
+    }
     return nativeArch;
   }
 
--- a/core/src/processing/opengl/PGraphicsOpenGL.java
+++ b/core/src/processing/opengl/PGraphicsOpenGL.java
@@ -7046,9 +7046,15 @@ public class PGraphicsOpenGL extends PGraphics {
     if (smooth < 1) {
       pgl.disable(PGL.MULTISAMPLE);
     } else {
-      pgl.enable(PGL.MULTISAMPLE);
+      // work around runtime exceptions in Broadcom's VC IV driver
+      if (false == OPENGL_RENDERER.equals("VideoCore IV HW")) {
+        pgl.enable(PGL.MULTISAMPLE);
+      }
     }
-    pgl.disable(PGL.POLYGON_SMOOTH);
+    // work around runtime exceptions in Broadcom's VC IV driver
+    if (false == OPENGL_RENDERER.equals("VideoCore IV HW")) {
+      pgl.disable(PGL.POLYGON_SMOOTH);
+    }
 
     if (sized) {
 //      reapplySettings();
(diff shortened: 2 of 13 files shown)


gohai:
@benfry Thanks for the review!

gohai:
@benfry I added a RuntimeException for when os.arch is unset - hope this looks good to you!

benfry:
Mmm... not in a `static` block. There's no deterministic way to know when that code will be called, so the error may or may not ever present to the user, but will break the calling thread. So if it's actually an error condition, we should explicitly detect it somewhere that we can notify the user about next steps. 

For instance, if it'll prevent the PDE from working, it needs to happen on startup (like when the Mode is initialized). If it prevents only Export from working, then it can be an `Exception` thrown during the build process.

gohai:
The getter shouldn't be called from static - the current call sites are JavaBuild.exportApplication(), Runner.Runner(), Commander.Commander(), so I believe an exception there would be reasonable?

gohai:
@benfry I believe this should be in a good state now - let me know if you spot any other things that hurt your eyes. If it's easier to review I can also merge those two follow-up patches that arose out of the discussion, and send a new pull request with them included in-line? Thanks!

(Also: yes to jokes about esoteric technical details!)

core/src/processing/opengl/PSurfaceJOGL.java
```diff
@@ -170,7 +170,8 @@ protected void initGL() {
     if (profile == null) {
       if (PJOGL.profile == 2) {
         try {
-          profile = GLProfile.getGL2ES1();
+          // request at least GL2 or GLES2
+          profile = GLProfile.getGL2ES2();
```
benfry: @codeanticode Can you confirm this won't cause trouble?

core/src/processing/opengl/PGraphicsOpenGL.java
```diff
@@ -7158,14 +7164,27 @@ protected void getGLParameters() {
...
       pgl.getFloatv(PGL.MAX_TEXTURE_MAX_ANISOTROPY, floatBuffer);
       maxAnisoAmount = floatBuffer.get(0);
     }
 
+    // overwrite the default shaders with vendor specific versions
+    // if needed
+    if (OPENGL_RENDERER.equals("VideoCore IV HW") ||    // Broadcom's binary driver for Raspberry Pi
+      OPENGL_RENDERER.equals("Gallium 0.4 on VC4")) {   // Mesa driver for same hardware
+        defLightShaderVertURL =
+          PGraphicsOpenGL.class.getResource("/processing/opengl/shaders/quirks/vc4-LightVert.glsl");
```
benfry: Let's make this `LightVert-vc4.glsl` and keep it in the main directory, unless @codeanticode has a preference otherwise.

app/src/processing/app/Base.java
```diff
@@ -98,6 +98,11 @@
     }
   }
 
+  static String nativeArch;
```
benfry: `static String nativeArch = System.getProperty("os.arch");`

java/src/processing/mode/java/Commander.java
```diff
@@ -280,14 +281,16 @@ public Commander(String[] args) {
...
-//                platformBits = Base.getNativeBits();
-//              }
-//              if (platformBits == 0 &&
-//                Library.hasMultipleArch(platform, build.getImportedLibraries())) {
-//                complainAndQuit("This sketch can be exported for 32- or 64-bit, please specify one.", true);
-//              }
-              success = build.exportApplication(outputFolder, platform, platformBits, embedJava);
+
+                String variant;
+                if (platform == PConstants.LINUX && platformBits == 32 && Base.getNativeArch().equals("arm")) {
```
benfry: This can be `"arm".equals(Base.getNativeArch())` so we can skip the `RuntimeException`.

java/src/processing/mode/java/JavaBuild.java
```diff
@@ -787,17 +787,24 @@ protected boolean exportApplication() throws IOException, SketchException {
         if (Library.hasMultipleArch(platform, importedLibraries)) {
           // export the 32-bit version
           folder = new File(sketch.getFolder(), "application." + platformName + "32");
-          if (!exportApplication(folder, platform, 32, embedJava && Base.getNativeBits() == 32)) {
+          if (!exportApplication(folder, platform, "32", embedJava && Base.getNativeBits() == 32 && Base.getNativeArch().equals("x86"))) {
```
benfry: Same here... `"x86".equals(Base.getNativeArch())`

app/src/processing/app/Base.java
```diff
@@ -1540,6 +1545,13 @@ static public int getNativeBits() {
     return nativeBits;
   }
 
+  static public String getNativeArch() {
+    if (nativeArch == null) {
```
benfry: `return System.getProperty("os.arch");`

java/src/processing/mode/java/runner/Runner.java
```diff
@@ -90,9 +90,19 @@ public Runner(JavaBuild build, RunnerListener listener) throws SketchException {
 
     // Make sure all the imported libraries will actually run with this setup.
     int bits = Base.getNativeBits();
+    String arch = Base.getNativeArch();
```
benfry: Rather than doing this a second time here, let's do methods in Base:
```
static public String getVariant(String arch, int bits) {
  // all the logic goes here
}

static public String getVariant() {
  return getVariant(getNativeBits(), getNativeArch());
}
```

gohai:
@benfry I think I addressed all good points you raised and send a new pull request with all the changes folded in (#3555)

PR #3540, opened 2015-08-06
Fix bug with tab sorting when adding new tabs

### LINKED ISSUE #3099: after creating a tab, must switch to another before opening the new tab
When you create a new Tab you have to switch to another Tab before you can open the one you just created.

aengelke:
Fixes #3099 

There is also a detailed explanation of what's going on.


benfry:
Thanks for the fix!

Issue #3534, opened 2015-08-05
Deleted tab still present in tab menu


JakubValtar:
1. Create a tab
2. Open tab menu (button with small arrow) and delete the new tab
3. Open tab menu again, deleted tab is still present
![tab](https://cloud.githubusercontent.com/assets/3177098/9099827/6f6b6eb0-3ba2-11e5-87ac-3778aa3cbf1c.png)

benfry:
Same problem as #3099 (the toolbar is not refreshed properly).

PR #3541, opened 2015-08-06
Rebuild tab menu on delete

aengelke:
Fixes processing/processing#3534


benfry:
Fixed with https://github.com/processing/processing/pull/3542

PR #3563, opened 2015-08-09
tweakmode: place tweak init function right after the size function

### LINKED ISSUE #3562: code runs in Run but not in Tweak
The following code runs in 'Run'.
Trying the same code with 'Tweak' produce ArralyIndexOutOfBoundsException on 'val[0] = 10;'

```java
int[] vals;

void setup()
{
  size(60, 60);
  function_from_setup();
}

void function_from_setup()
{
  vals = new int[4];
  vals[0] = 10;
  vals[1] = 20;
  vals[2] = 30;
  vals[3] = 40; 
}

void draw()
{
  background(0);
  stroke(255);
  for (int i=0; i<4; i++) {
    line(10, vals[i], 50, 50);
  }
}
```

The problem is that recently tweakmode started to place its init functions at the end of setup instead of at the beginning, because of a requirement to put size as the first call in the sketch. Therefore the value '4' don't get initialized before calling the problematic line above.

Because size must be the first line of code in the sketch, the solution could be to place the init calls right after size.

For reference, this is the code produced by automateSketch:

```java
/*************************/
/* MODIFIED BY TWEAKMODE */
/*************************/


import java.net.*;
import java.io.*;
import java.nio.*;

int[] tweakmode_int = new int[16];
TweakModeServer tweakmode_Server;
void tweakmode_initAllVars() {
  tweakmode_int[0] = 4;
  tweakmode_int[1] = 0;
  tweakmode_int[2] = 10;
  tweakmode_int[3] = 1;
  tweakmode_int[4] = 20;
  tweakmode_int[5] = 2;
  tweakmode_int[6] = 30;
  tweakmode_int[7] = 3;
  tweakmode_int[8] = 40;
  tweakmode_int[9] = 0;
  tweakmode_int[10] = 255;
  tweakmode_int[11] = 0;
  tweakmode_int[12] = 4;
  tweakmode_int[13] = 10;
  tweakmode_int[14] = 50;
  tweakmode_int[15] = 50;
}

void tweakmode_initCommunication() {
 tweakmode_Server = new TweakModeServer();
 tweakmode_Server.setup();
 tweakmode_Server.start();
}






int[] vals;

void setup()
{
  size(60, 60);

  function_from_setup();

 tweakmode_initAllVars();
 tweakmode_initCommunication();

}

void function_from_setup()
{
  vals = new int[tweakmode_int[0]];
  vals[tweakmode_int[1]] = tweakmode_int[2];
  vals[tweakmode_int[3]] = tweakmode_int[4];
  vals[tweakmode_int[5]] = tweakmode_int[6];
  vals[tweakmode_int[7]] = tweakmode_int[8]; 
}

void draw()
{
  background(tweakmode_int[9]);
  
  stroke(tweakmode_int[10]);
  for (int i=tweakmode_int[11]; i<tweakmode_int[12]; i++) {
    line(tweakmode_int[13], vals[i], tweakmode_int[14], tweakmode_int[15]);
  }
  
}class TweakModeServer extends Thread
{
  protected DatagramSocket socket = null;
  protected boolean running = true;
  final int INT_VAR = 0;
  final int FLOAT_VAR = 1;
  final int SHUTDOWN = 0xffffffff;
  public TweakModeServer() {
    this("TweakModeServer");
  }
  public TweakModeServer(String name) {
    super(name);
  }
  public void setup()
  {
    try {
      socket = new DatagramSocket(12345);
      socket.setSoTimeout(250);
    } catch (IOException e) {
      println("error: could not create TweakMode server socket");
    }
  }
  public void run()
  {
    byte[] buf = new byte[256];
    while(running)
    {
      try {
        DatagramPacket packet = new DatagramPacket(buf, buf.length);
        socket.receive(packet);
        ByteBuffer bb = ByteBuffer.wrap(buf);
        int type = bb.getInt(0);
        int index = bb.getInt(4);
        if (type == INT_VAR) {
          int val = bb.getInt(8);
          tweakmode_int[index] = val;
        }
        else         if (type == SHUTDOWN) {
          running = false;
        }
      } catch (SocketTimeoutException e) {
        // nothing to do here just try receiving again
      } catch (Exception e) {
      }
    }
    socket.close();
  }
}
```

I'll fix it and file a PR

galsasson:
This PR fixes #3562.

TweakMode init functions:

```java
tweakmode_initAllVars();
tweakmode_initCommunication();
```

are placed now right after the call to size.

In addition, the parser now have a function ignore list. Right now it ignores all numbers in the 'settings' and 'setup' functions.

### DIFF
--- a/java/src/processing/mode/java/JavaEditor.java
+++ b/java/src/processing/mode/java/JavaEditor.java
@@ -2898,8 +2898,8 @@ public class JavaEditor extends Editor {
       return false;
     }
 
-    int setupEndPos = SketchParser.getSetupEnd(baseCode[0]);
-    if (setupEndPos < 0) {
+    int afterSizePos = SketchParser.getAfterSizePos(baseCode[0]);
+    if (afterSizePos < 0) {
       return false;
     }
 
@@ -2992,12 +2992,14 @@ public class JavaEditor extends Editor {
 
     // add call to our initAllVars and initOSC functions
     // from the setup() function.
-    String addToSetup = "\n"+
-      " tweakmode_initAllVars();\n"+
-      " tweakmode_initCommunication();\n\n";
+    String addToSetup = "\n\n\n"+
+      "  /* TWEAKMODE */\n"+
+      "    tweakmode_initAllVars();\n"+
+      "    tweakmode_initCommunication();\n"+
+      "  /* TWEAKMODE */\n\n";
 
-    setupEndPos = SketchParser.getSetupEnd(c);
-    c = replaceString(c, setupEndPos, setupEndPos, addToSetup);
+    afterSizePos = SketchParser.getAfterSizePos(c);
+    c = replaceString(c, afterSizePos, afterSizePos, addToSetup);
 
     // Server code defines a class, so it should go later in the sketch
     String serverCode =
--- a/java/src/processing/mode/java/tweak/SketchParser.java
+++ b/java/src/processing/mode/java/tweak/SketchParser.java
@@ -39,7 +39,8 @@ public class SketchParser {
 
   List<List<Range>> scientificNotations;
 
-  Range setupFunction;
+  // currently is used to ignore numbers in 'setup' and 'settings' functions
+  List<Range> ignoreFunctions;
 
   List<List<Range>> commentBlocks;
   List<int[]> curlyScopes;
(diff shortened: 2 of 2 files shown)


benfry:
In for beta 2—just under the wire!

galsasson:
Awesome, thanks!

Issue #3505, opened 2015-07-27
Remove support for fixed-function pipeline


codeanticode:
Right now the [default profile](https://github.com/processing/processing/blob/master/core/src/processing/opengl/PSurfaceJOGL.java#L173) in the GL renderer is [GL2ES1](https://jogamp.org/deployment/jogamp-next/javadoc/jogl/javadoc/com/jogamp/opengl/GL2ES1.html), so users still have the chance to access the fixed-function pipeline (FFP).

However, with the [deprecation of FFP](https://www.opengl.org/wiki/History_of_OpenGL#Deprecation_Model) already being several years old,  and GLSL programming becoming more widespread, there is little reason to keep FFP around. The GL renderer does not use it at all, since we run our own shaders to handle all rendering operations, including lighting and texturing. We do incur in some redundancy by not removing it, since GL2ES1 still handles the FFP uniforms (transformation matrices, lighting parameters), but Processing does not use them.

benfry:
So is your plan to remove it? Or are you asking whether you should? (FWIW, I definitely say remove it.)

codeanticode:
Mainly the second, I wanted to get some feedback before going ahead.

JakubValtar:
Are we done here?

benfry:
That's for @codeanticode to answer.

PR #3566, opened 2015-08-09
Second bunch of arm patches (v4)

gohai:
Rebased against master, supersedes #3555.

### DIFF
--- a/core/src/processing/opengl/PSurfaceJOGL.java
+++ b/core/src/processing/opengl/PSurfaceJOGL.java
@@ -195,7 +195,8 @@ public class PSurfaceJOGL implements PSurface {
     if (profile == null) {
       if (PJOGL.profile == 2) {
         try {
-          profile = GLProfile.getGL2ES1();
+          // request at least GL2 or GLES2
+          profile = GLProfile.getGL2ES2();
         } catch (GLException ex) {
           profile = GLProfile.getMaxFixedFunc(true);
         }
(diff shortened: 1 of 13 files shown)


core/src/processing/opengl/PSurfaceJOGL.java
```diff
@@ -195,7 +195,8 @@ protected void initGL() {
     if (profile == null) {
       if (PJOGL.profile == 2) {
         try {
-          profile = GLProfile.getGL2ES1();
+          // request at least GL2 or GLES2
+          profile = GLProfile.getGL2ES2();
```
benfry: @codeanticode If you can confirm we're safe to do this, I can merge this patch.

gohai: So much pressure.. :)

codeanticode: bumping the profile to GL2ES2 requires upgrading the shader code to [v330](https://github.com/mattdesl/lwjgl-basics/wiki/GLSL-Versions#version-330), at least on Mac OSX.

benfry: So we'll need to hold on this until https://github.com/processing/processing/issues/3505

benfry: But it sounds like that might take a little while, so let's add a platform check there and get the ES2 profile if it's running on ARM, and we can use that for the time being until #3505 is finished.

gohai: Great… will update and send a new PR.

gohai:
@benfry Send new PR #3583

PR #3571, opened 2015-08-10
Breakpoints don't 'jump' after hitting Enter on blank line

### LINKED ISSUE #3552: Breakpoint "jumps" after setting it (and subsequently hitting Enter) on a blank line
To reproduce this issue:

1. Hit Enter a few times
2. Move the cursor to somewhere in between, and create a new breakpoint, like so:

image

<ol start="3"><li>Hit Enter again:</li></ol>

image

NB: This only works when hitting Enter after setting a breakpoint on a blank line. Setting a breakpoint on a blank line, typing text below/above the line, and going back to the line and hitting Enter makes the breakpoint jump too.

joelmoniz:
This PR resolves #3552.

The issue was an off-by-one-error when a line consists purely of white-spaces.


benfry:
Thanks, Joel.

joelmoniz:
Sure :)

PR #3596, opened 2015-08-14
CM as in mockup

Akarshit:
I have tried to make the CM as given in the mockups.
Please tell if it breaks on any platform or the colors are off.


benfry:
Definitely a step in the right direction, thanks!

screen shot 2015-08-14 at 11 13 16 am

Akarshit:
@benfry this This is how it looks on linux. Can you please put up a screenshot of update tab and library tab with something selected.

benfry:
screen shot 2015-08-14 at 11 50 13 am

Since the window tabs at the top have a rigid size, making the window smaller than the minimum also cuts things off at the right-hand side (as seen in the previous screenshot). So either we need to allow it to adjust a little more, or use `frame.setMinimumSize()` if there's truly a minimum.

alignedleft:
Thank you, @Akarshit! This is great progress.

PR #3605, opened 2015-08-14
Get max core profile on OSX

JakubValtar:


core/src/processing/opengl/PJOGL.java
```diff
@@ -1240,7 +1240,7 @@ protected Object getDerivedFont(Object font, float size) {
 
   @Override
   protected String[] loadVertexShader(String filename, int version) {
-    if (PApplet.platform == PConstants.MACOSX) {
+    if (PApplet.platform == PConstants.WINDOWS) {
```
codeanticode: @JakubValtar shouldn't this be MACOSX, shader code is not pre-processed on mac, giving this https://github.com/processing/processing/issues/3617

Issue #158, opened 2013-02-10
Implement efficient version of copy() in PGraphicsOpenGL


processing-bugs:
_Original author: b...@processing.org (June 07, 2010 01:06:36)_

This bug automatically added from:
http://dev.processing.org/bugs/show_bug.cgi?id=941

Comment from fry, 2008-10-05 20:12

currently calling loadPixels(), the PImage.copy() method, then
updatePixels(), which is ridiculously slow. simple to fix with just a
little time spent hacking a few lines of OpenGL code together.

_Original issue: http://code.google.com/p/processing/issues/detail?id=119_

PR #3622, opened 2015-08-15
Third bunch of arm patches

gohai:
The remaining patches that I carried in my branch.

### DIFF
--- a/java/src/processing/mode/java/runner/Runner.java
+++ b/java/src/processing/mode/java/runner/Runner.java
@@ -59,6 +59,7 @@ public class Runner implements MessageConsumer {
 
   // Running remote VM
   protected VirtualMachine vm;
+  protected boolean vmReturnedError;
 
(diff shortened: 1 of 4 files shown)


java/src/processing/mode/java/runner/Runner.java
```diff
@@ -659,6 +670,9 @@ public static boolean handleCommonErrors(final String exceptionClass,
       }
     } else if (exceptionClass.equals("java.lang.UnsatisfiedLinkError")) {
       listener.statusError("A library used by this sketch is not installed properly.");
+      if (PApplet.platform == PConstants.LINUX) {
+        System.out.println(message);
```
aengelke: Related to #158 and #2827: It might be better to use `err.println` to write the output to the window of the sketch and not to the active window.

gohai: @aengelke Good point, fixing.

Issue #3609, opened 2015-08-15
CM column widths change with selection


alignedleft:
This is a weird one. Here is the CM panel showing the Android mode selected. Note the column widths.

![screen shot 2015-08-14 at 9 47 45 pm](https://cloud.githubusercontent.com/assets/1034002/9287532/3c61eaa2-42ce-11e5-85a2-ae117d38179f.png)

Now, when I select the Python mode in the list (which is disabled and incompatible), the Name column jumps to be much wider, pushing the Author column (and install/update/remove buttons) out to the right.

![screen shot 2015-08-14 at 9 47 55 pm](https://cloud.githubusercontent.com/assets/1034002/9287535/507e6588-42ce-11e5-8173-645ed592121f.png)

The column widths should not change when the selection changes.

Akarshit:
@benfry I am unable to track the exact cause of the issue, but i have found out that this is occurring in libraries whose pretty version contains some text.  
These libraries are - 
1. Python - prettyVersion=Build 0412, built 2015/07/08 14:03
2. Syphon - prettyVersion=1.0-RC3
3. HE_Mesh 2014 - prettyVersion=2.0.15 „Random”
I think this is occurring because of somewhere a exception getting thrown when these are tried to convert to a integer form or something like that.
Do you have any idea regarding this.

benfry:
I don't really understand--the point of prettyVersion is that it can contain text. So all of them should.

Akarshit:
I am sure about this because if i edit the contribs.txt and change Python's prettyVersion to something like 01 the column width doesn't change when selecting Python now.

Akarshit:
@benfry i figured out what was wrong. Actually there is a label above the update button which shows text like "Version `prettyVersion` available/installed". This was set invisible but coded to occupy space still. This was causing the statusPanel to resize.  
One of a solution would be to extract the version part from the `prettyVersion` but to do this all the libraries having different forms of prettyVersion would require a check. I was wondering if this was supposed to be done at server-side ?

benfry:
No, don't mess with the contents of `prettyVersion`. The Python Mode is an issue that should be fixed separately in contribs—that's a whole build string he's using, not a version. But even in fixing that, it sounds like you're looking for ways to paper over something bad in the code. There shouldn't be a reason that columns are resizing when a user clicks different rows in an interface.

Akarshit:
I think i was not able to express my self clearly.  
When i select a contibution there is text displayed below the install button like below
[...]

PR #3629, opened 2015-08-16
A temporary fix for column width

Akarshit:
I have changed the text from "Version 1.0.2 available" to "v1.0.2 available"and also tried to shorten the version.
This is a temporary fix for #3609

### DIFF
--- a/app/src/processing/app/contrib/ContributionListPanel.java
+++ b/app/src/processing/app/contrib/ContributionListPanel.java
@@ -32,8 +32,6 @@ import javax.swing.border.Border;
 import javax.swing.event.*;
 import javax.swing.table.*;
 
-import com.sun.xml.internal.bind.v2.runtime.reflect.Lister.Pack;
-
 import processing.app.Base;
 import processing.app.Platform;
 import processing.app.ui.Toolkit;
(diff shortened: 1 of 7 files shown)


app/src/processing/app/contrib/StatusPanel.java
```diff
@@ -235,9 +236,9 @@ public void update(ContributionPanel panel) {
       latestVersion = "Update";
     }
 
-    if(currentVersion != null){
-      currentVersion = "Version " + currentVersion;
-    }else{
+    if (currentVersion != null) {
+      currentVersion = "v" + currentVersion;
```
benfry: This makes an assumption about the format of the version field that may not always be correct. Putting "v" in front won't make sense in some cases. Just leave it off.

app/src/processing/app/contrib/ContributionListPanel.java
```diff
@@ -32,6 +32,8 @@
 import javax.swing.event.*;
 import javax.swing.table.*;
 
+import com.sun.xml.internal.bind.v2.runtime.reflect.Lister.Pack;
```
benfry: Why is this here?

benfry:
I'll wait until you have the final fix.

app/src/processing/app/contrib/StatusPanel.java
Akarshit: @benfry Do you want the label to read "1.0.1 available" only ?

app/src/processing/app/contrib/ContributionListPanel.java
Akarshit: My bad.

app/src/processing/app/contrib/StatusPanel.java
benfry: Yes, but via `Language.interpolate()`, and make the string "%s available" so that it can be localized to languages that might need the equivalent of "available" in front.

PR #3635, opened 2015-08-16
Allow Movie Maker to cope with bad files.

GKFX:


benfry:
Thanks for the patch!

Issue #3599, opened 2015-08-14
CM blue bar missing


alignedleft:
In the reference mockup, there is a 1px tall horizontal light blue bar that separates the tabs from the content below.  Look closely here:

![screen shot 2015-08-14 at 3 01 06 pm](https://cloud.githubusercontent.com/assets/1034002/9284922/511541b2-4295-11e5-824e-a7e65b303231.png)

Currently, that bar doesn't appear:

![screen shot 2015-08-14 at 2 59 08 pm](https://cloud.githubusercontent.com/assets/1034002/9284929/67a2280a-4295-11e5-8c83-0172117e919b.png)

benfry:
Fixed in #3636 for beta 5.

PR #3636, opened 2015-08-16
Bar below the tab is painted

Akarshit:
Fix for #3599 
Please check for compatibility with mac.

### DIFF
--- a/app/src/processing/app/contrib/ContributionManagerDialog.java
+++ b/app/src/processing/app/contrib/ContributionManagerDialog.java
@@ -261,7 +261,7 @@ public class ContributionManagerDialog {
         @Override
         protected void paintComponent(Graphics g) {
           g.setClip(Toolkit.createRoundRect(0, 0,
-                                          getWidth(), getHeight()- 2,
+                                          getWidth(), getHeight(),
                                           temp == 0 ? 6 : 0,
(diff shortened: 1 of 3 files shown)


benfry:
I'll wait for confirmation from @alignedleft

alignedleft:
Blue bar looks good, but why is there a gap between it and the tabs above? They should be touching; there should be no vertical space. See:

screen shot 2015-08-16 at 2 45 24 pm

Also, when you switch to the Update tab, the blue bar disappears and I see a dark gray bar. The "Updates" tab view should be consistent with the other views in this respect.

screen shot 2015-08-16 at 2 45 30 pm

Akarshit:
@alingedleft Please check if this resolves the issue now.

alignedleft:
No, there is still a gap.  See:

screen shot 2015-08-19 at 12 36 52 pm

Also, the dark gray 1px horizontal line just above the column names should not appear. Note it appears above, and is even more noticeable in the Updates view:

screen shot 2015-08-19 at 12 36 59 pm

Akarshit:
@alignedleft There is a separate issue for the gap. This PR was only for that bar getting painted in all the tabs. I hope that is resolved

alignedleft:
@Akarshit sorry for that. Yes, the blue bar is now present and working, thanks!

Issue #707, opened 2013-02-10
Undo does not move to the correct location in the editor window


processing-bugs:
_Original author: mw.mzp...@gmail.com (May 08, 2011 15:07:01)_

Jump to the line and column where the done-and-undone modification took place.

_Original issue: http://code.google.com/p/processing/issues/detail?id=668_

JakubValtar:
Looks like it is fixed, needs more testing.

benfry:
I think this one isn't, it's been fixed many times but I feel like we're still getting reports about it.

benfry:
@REAS @shiffman don't you guys have trouble with this one and students?

benfry:
Or better yet, can someone help us with testing this one? @JakubValtar couldn't get it to reproduce, but we'd like some additional testing before we close.

joshgiesbrecht:
Just tested it and it's behaving about the same way as I described above.  It more-or-less works except sometimes it's leaving the caret a few lines below the edit instead of directly below.  It seems to be just the caret position that's off, it still scrolls up enough to see the edited line.
(Win 7 in case that matters)

shiffman:
The problem I experienced (in the past) was that code would disappear after several undos.  I have not noticed this in the recent 3.0 alpha and beta releases and undo seems to be working better. I'll keep a closer eye out.

joshgiesbrecht:
I hadn't noticed when #3003 came up.  I'll try to look into it soonish, and keep an eye out for the misplaced-caret problem along the way.  (Finally got my local build working with 3.0 just recently)

Unless the misplaced caret problem is significant enough to still bother anyone, maybe this one could be closed and the focus can be on #3003?

benfry:
There's no reason we shouldn't be able to fix the misplaced caret problem—sounds like just a random quirk? I seem to recall that the Undo state information records caret placement, so the code should be in place for handling it. 

@joshgiesbrecht Was there a problem with the default build or just something on your machine?

joshgiesbrecht:
Nah just my local setup. I hadn't cleaned things out and rebuilt since pre-3.0.

JukePlz:
I get this and the mentioned dissapearing text a lot. Im pretty sure it's some race condition with automatic markup (ctrl+t) previously to an undo action.
Will try to find better repro steps later.

joshgiesbrecht:
Ok, actually looking at this today. I have a good idea why the caret is off by a line or two.  Could definitely use repro steps for anything else.

[...]

PR #3660, opened 2015-08-19
storing caret for undo at start of edit

joshgiesbrecht:
...so that when we undo we're moving to where we began,
not where things would have been in the edited text after.

Fix for #707.


benfry:
Incorporated for 3.0 beta 5.

PR #3678, opened 2015-08-20
Fix ARM JDK download for 8u60

gohai:
ARM started offering builds for 64 bit ARM (ARMv8), which made them rename the download file.


benfry:
Yeah, was also excited to see that they're releasing the ARM builds alongside the others—that's a good sign.

PR #3681, opened 2015-08-21
Icons are implemented

Akarshit:
I hope that this is the correct implementation.  
Shouldn't there be a field in Toolkit which is simply set to 1 or 2 depending upon the resolution.
screenshot from 2015-08-21 08 51 57
But the notification circle is a little small as can be seen above.


alignedleft:
@Akarshit I think the small Updates circle is OK. It will be very rare for people to have so many items installed and need of updating.

PR #3686, opened 2015-08-21
FX - Make key events little bit more sane

JakubValtar:
Make key events behave similar to Java2D.

Returned key chars in pressed/released events are always uppercase, but keyCodes are correct.
Key typed events give correct key char, but zero keyCode.

All non-alphanumeric keys I could find on my keyboard now return key char PConstants.CODED.

### DIFF
--- a/core/src/processing/javafx/PSurfaceFX.java
+++ b/core/src/processing/javafx/PSurfaceFX.java
@@ -706,17 +706,82 @@ public class PSurfaceFX implements PSurface {
     }
 
     long when = System.currentTimeMillis();
-    KeyCode kc = fxEvent.getCode();
-    // Are they f*ing serious?
-    char key;
-    if (et == KeyEvent.KEY_TYPED) {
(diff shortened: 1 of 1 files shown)


core/src/processing/javafx/PSurfaceFX.java
```diff
@@ -706,17 +706,82 @@ protected void fxKeyEvent(javafx.scene.input.KeyEvent fxEvent) {
     }
 
     long when = System.currentTimeMillis();
-    KeyCode kc = fxEvent.getCode();
-    // Are they f*ing serious?
-    char key;
-    if (et == KeyEvent.KEY_TYPED) {
-      key = fxEvent.getCharacter().charAt(0);
```
benfry: I did get an `ArrayIndexOutOfBoundsException` on this line earlier today (I think I hit ESC to close a sketch), in case that's not fixed in this commit.

PR #3699, opened 2015-08-22
Rendering blank components for #3667

Akarshit:
Possibly would render blank components for #3667 and then inform the listeners that the value at that cell will be changed.


alignedleft:
@Akarshit Sorry, I'm not sure what you mean by "blank components," but I just tested this branch and, while the errors are no longer thrown, now most of the time elements cannot be selected at all.

I made a video which shows clicks so you can see how I'm (usually) unable to select items when clicking on them:

https://www.dropbox.com/s/4by5nfl5ny4tv1k/CM_click_tests.mov?dl=0

benfry:
Closing this PR based on @alignedleft's feedback.

Issue #3003, opened 2014-12-09
Undo sometimes causes the editor to go blank


REAS:
I have heard reports from a number of my students that Undo is eating code. This is a note to record the potential issue. I need to research further.

REAS:
OK, I have finally seen this after continued reports from students. The situation here is that with multiple undos (more than a few) all of the code can disappear from the editor window. Then on the next undo, the code reappeared in its different "undone" state. My guess is that students are panicking when they see the code completely gone and they may not continue to undo to bring it back. So, what might be causing this?

benfry:
The issue is probably tied to the idea of “compound edits”... The editor has to be blanked before new code is added, but those two actions have to be bundled together as a compound edit to avoid this situation. 

Where possible, the code should already be doing this, but perhaps we're missing a case.

joelmoniz:
I was testing out the undo functionality during the process of easing in the RSTA, and noticed that undo clears everything in the current tab when one does a format. Undoing again adds back the text in its original form. Is it possible that this is one of the cases not bundled as a compound edit, and causing this issue to arise?

benfry:
@joelmoniz's patch for the auto-format case is now merged in for 3.0 beta 5. For people who have seen this, do you think this is the likely cause of the issue?

shiffman:
Well, the good news is that this has helped my identify the error and reproduce it consistently.  The bad news is that I am still getting it.   Here is a screen capture of "auto-format-undo-code-disappearing" sequence.

https://www.dropbox.com/s/qizua1oyv18cpch/auto_format_error.mov?dl=0

joshgiesbrecht:
Were you doing something specific before Auto Format in that video?  I'm trying the same thing here and unable to reproduce the bug.

Steps I've done:
1) Type some code.
2) Select AutoFormat from the Edit menu.
3) Undo

In fact ... in a case where Auto-Format responds that no changes are necessary, the Undo actually undoes the last thing I typed.  ie. the Auto-Format isn't even put on the Undo stack.

shiffman:
@joshgiesbrecht My steps:

1. Open new sketch.
2. Type code
3. Auto-format.
4. Undo

I'm running 0243 on Mac OS X 10.10.4.

benfry:
And you're definitely running the latest source? I can't reproduce it.

shiffman:
[...]

PR #3702, opened 2015-08-23
Fix (another) issue with undoing auto-format

joelmoniz:
This PR fixes the (second) "undo-autoformat-clears-text" issue triggered as described [here](https://github.com/processing/processing/issues/3003#issuecomment-133811839).

### DIFF
--- a/app/src/processing/app/ui/Editor.java
+++ b/app/src/processing/app/ui/Editor.java
@@ -2216,11 +2216,23 @@ public abstract class Editor extends JFrame implements RunnerListener {
       if (formattedText.equals(source)) {
         statusNotice(Language.text("editor.status.autoformat.no_changes"));
       } else {
+        /*
+         * Call endTextEditHistory() before starting a new CompoundEdit,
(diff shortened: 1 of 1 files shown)


benfry:
Thanks, I've moved the `endTextEditHistory()` to `startCompoundEdit()` so it'll always work this way (unless you can see any reason why we shouldn't?)

joelmoniz:
Oh yes, that's good, I can't think of any reason not to do this. 

NB: Perhaps the [`stopCompundEdit()`](https://github.com/processing/processing/commit/7628427d01f25ffdfa0197c389c91b60c3b2179a#diff-b20c022062e9f7e4fcd445115c49024dR1928) call can be removed, since `endTextEditHistory()` calls `stopCompundEdit()` from within it (although it doesn't make much difference, since `stopCompundEdit()` accounts for the possibility of it being called multiple times). Shall I submit a quick PR for this?

benfry:
Good point; I've just made the change and we should be all set.

PR #3709, opened 2015-08-24
FX - basic pixel operations (get, set, load, update)

JakubValtar:
Will be refactored later when I will have more complete picture of what should be happening inside the FX renderer and how to make it efficient. Density might not work correctly, that's also something I will be dealing with later.


benfry:
Yeah, density will be additional work, but more just being mindful of the coordinates changing.

PR #3710, opened 2015-08-24
FX - fix transformation stack NPE

JakubValtar:


benfry:
Boy, from the nature of some of these fixes, it's as if the original author were writing the code after a long day and while having a drink...

JakubValtar:
I was staring at this line and it took me a while to figure it out, and I'm yet to have a drink today (I need one).

PR #3712, opened 2015-08-24
FX - align to pixel grid when drawing 1 px strokes

JakubValtar:


benfry:
Would we be better off moving everything by 0.5? (We sometimes do a `translate(0.5, 0.5)` at the top of canvas-based projects...)

JakubValtar:
Then you would have blurred edges of shapes. It's not perfect, but there's no clear winner among the solutions I'm aware of.

benfry:
"No clear winner" wins again.

JakubValtar:
It's not event a "win" when there is no competition.

PR #3716, opened 2015-08-25
FX - keep track of whether pixels are up to date

JakubValtar:
This prevents taking snapshot of the whole stage in loadPixels() when it
is not needed. Otherwise setting pixels one by one loads the pixels for
each one of them and freezes the sketch.


JakubValtar:
Implemented in #3725

PR #3730, opened 2015-08-27
FX - normalize enter key

JakubValtar:


benfry:
Mmm.. I don't think we want to do this. RETURN and ENTER mean different things. Or are we doing that elsewhere?

JakubValtar:
- Java2D is returning only \n
- FX was undefined for press/release, \r for typed
- JOGL was \r for press/release, no typed events

We can only return one char, so no \r\n. Why don't we just make it \n
everywhere and users can replace it with System.lineSeparator() if they
want to use it as text input?

Is there any practical benefit of returning \r on some systems and \n on
other systems?

benfry:
Awesome. Well, \n it is. Is that what this patch will give us?

JakubValtar:
Just for FX. Will send a new one for JOGL tomorrow.

PR #3731, opened 2015-08-27
FX - add FX2D keyword, remove JFX keyword

JakubValtar:


benfry:
Whups, I just merged this, but I think that file is auto-generated. @REAS?

JakubValtar:
Auto-generated from docs? Because I don't see JFX in PConstants and I'm
fairly certain I've ant cleaned recently.

JakubValtar:
Ah sorry, I've just read the header. I did global search and jumped right in the middle of the file.

REAS:
Should I re-generate the file and commit it? 

@JakubValtar, you can make those modifications here: https://github.com/processing/processing-docs/blob/master/generate/keywords_base.txt

You can generate the keywords file by running this Perl code:
https://github.com/processing/processing-docs/blob/master/generate/keywords_create.cgi

JakubValtar:
Thanks @REAS. We want to keep these changes. Is it enough to just edit the `keywords_base.txt` file you linked then? I'm not sure what the Perl script actually does and if I should run it at all.

REAS:
Yes, you can update `keywords_base.txt`. I'll do the rest.

JakubValtar:
Thanks Casey, I sent a PR to processing-docs. Sorry for the mess!

PR #3737, opened 2015-08-28
Draw text with empty lines correctly

### LINKED ISSUE #3736: Incomplete text rendering of strings with consecutive line breaks
When using the 5-argument version of text(), if there are multiple consecutive line breaks, the text after the line break is not rendered. Minimal example:

```java
void draw() {
  background(0);
  String s = "Hello\n\nWorld.";
  fill(255);
  text(s,10,10,100,100);
}
```

Output only shows "Hello"

This bug appears to have been introduced after 3.0b3, where the text is rendered correctly.

Tested with Processing 3.0b4 on OSX, Linux, and Windows,, and 3.0b5 on OSX.

JakubValtar:
Fixes #3736


benfry:
Awesome, thank you.

PR #3745, opened 2015-08-30
Update Base.java

atduskgreg:
Change Processing from referring to itself as "himself" to "itself" in an error message. No reason to gender it.


benfry:
Thanks, I saw that the other day and thought it was weird. I don't recall writing it to say "himself," but who knows.

PR #3748, opened 2015-08-31
Improve sketchPath calc on Win & Linux

JakubValtar:
Some people may put core.jar in a subdirectory of /lib/ in their project. Adjusted the code to deal with it.


benfry:
Definitely not something we want to support. They can set the sketch path on the command line if they want to f*k around with the location of basic files like this.

Issue #3304, opened 2015-05-20
Examples window flickering when open during library install / remove


shiffman:
3.0a9, Mac OS X 10.10

To reproduce:

* Open examples window
* Open library manager
* Install or remove a library

This is a good one for GSoC, cc @alignedleft @Akarshit

Akarshit:
@alignedleft  I am not able to reproduce this on Linux. Can you please check if this has been fixed.  
Though after a contribution gets installed or removed the Examples window is made again so it closes and than opens again.

alignedleft:
@Akarshit Yes, this is still an issue. 

I followed @shiffman’s steps above, and installed the Arduino library. The window flashed closed, then reopened, and I saw the Arduino examples now appeared in the Examples window.

But it threw an error (pasted below).

Also, after this process, all libraries listed under "Arduino" had disappeared from the list (see screenshot).

![screen shot 2015-08-24 at 1 53 03 pm](https://cloud.githubusercontent.com/assets/1034002/9452172/aa8c9f22-4a67-11e5-8d8e-9f8fafce3c85.png)


```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at processing.app.contrib.ContributionListPanel$StatusRendere.getTableCellRendererComponent(ContributionListPanel.java:354)
	at javax.swing.JTable$AccessibleJTable.getAccessibleChild(JTable.java:7031)
	at javax.swing.JTable$AccessibleJTable.getAccessibleAt(JTable.java:7418)
	at javax.swing.JTable$AccessibleJTable.valueChanged(JTable.java:6933)
	at javax.swing.DefaultListSelectionModel.fireValueChanged(DefaultListSelectionModel.java:184)
	at javax.swing.DefaultListSelectionModel.fireValueChanged(DefaultListSelectionModel.java:164)
	at javax.swing.DefaultListSelectionModel.fireValueChanged(DefaultListSelectionModel.java:211)
	at javax.swing.DefaultListSelectionModel.changeSelection(DefaultListSelectionModel.java:405)
	at javax.swing.DefaultListSelectionModel.changeSelection(DefaultListSelectionModel.java:415)
	at javax.swing.DefaultListSelectionModel.setSelectionInterval(DefaultListSelectionModel.java:459)
	at javax.swing.JTable.setRowSelectionInterval(JTable.java:2166)
	at processing.app.contrib.ContributionListPanel.updatePanelOrdering(ContributionListPanel.java:442)
[...]

PR #3804, opened 2015-09-16
Examples frame does not flikers when contribution is changed

Akarshit:
Fix for #3304


benfry:
Reverted this change because of #3813.

Issue #3806, opened 2015-09-16
CM Buttons enabled even after clicking them


Akarshit:
When the install button is pressed the CM start s downloading the contribution, during that time the button is still enabled and the contribution can be downloaded again by pressing the install button

benfry:
Fixed in #3817.

PR #3817, opened 2015-09-18
Buttons disabled when clicked

Akarshit:
Fix for #3806 
I don't know why git isn't rebasing this.


benfry:
From here it looks like it's rebasing correctly?

Akarshit:
Aren't - Revert "Update to a incompatible version is not possible" and Revert "Revert "Update to a incompatible version is not possible"" supposed to be removed while rebasing ?

PR #3833, opened 2015-09-20
Make ARM JDK download work again with 8u51

gohai:


benfry:
Thanks for staying on top of this...

PR #3875, opened 2015-09-26
Low-WeakReference diet

JakubValtar:
- keeps the good stuff in
- no side effects
- Dr. C. stamp of approval

### DIFF
--- a/core/src/processing/opengl/FontTexture.java
+++ b/core/src/processing/opengl/FontTexture.java
@@ -155,13 +155,10 @@ class FontTexture implements PConstants {
       // REPLACE to preserve color of transparent pixels.
       Texture tex0 = textures[currentTex];
 
-      PGraphicsOpenGL g = tex.pg.get();
-      if (g != null) {
-        g.pushStyle();
-        g.blendMode(REPLACE);
-        tex.put(tex0);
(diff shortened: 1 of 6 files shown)


codeanticode:
@JakubValtar I merged this PR, but then the framebuffer and texture resources in the test case I have been using:

```java
  public void settings() {
    size(640, 360, P2D);
     smooth(2);
  }

  public void setup() {
  }

  public void draw() {
      PGraphicsOpenGL pg = (PGraphicsOpenGL)createGraphics(width, height, P2D);
      pg.beginDraw();
      pg.background(random(255), random(255), random(255));
      pg.endDraw();
      image(pg, 0, 0);
  }  
```

don't get disposed. Could you double check?

JakubValtar:
Ok I'm checking it now

JakubValtar:
Why do you think so? Heap dump shows no letovers and when I put println() into disposeNative() I see that resources are deleted every GC cycle (which runs once every few seconds).

codeanticode:
my mistake, sorry!

Yes, resource disposal is working great!

JakubValtar:
Ah great, I was worried that there is something platform specific :)

PR #3880, opened 2015-09-26
Manage changing inset size when calling setResizable()

### LINKED ISSUE #3868: Frame size displays incorrectly if surface.setResizable(true) 3.0b7
If surface.setResizable(true) for the default renderer, the frame displays with the circle out of bounds on the right and bottom of frame.  (Processing 3.0b7)

CODE:
```
public void settings() {
    size(400, 400);
}

public void setup() {
    surface.setResizable(true);
}

public void draw() {
    fill(0,255,255);
    stroke(0,0,255);
    strokeWeight(10);
    ellipse(200, 200, 390, 390);
}
```

JakubValtar:
This is ugly, but I didn't find any other way.

Please test on Mac and Linux before merging.

Fixes #3868

### DIFF
--- a/core/src/processing/awt/PSurfaceAWT.java
+++ b/core/src/processing/awt/PSurfaceAWT.java
@@ -74,6 +74,11 @@ public class PSurfaceAWT extends PSurfaceNone {
   // Note that x and y may not be zero, depending on the display configuration
   Rectangle screenRect;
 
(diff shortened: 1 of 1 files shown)


benfry:
The test in the report still performs w/o trouble on OS X, so hopefully no regression there.

Issue #3827, opened 2015-09-19
Some keys is not recognized properly in P2D / P3D rendering modes


StanislavGrinkov:
Issue: Some keys is not recognized in P2D / P3D mode on ubuntu trusty (14.04).

Environment: Ubuntu Trusty, Processing 3.0b6

How to replicate: run sketch in different render modes and press keys

Expected: Keys is recognized properly in each render mode
![render-keys-default](https://cloud.githubusercontent.com/assets/7836782/9974829/80b05c76-5ec5-11e5-8a03-2bccb8b0cd8f.png)

Actual: see screenshots
P2D
![render-keys-p2d](https://cloud.githubusercontent.com/assets/7836782/9974833/99e192d2-5ec5-11e5-871b-14bb15cd32e7.png)

P3D
![render-keys-p3d](https://cloud.githubusercontent.com/assets/7836782/9974834/9b8380be-5ec5-11e5-9a77-8329141b05de.png)

Possibly linked to #3352

// ~~~~~~  Sketch  ~~~~~~~
import java.awt.event.KeyEvent;

void setup()
{
  size(100,100);
  // size(100,100, P2D);
  // size(100,100, P3D);
}

void draw()
{
  background(123);
}

void keyPressed() 
{
  println("Key is ==>"+KeyEvent.getKeyText(keyCode)+"<==");
}

nakednous:
Same processing version, but in my archlinux-64 bit all the keys are recognized. Perhaps it has to do with your ubuntu version which seems a bit dated. Can you try to test it under a different (perhaps newer) environment?

StanislavGrinkov:
I will try on another distro version when possible.
p.s. 14.04 is a LTS version of ubuntu and I have all available updates installed. (I feel that it could be issue with my system, as usual when I report issues :) )

StanislavGrinkov:
Tried on live ubuntu 15.04. No issues. It could be something with my java version /keybd daemon or something else.
Sorry for taking your time.
Please close this issue.

JakubValtar:
Hmm, I don't think AWT will give you correct results for NEWT events which are used in OpenGL. Try to use something like this:
```java
import com.jogamp.newt.event.KeyEvent;

void keyPressed()
{
  switch (keyCode) {
    case KeyEvent.VK_DELETE:
      println("DELETE");
      break;
    case KeyEvent.VK_INSERT:
      println("INSERT");
      break;
  }
}
```

benfry:
[...]

PR #3888, opened 2015-09-28
Convert all documented hacky keys in OpenGL

JakubValtar:
These are BACKSPACE, TAB, ENTER, ESC, and DELETE. Enter and Return can't be distinguished, both are reported as keyCode 10 (keyChar \n).

### DIFF
--- a/core/src/processing/opengl/PSurfaceJOGL.java
+++ b/core/src/processing/opengl/PSurfaceJOGL.java
@@ -995,12 +995,10 @@ public class PSurfaceJOGL implements PSurface {
       keyCode = mapToPConst(code);
       keyChar = PConstants.CODED;
     } else if (isHackyKey(code)) {
-      keyCode = code;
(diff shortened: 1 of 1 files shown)


benfry:
Does this take care of #3827 or are there remaining issues?

JakubValtar:
That issue is that they are using `java.awt.event.KeyEvent.getKeyText(keyCode)` to get a key label from the keyCode, but JOGL uses different keyCodes than AWT for some keys. We could have KeyCode constants for all the non-ascii keys and convert whatever AWT or JOGL gives us to those so users don't have to mess with the native events, but I'm not sure if we want to do all of this and support it.

benfry:
Closing out the remaining problems as something we have to handle in docs since it's essentially unsolvable: https://github.com/processing/processing-docs/issues/318 (Or at least can't be resolved without an amount of code and testing that's way out of proportion with the impact of the issue at hand.)

PR #3890, opened 2015-09-28
Added the remove filter feature

Akarshit:
Adds a cross button on the right side of the search field in CM. Currently a "X" is being displayed but can be replaced by a icon.
Might be a cool feature. Looks like below on linux - 
screenshot from 2015-09-28 19 35 58

### DIFF
--- a/app/src/processing/app/contrib/ContributionTab.java
+++ b/app/src/processing/app/contrib/ContributionTab.java
@@ -23,6 +23,7 @@
 package processing.app.contrib;
 
 import java.awt.Color;
+import java.awt.Cursor;
 import java.awt.Dimension;
 import java.awt.Font;
 import java.awt.event.*;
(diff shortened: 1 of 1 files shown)


Akarshit:
@jamesjgrady @alignedleft What are your thoughts on this one ?

alignedleft:
@Akarshit Yes, I think this is a nice idea, as long as an icon can be used that is consistent with the rest of the UI.

Akarshit:
@alignedleft I have added the icon. It looks like this now   
screenshot from 2015-12-11 11 25 45

alignedleft:
This looks very nice to me. @jamesjgrady would you mind just looking and letting us know if you approve?

jamesjgrady:
I think the x looks fine. I don't love the dark orange highlight box. Is there anything that can be done about that, or is it OS dependent?

benfry:
Implemented for 3.2.4. 

@jamesjgrady the orange is from Linux, on the Mac we get its selection highlight color, so yeah, it's OS dependent.

Issue #3518, opened 2015-07-30
Implement UI for number of updates available in the footer


shiffman:
I believe we've had this discussion in voice or e-mail so I thought I would open a github issue to track.  In light of the improvements to the CM that @Akarshit and @alignedleft are working on, I am wondering if we need the "there are updates available" pop-up.   For me, the pop-up appears a good time after Processing first launches and always seems to be in my way.  I believe I can turn it off in the preferences under "Allow update checking" but I'm unclear on this preference -- I would like to allow update checking, but I just don't want a pop-up.   Is there a better way we could indicate there are updates available perhaps in a menu or with a small icon in the PDE itself?  Or should the pop-up have a "don't show again" checkbox?

alignedleft:
Great idea. I would love to see a smoother UI for updates, but my sense was it perhaps was too much to tackle as part of 3.0.

I believe the way we last left this, we said we'd keep the existing "PDE update available" popup dialog, but there would be no such update notifications for contributions — those would all appear in the CM.

Generally, I support the idea of communicating "updates available" to the user, especially in a form less intrusive than a popup, yet more actively than hidden in the CM window.  A small badge/icon in the PDE window or even a menu item (e.g., "Updates (3 available)…") make sense to me.  (I can't speak to whether or not now is the time to implement that.)

Akarshit:
I am sot sure about this but it might be best to do this in form of something equivalent to a Toast in android.

benfry:
Let's figure out something less intrusive than a popup, like a red dot in the interface somewhere appropriate. We can have the popup show instantly (the delay is actually in the code was inserted by someone to not further annoy users by getting in the way before they can even do anything), but I think it's annoying no matter what. 

Thinking out loud, we could potentially even have a new icon show up on the right-hand side of the toolbar... But adding a preference to paper over bad design choices is never a good route.

alignedleft:
What about something like the following options?

![artboard 1](https://cloud.githubusercontent.com/assets/1034002/9043868/90bdec7c-39cd-11e5-9529-60248b382de8.png)

![artboard 2](https://cloud.githubusercontent.com/assets/1034002/9043869/93a1eb32-39cd-11e5-9587-3a5f30c8220d.png)

benfry:
James is working on this while we sort out the last round of changes to the editor UI.

benfry:
I've removed the dialog box popup from the next release, but the new UI that shows the indicator probably won't make it in time for that. Stay tuned...

[...]

PR #3896, opened 2015-09-28
CM: Send update count to Editors

JakubValtar:
Requested for #3518

See https://github.com/processing/processing/issues/3518#issuecomment-143768021

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -342,6 +342,9 @@ public class Base {
 
     // check for updates
     new UpdateCheck(this);
+
+    ContributionListing cl = ContributionListing.getInstance();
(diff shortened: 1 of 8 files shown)


benfry:
Much thank you.

JakubValtar:
I just figured out: if you open new Editor window, it should be able to get the update count and display it, so we might want to store the update count in Base and let the Editor query it on stratup.

benfry:
The issue is that it (also) needs to be event driven, because the number will change as updates happen.

JakubValtar:
Yes, event driven is already in place an there to stay. I was mentioning a situation when you have the PDE running and you open a new Editor window, the Editor has to get the update count somewhere because there will be no event from the Contribution Manager.

PR #3942, opened 2015-10-01
Count only compatible updates

### LINKED ISSUE #3941: Wrong information about update
Processing 3.0 indicate me, I've "updates" available, but all my libraries, example are ok.
May be because I've a lot of libraries, and few of them are very old ? I add a little screen shot of a part of my library manager may be can help.
capture d ecran 2015-10-01 a 09 28 33

capture d ecran 2015-10-01 a 09 24 51

JakubValtar:
Fixes #3941


benfry:
Did you mean to close this?

JakubValtar:
Yes, it already checks for this in hasUpdates() so the problem is somewhere else.

Issue #3945, opened 2015-10-01
Core libraries not included in update count displayed on the editor


Akarshit:
The update notification on the editor shows 4 updates - 
![screenshot from 2015-10-01 22 23 53](https://cloud.githubusercontent.com/assets/7762605/10227831/3ed24734-688d-11e5-847a-9e317ba49b14.png)
But the CM shows 5 updates
![screenshot from 2015-10-01 22 24 48](https://cloud.githubusercontent.com/assets/7762605/10227833/425c5052-688d-11e5-9e73-6c0a6842de88.png)

PR #3946, opened 2015-10-01
Core libraries added to the count of updates

Akarshit:
Fix for #3945


benfry:
Shouldn't this walk the list of `Mode` objects and get `coreLibraries` from each?

Akarshit:
Will confirm and modify this.

Akarshit:
Wow! After 3 years. :)

PR #3985, opened 2015-10-11
I/O library

gohai:
Hi @benfry @REAS!

This is the first chunk of the I/O library. I am happy with the interfaces and the source-file documentation. Will do additional testing in the afternoon and rework the example sketches. Those will come as a separate chunk later today.

Best
G


gohai:
Added (revamped) the following examples to this tree as well:

* make a LED blink (classic..)
* color an ellipse according to the state of an input pin
* binary counting using the PI's on-board LEDs
* jump into a custom interrupt handler when the state of an input pin changes
* control an analog-to-digital converter via I2C
* the same, but done in an fun OOP way
* get a compass heading via I2C and draw it on the screen
* color the screen background with values from a very fast SPI digital-to-analog-converter
* and again, also done in an OOP way

REAS:
Hey gents, I'm writing an email to you both about some details...

gohai:
Awesome, thanks Casey! I'll have some time over next couple of days to work on documentation or whatever else you find is missing..

shiffman:
Would love to be in the loop, am planning a Pi workshop using this at ITP hopefully in November (or whenever appropriate!)

Issue #3913, opened 2015-09-30
Editor menu is outside the visible screen when using the Raspberry Pi 800×480 7" TFT 


xranby:
On 8 sept 2015 Raspberry Pi foundation announced an 7" TFT add-on display with 10 point multi-touch for use with the Raspberry Pi
https://www.raspberrypi.org/blog/the-eagerly-awaited-raspberry-pi-display/

This screen has a quite small screen size 800×480 causing issues displaying the processing editor, the main menu gets hidden outside the screen as seen on the screenshot below:
![3b7 on raspbery pi with 7" touchscreen](http://labb.zafena.se/processing/3.0b7+rpi/image20150930_133211201.jpg)

workarounds:
The user has to know arcane workaround such as holding down ALT and moving the window or right clicking on the menu bar and forcing the editor window to maximize.

Tested using Processing 3.0 b7

gohai:
Will look into shrinking the Text Editor on small resolutions. Post 3.0 material in any way.

benfry:
To fix, set `editor.window.height.min` inside ~/.processing/preferences.txt to a value that works. Then we can add that to the `defaults.txt` file.

benfry:
Fixed for the next release.

PR #3992, opened 2015-10-13
Small usability fixes for tiny screens

gohai:
We could also resize the editor proportionally to fit inside the screen. Not sure if it's worth it though, since that means even less real estate..

Attaching before and after screenshot of the change regarding the display window position - please let me know if you have any other ideas for the exact coordinates.

2015-10-13-120841_800x480_scrot
2015-10-13-121248_800x480_scrot

### DIFF
--- a/app/src/processing/app/ui/EditorState.java
+++ b/app/src/processing/app/ui/EditorState.java
@@ -161,6 +161,10 @@ public class EditorState {
     int defaultWidth = Preferences.getInteger("editor.window.width.default");
     int defaultHeight = Preferences.getInteger("editor.window.height.default");
 
+    // go even smaller for tiny screens
+    defaultWidth = Math.min(defaultWidth, deviceBounds.width);
+    defaultHeight = Math.min(defaultHeight, deviceBounds.height);
+
     if (editors.size() == 0) {
       // If no current active editor, use default placement.
       // Center the window on ths screen, taking into account that the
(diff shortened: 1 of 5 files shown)


gohai:
Fixes #3913.

benfry:
@jamesjgrady do you have a preference for where the sketch window pops up relative to the editor window? I agree with @gohai that we should make sure it's not covering the play/stop buttons, but a more considered placement would be best. (This is only for cases where the sketch window doesn't fit to the left of the editor window.)

jamesjgrady:
@benfry I'd recommend the editor takes up the left half of the small screen and the canvas pops up on the right hand so not to cover the editor at all. I'm not sure if you have that kind of control?

benfry:
Nope. The question is more like “when the editor covers the whole screen, where should the sketch show up, relative to the editor?”

gohai:
(We *do* have the display window's dimensions, so centering, centering-unless-window-bar-is-offscreen, etc. would be very feasible.)

jamesjgrady:
@benfry ok, then I'd say place it in the center. It's going to depend on the size you make your sketch so that's why I place it close to the top but not covering the other title bar. 
screen shot 2015-10-13 at 2 17 12 pm copy

benfry:
Going with a hybrid of the two PRs for 3.0.1.

PR #3997, opened 2015-10-14
Medium-sized I/O updates

gohai:


java/libraries/io/src/processing/io/GPIO.java
```diff
@@ -165,7 +165,7 @@ protected static void checkValidPin(int pin) {
    *  the sketch is going to achieve.
    *  @param ms milliseconds to pause
    */ 
-  protected static void delay(int ms) {
+  public static void delay(int ms) {
```
benfry: Why is there another `delay()` method that duplicates the one in `PApplet`?

java/libraries/io/src/processing/io/LED.java
```diff
@@ -141,7 +141,7 @@ public void close() {
    *  Sets the brightness
    *  @param bright 0.0 (off) to 1.0 (maximum)
    */
-  public void set(float bright) {
+  public void setBrightness(float bright) {
```
benfry: `brightness()` would be the correct name here (assuming there aren't other brightness-related methods planned).

java/libraries/io/src/processing/io/GPIO.java
gohai: Thanks for pointing out - must have totally forgotten about the other one. Will remove. And since microseconds-type granularity is a little bit far fetched anyway, I guess the other one can go as well. Or perhaps a variant of delay() in PApplet that takes a float?

java/libraries/io/src/processing/io/LED.java
gohai: I initially picked set() because there's also a clear() that undos the user's actions and restores the initial value. But brightness() sounds good to me. Perhaps together with reset()?

java/libraries/io/src/processing/io/GPIO.java
benfry: Yeah, I'd skip the microseconds version. We're unlikely to get that kind of timing from the device anyway. Easy enough to add later (and part of the philosophy is to add as necessary—never in anticipation).

gohai: Makes sense! Those two functions are gone in later commits to this PR (below) - setBrightness is now brightness as well.

PR #3999, opened 2015-10-14
Small usability fixes for tiny screens, take 2

gohai:
This implements @jamesjgrady's suggestion from #3992.

### DIFF
--- a/app/src/processing/app/ui/EditorState.java
+++ b/app/src/processing/app/ui/EditorState.java
@@ -161,6 +161,10 @@ public class EditorState {
     int defaultWidth = Preferences.getInteger("editor.window.width.default");
     int defaultHeight = Preferences.getInteger("editor.window.height.default");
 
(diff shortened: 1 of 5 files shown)


gohai:
Screenshots attached. (The less wide editor window looks so much better to me - perhaps it would really make sense to resize the window proportionally if it the default doesn't fit onto the screen? Maybe we could spawn it at 0,0 in this case too. This way the play/stop buttons are most likely to be unobstructed.)

2015-10-14-155244_800x480_scrot
2015-10-14-155014_800x480_scrot

gohai:
Another alternative would be to keep the initial behavior, but make it so that the display window spawns right next to the stop button. So like my initial patch, but keeping the 66px from the top - just pushing it more to the right so that the user can click stop.

benfry:
Going with a hybrid of the two PRs for 3.0.1.

PR #4004, opened 2015-10-15
Turn off fixed rate scheduling in OpenGL

JakubValtar:
After some testing, it appears that fixed rate scheduling tries to catch
up later when it runs slower than desired, creating weird speedups when
the load decreases. Turning it back off.

### DIFF
--- a/core/src/processing/opengl/PSurfaceJOGL.java
+++ b/core/src/processing/opengl/PSurfaceJOGL.java
@@ -407,7 +407,7 @@ public class PSurfaceJOGL implements PSurface {
 
 
   protected void initAnimator() {
-    animator = new FPSAnimator(window, 60, true);
+    animator = new FPSAnimator(window, 60);
     drawException = null;
     animator.setUncaughtExceptionHandler(new GLAnimatorControl.UncaughtExceptionHandler() {
       @Override


gohai:
@xranby?

barefists:
I'm having a problem where the FPS in my sketch rendered in P2D runs at odd framerates. When I set it at 60, it runs at 32, then when I set it at 30, it runs at 21.3. This doesn't happen all the time - sometimes it runs at the desired FPS, and other times, it seems to throttle at a very specific number mentioned above.

Googling about this problem in Processing yields no result, however when I search with regards to JOGL, I ran across this post: http://stackoverflow.com/questions/12936690/why-does-this-code-sometimes-report-21-3-fps-when-i-ask-for-30-fps

It caught my eye because of the oddly specific throttled FPS number "21.3". The thread seems to point to inaccuracy of java.util.Timer.schedule(), which may be fixed by turning on fixed-rate scheduling in FPSAnimator. I know there's a good reason why you turned off fixed-rate scheduling in JOGL, but can you suggest a workaround for my case? Thanks!

xranby:
As JakubValtar suggested, the best way to get non-odd numbers such as V-Sync to fractions of display refresh rate (15, 30, 60) then processing should use Animator instead of FPSAnimator. Using a class such as the standalone https://github.com/LWJGL/lwjgl/blob/master/src/java/org/lwjgl/opengl/Sync.java can of course be combined with the Animator to slow it down to the expected framerate, you do this by then running Animator with sync set to 0 and then use the Sync class to throttle it down to the desired FPS.

barefists:
So my current workaround hack is to set frameRate at 70 and swap buffer to 1, which brings it back down to my target FPS of 60 on my current monitor. Very hacky, and it'll probably break if I encounter a monitor with another refresh rate, but I don't have much of a choice at the moment. Do let me know if there are any changes regarding this topic. :)

JakubValtar:
Thanks guys, I will give Animator a shot in combination with LWJGL Sync and hopefully it will improve your frame rate.

I think I got it working before, but didn't see any noticable difference from what we have (except for adding bunch of new code).

JakubValtar:
Just a quick question for @xranby, do you think it is possible to use Animator + Sync together with VSync? Right now we set setSwapInterval as follows:
> 60: setSwapInterval(0)
60-30: setSwapInterval(1)
< 30: setSwapInterval(2)

xranby:
i think Animator + Sync can work well together with VSync. sync calculates the optimal time to delay the draw frame to archive the desired FPS.

The Sync class needs a JogAmp provided accurate timer function:
JogAmp gluegen have the currentTimeMicros that can be used in the Sync class.
https://jogamp.org/deployment/v2.3.2/javadoc/gluegen/javadoc/com/jogamp/common/os/Platform.html

Issue #3938, opened 2015-09-30
'key' variable is always uppercase in FX2D renderer


tsulej:
When using FX2D renderer, 'key' variable in keyPressed() method returns uppercase character.

test code:
void setup() {
  size(500,500,FX2D);
}
void draw() {}
void keyPressed() {
  println(key);
}

JakubValtar:
This is how FX key events work. Use `keyTyped()` if you need correct case.

iamutkarshtiwari:
I have written a fix for this issue. Could anyone here please guide me how to create a patch and a pull request?

iamutkarshtiwari:
### Patch(PR) to fix this bug
#### https://github.com/processing/processing/pull/4006

Aeoll:
Just ran into this - it would be nice if the behaviour was documented since it causes inconsistencies across renderers. 

```
void keyPressed() {
    switch (key) {
      case 'a' :
        // do something
        break;
    }
}
```
The user should surely get some kind of warning when trying to do this in FX2D?

PR #4006, opened 2015-10-15
'key' variable is always uppercase in FX2D renderer

iamutkarshtiwari:
This patch is fix to this issue -> https://github.com/processing/processing/issues/3938

Error - "'key' variable is always uppercase in FX2D renderer"

With this patch, key variables will be displayed in the console with changed cases depending upon the status of the CapsLock and Shift keys.


benfry:
This is a good workaround to use in your own code, but it's not going to work across all operating systems, keyboard layouts, languages, input methods, etc, so we can't use this.

Issue #3983, opened 2015-10-10
Error console will stop refreshing itself until sketch is closed and reopened


serebit:
This persists even as the user edits the sketch.

JakubValtar:
Sorry, this is like writing "my computer is broken" to a tech support.

Could you please provide some details? What do you mean by "error console"? There is "Console" pane and there is "Errors" pane.

I need to know: which of the two you are talking about, what it is doing, what it should be doing, and possibly when it happens and how can I reproduce it.

Thanks.

gingerdeadshot:
The errors pane. It should be refreshing itself as I type, to both remove errors that no longer apply and to display errors that currently apply. Occasionally, the errors pane will stop automatically refreshing. The only way to get it to refresh is to either close and reopen the sketch, or to switch to a different tab within the sketch. Upon returning to the original tab, the error checking refreshes normally.

JakubValtar:
Are you using Processing 3.0 final? There has been a lot of work done on it between beta and final.

gingerdeadshot:
Yes. I make sure to keep up with the updates. In fact, the final 3.0 has a
lot more bugs in my case than 3.0 beta 6 did.

serebit:
No idea if this is still a problem, but given it's been almost 5 years, probably not. Closing.

PR #4009, opened 2015-10-16
Add error checker document listeners to all tabs

JakubValtar:
Check if #3983 is still an issue


benfry:
Not sure if this will do anything—it may use the same `Document` object across all tabs. (But that code is really old and I didn't look into it closely.)

JakubValtar:
I checked all the accessors, each tab has its own SyntaxDocument. This solves an issue that all tabs except for the first one won't trigger error checker on typing, it was triggered only when you switched tabs. The reason was missing listeners. With this fix, each document should get a listener either on startup as soon as the error checker is initialized or when the document is set.

benfry:
Cool, thanks.

Issue #4017, opened 2015-10-18
Tweak Mode: Some numbers ignored in second tab.


erniejunior:
While trying out the new pull request by @galsasson https://github.com/processing/processing/pull/4014 I found another bug in Tweak Mode when using more than one tab. Some decimals in my second tab were mysteriously ignored. After digging a bit around in [SketchParser.java](https://github.com/processing/processing/blob/master/java/src/processing/mode/java/tweak/SketchParser.java) I found out that they are falsely accused of being inside one of the ignored functions (setup or settings). After looking a bit deeper into that ignore mechanism, which is based on ranges over characters I figured, that it is tab agnostic:

Assume our setup function spans char 5 to char 80 in the first tab, then numbers between char 5 to char 80 are ignored in every other tab! This was the reason why only some of my numbers were ignored and others not.

I am happy to work on a fix of this bug as soon as I get processing up and running in an IDE. I will let you know when I start the actual work so there is no double effort.

Edit: Working on it now ...

galsasson:
Hi @erniejunior, thanks for catching that bug and fixing it. Its impossible to see the changes you made to SketchParser.java, can you provide a pull request that doesn't replace the whole file?

erniejunior:
Hi @galsasson, I do not know why git does not show the individual changes. Maybe it has something to do with the fact, that I applied an automatic code formatter to comply with the [Style Guide](https://github.com/processing/processing/wiki/Style-Guidelines)? That was probably a stupid idea ...

Basically I added a ``tab`` property to the ``Range`` class and the ``Range.contains`` method now also checks if the tabs are equal. Then I worked my way up to make any necessary changes for the rest of the code to work again. We certainly could use a more elaborate method to filter the actually used numbers and also remove some duplicate code. But for now I wanted to respect your work and make only minimal changes to fix the bug as quickly as possible.

Is this enough information to understand the changes that I made or should I redo them without applying the code formatter?

galsasson:
I would prefer to handle ignoreFunctions the same way I'm handling now commentBlocks instead of introducing 'tab' into the Range class. This would be more efficient and will require less changes to the code while keeping them local to SketchParset.java.
commentBlocks now keeps comments per tab. You will need to change ignoreFunctions to be the same data type, and initialize it correctly.
Then, in 'addAll***Numbers' you can replace:
```java
        if (isInRangeList(start, ignoreFunctions)) {
          // ignore numbers in predefined functions
          continue;
        }
```
With:
```java
[...]

PR #4023, opened 2015-10-18
Improved fix for Issue #4017 where tweak mode ignores some numbers in a second tab

erniejunior:
This is an improved version of my previous pull request after some friendly hints by @galsasson. This one has even less changes and should be more transparent.
We now maintain the ignore-lists for functions in separate lists for the different tabs.

### DIFF
--- a/java/src/processing/mode/java/tweak/SketchParser.java
+++ b/java/src/processing/mode/java/tweak/SketchParser.java
@@ -40,7 +40,7 @@ public class SketchParser {
   List<List<Range>> scientificNotations;
 
   // currently is used to ignore numbers in 'setup' and 'settings' functions
-  List<Range> ignoreFunctions;
+  List<List<Range>> ignoreFunctions;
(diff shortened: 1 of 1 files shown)


benfry:
@galsasson will wait for your initial approval...

galsasson:
There is a warning related to the use of Collections.EMPTY_LIST. I'm not fluent with java, but I think its fine. ?
Other that that it works and fixes the issue, thanks @erniejunior

erniejunior:
Hi, @galsasson the warning is because the EMPTY_LIST is not typed (we do not specify what type of items it contains via generics). Since it is empty it does not really matter. But we could get rid of it by using ``ignoreFunctions.add(new LinkedList<SketchParser.Range>());`` (just create an empty linked list). But I thought that explicitly spelling out that the list is empty improves readability.

benfry:
Thanks all.

Issue #305, opened 2013-02-10
A sketch may not have the same name as an included library


processing-bugs:
_Original author: heue...@gmail.com (June 10, 2010 18:53:12)_

A problem somewhat related to Issue 168, a sketch may not have the same name as an included library.  There is a name collision on the jar file that is created when exporting to applet or application.

Running the applet or application results in a ClassNotFoundException for the sketch class.

_Original issue: http://code.google.com/p/processing/issues/detail?id=266_

JakubValtar:
See also #196 #1058

Issue #196, opened 2013-02-10
Saving sketch with the same name as class/primitive/reserved word won't work


processing-bugs:
_Original author: b...@processing.org (June 07, 2010 01:12:47)_

This bug automatically added from:
http://dev.processing.org/bugs/show_bug.cgi?id=1165

Comment from chandler, 2009-02-18 18:23

using 1.0.1 / osx 10.5
1) make a new sketch, name it "boolean" hit run
or
2) make new sketch, name it PImage, try to load in image, hit run

in both cases you get difficult to debug errors.

_Original issue: http://code.google.com/p/processing/issues/detail?id=157_

JakubValtar:
Related to #305 #1058

benfry:
First step here (and in the others) is figuring out how to create, maintain, and/or automatically/dynamically generate a list of things for which we need to avoid collisions.

iamutkarshtiwari:
One thing we can do is list the names of all the .java files in a check.txt file alongwith all the reserved words in Java language and Processing Language as well.

iamutkarshtiwari:
I am working currently working on writing a script which can generate a list of all "reserved keywords" in Processing and Java language and write it into a text file using regular expression.

iamutkarshtiwari:
https://github.com/processing/processing/issues/305

PR #4033, opened 2015-10-20
Saving sketch with the same name as a class #196

iamutkarshtiwari:
This patch is fix to the issue -> #196, #305

I have generated a list of all the possible reserved keywords( in Processing and Java language) in a **_"reservedKeywords.txt"**_ file in **_build/shared/**_ (which will be later copied to **build/linux/work/** during setup build) folder through a script(https://github.com/iamutkarshtiwari/ReservedWordsGenerator). Whenever the user will save a sketch with a new name, **_sanitizeName()**_ will search for that name in 'reservedKeywords.txt" file. If the name exist in that file, it will prompt the user to save with any another filename but inbuilt class name.


iamutkarshtiwari:
I also cleaned up a lot of trailing whitespaces in build.xml file.

iamutkarshtiwari:
You can test this patch by saving the sketch with any Processing class names (like PImage, PApplet, PSurfaceNone, etc) or with reserved Keywords(like abstract, int, long, boolean, etc) in Java.

alignedleft:
@iamutkarshtiwari This is a needed feature. I stumbled on this the other day, unintentionally saving a sketch using a name of an existing class or method.  I don't remember the specific outcome, but things went haywire. Thanks for looking at this!

iamutkarshtiwari:
@alignedleft Did you test this patch?

alignedleft:
@iamutkarshtiwari No, that’s not really my role. I was just voicing support for the idea.

iamutkarshtiwari:
@alignedleft Oh sorry. :)

GKFX:
@iamutkarshtiwari Some little criticisms:
Why are keywords from `processing.app` reserved? I tested a sketch called `CopyAsHtmlAction` (without applying your patch) and it runs fine.
There's a space missing before the bracket in `Name '%s' is a not allowed(reserved keyword)`.
You've forgotten to close or try-with-resources your `FileReader` and `BufferedReader`.

PR #4035, opened 2015-10-20
Regenerate Javadoc (adds IO library)

gohai:
Not sure if these get customarily rebuild right before the release anyway (in this case, please ignore this PR), but if not: this adds the IO library to the pre-built Javadoc.


benfry:
These are actually moving to a different repo /cc @REAS

Issue #3965, opened 2015-10-06
File modified externally gives infinite pop-ups


HenriAugusto:
Hi, how are you?

Today i've moved a .pde file in my sketch folder. I was happy to see that Processing automatically detect changes and offers a reload.

![reload issue](https://cloud.githubusercontent.com/assets/11790621/10314179/d3fb6d32-6c29-11e5-8dc1-cbca1fcbcc07.png)

The problem is that i had unsaved changes and when i clicked "no" that same popup simply opened again. I had to remove the file i've added so this window could desappear, then i could save and finally move it again to reload the sketch. Fortunetaly i've only added a file and could simply delete it. 

If the user had changed an file in a external text editor it would be hard to remember what was changed to unchange it, so to make this pop-up disappear. If by any means you couldn't unchange the file, then you would have to click yes and lose the unsaved changes you've made in the processing IDE.

**edit:** i've just tested and when you modify some text file, that pop-up opens only one time. Guess the problem is that the IDE constantly check for new files but don't memorize your answers and compare them with the new files it's detecting.

---

Also, please don't get me wrong, i really love your work and i don't want to be annoying. I know Processing is for sketching but i wanted to talk a little more about issue #3962. It was closed so fast that i couldn't even answer.

Maybe simply adding a way to scroll on the tab list with the mouse wheel would solve this problem.

![processing tab scroll](https://cloud.githubusercontent.com/assets/11790621/10314432/1043fda8-6c2b-11e5-91db-f595b9e1bfd6.png)

Thanks for your attention and patience.

**edit 2:** I didn't knew i could post on closed issues. Sorry, i'm new on github!

PR #4037, opened 2015-10-20
Bugfix infinite pop ups

iamutkarshtiwari:
This patch fixed this issue -> https://github.com/processing/processing/issues/3965

I have added a **_doReloadPrompt boolean flag**_ to check if the popup has been prompted atleast once if the user externally **_modifies/deletes**_ the main file. If the user selects **_'No'**_ option, the popup closes as expected(no infinite popups). If the user further modifies the file externally, the prompt will appear normally.


iamutkarshtiwari:
You may test the infinite popups by saving a new sketch and then ***deleting it's .pde*** file while Processing running alongside. After deletion, when you return back to the Processing editor window a popup appears, click ***"No"*** on which prompts the popup again and again.

benfry:
Thanks for this, though we're opting for the solution found in https://github.com/processing/processing/pull/4762

Issue #3651, opened 2015-08-17
Pasting text into PDE results in "Clipboard does not contain a string"


JakubValtar:
Sometimes when I copy the sketch code from PDE by pressing Ctrl+A and Ctrl+C, then close the PDE, open it again and paste the code back by pressing Ctrl+V, nothing is pasted and I get "Clipboard does not contain a string" error message.

benfry:
Added debug messages, see what it prints and report here.

JakubValtar:
Can't reproduce right now, but it happens from time to time. Will post here when it happens again.

JakubValtar:
Good, got it right now!
```
Clipboard does not contain a string
Exception in thread "AWT-EventQueue-0" java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.util.TimSort.mergeHi(TimSort.java:895)
	at java.util.TimSort.mergeAt(TimSort.java:512)
	at java.util.TimSort.mergeForceCollapse(TimSort.java:453)
	at java.util.TimSort.sort(TimSort.java:250)
	at java.util.Arrays.sort(Arrays.java:1438)
	at sun.awt.datatransfer.DataTransferer.setToSortedDataFlavorArray(DataTransferer.java:2410)
	at sun.awt.datatransfer.DataTransferer.getFlavorsForFormatsAsArray(DataTransferer.java:873)
	at sun.awt.datatransfer.SunClipboard.getAvailableDataFlavors(SunClipboard.java:172)
	at processing.app.syntax.JEditTextArea.paste(JEditTextArea.java:1916)
	at processing.app.ui.Editor$PasteAction.actionPerformed(Editor.java:1536)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2346)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
	at javax.swing.AbstractButton.doClick(AbstractButton.java:376)
	at javax.swing.AbstractButton.doClick(AbstractButton.java:356)
	at javax.swing.plaf.basic.BasicMenuItemUI$Actions.actionPerformed(BasicMenuItemUI.java:802)
	at javax.swing.SwingUtilities.notifyAction(SwingUtilities.java:1663)
[...]

PR #4040, opened 2015-10-21
Pasting text into PDE results in "Clipboard does not contain a string…

iamutkarshtiwari:
This patch is fix to this issue -> https://github.com/processing/processing/issues/3651

With this patch the user will be able to retain the copied content on the **_system clipboard**_ even after the Processing environment is closed and reopened. For this I have created a **_"localClipboard.txt"**_ file in **_"build/shared/lib/"**_ directory which will save the contents in itself whenever copy() method is called. It will automatically clean itself(by moving it's contents to system clipboard) whenever paste() method is called in order to avoid further conflicts.

However it gives preference to SystemClipboard over LocalClipboard.

### DIFF
--- a/app/src/processing/app/syntax/JEditTextArea.java
+++ b/app/src/processing/app/syntax/JEditTextArea.java
@@ -20,6 +20,8 @@ import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.Vector;
 
+import java.io.*;
+
 import javax.swing.event.*;
 import javax.swing.text.*;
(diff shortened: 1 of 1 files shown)


benfry:
Thanks, though I think this only works around the underlying problem without fixing it. Closing this one for now because it's hopefully resolved in more recent Java release. 

If not, please re-open a fresh issue with details on how to reproduce in the Processing 4 repo: https://github.com/processing/processing4/issues/new

PR #4042, opened 2015-10-22
Error checker update

### LINKED ISSUE #4034: switch on Strings
Hi,

I've tried to switch on String variables whith Processing 3. According to the documentation, this might not work : it should only be right with int, char or byte. Processing underlines the variable in red. Clicking on the variable, the message "Cannot switch on a value of type String. Only convertible int values, strings or enum variables are permitted."  is displayed.

Nevertheless, if I try to lauch the code, it works as expected.

Can you please explain to me : 
- if this is a bug ?
- if the documentation is right ?
- what is the precise meaning of the error message ? What is this "string variable" permitted ?

Thanks,
Samuel

EDIT : I use JDK 8, which could explain why it works. If so, shouldn't we update the documentation and/or the software to indicate when it will work ?

JakubValtar:
Fixes #4034, does not guarantee support for all Java 7 features (only some), but warns about lambdas not being available. Lambdas are not supported, because they make ANTLR angry.

While I was at it I also simplified error checking once again and deleted lots of unused code. CompilationChecker was reduced to one method and one inner class, which I moved to error checker itself.

I enabled some warnings which should help beginners with basic programming problems:
- no effect assignment `x = x;`
- accidental boolean assignment instead of comparison `boolean x = true; if (x = false) return;`
- null reference `vec = null; vec.copy();`
- potential null reference (under some conditions, your code can result in NPE here)
- redundant null check (it is guaranteed that variable is not null here, no need to check it)
- ~~redundant type arguments (you can use diamond here, e.g. new ArrayList<>())~~ ignored, ANTLR can't do diamonds
- unused labels, unused local variables, unused parameters in functions, unused private members
- unused object allocation `new PVector(); // no assignment and no method call`


benfry:
Nice.

Issue #3926, opened 2015-09-30
show size of download in the Contribution Manager


benfry:

iamutkarshtiwari:
I am currently working on this enhancement.

iamutkarshtiwari:
Adding this functionality introduces a bit ***lag*** in Contribution Manager b/c everytime we click a contributions, the will need fetching and ***calculating the size*** of the download file which takes a little time. Is this lag be acceptable?

benfry:
That should be handled on the server side instead, when the contrib list is built. That's a lot of connections and a lot of lag. @prisonerjohn might have thoughts on that.

iamutkarshtiwari:
@benfry I somehow figured out how to handle that so that it may not end up introducing a lag in the CM. The above patch handles that problem and also integrates the "Download FileSize" feature into CM.

iamutkarshtiwari:
@benfry After the above patch, the ***show dowload size*** feature will show up in the CM as shown in the screenshot below.
![screen](https://cloud.githubusercontent.com/assets/6258810/10713928/e1f02120-7af8-11e5-82d7-dd6391caba77.png)

DarkPrince304:
Hello,

You mentioned that something had to be done on the server-side of the contribs file. Did you mean to make a size parameter for each entry and then retrieve it like everything else is retrieved on client-side (PDE Manager) ?

Like this:

"
library
minRevision=0
name=Simple Multi-Touch (SMT)
maxRevision=228
sentence=Multi-touch prototyping and development made simple.
url=http://vialab.science.uoit.ca/SMT/
prettyVersion=4.2
paragraph=SMT provides back-end device support for Tuio, Windows Touch* ( requires [VC2012 Redistributable](http://www.microsoft.com/en-au/download/details.aspx?id=30679) ), SmartSDK*, Leap Motion*, and Mouse Emulation. *Windows-Only
version=23
authors=[Erik Paluka](http://www.erikpaluka.com/), [Kalev Kalda Sikes](http://vialab.science.uoit.ca/portfolio/kalev-kalda-sikes), [Zachary Cook](http://vialab.science.uoit.ca/portfolio/zachary-cook), [Dr. Mark Hancock](http://markhancock.ca/), and [Dr. Christopher Collins](http://vialab.science.uoit.ca/portfolio/christopher-m-collins)
download=http://vialab.science.uoit.ca/smt/dl/SMT.zip
type=library
id=098
size=1024  <----------------
categories=I/O
"

[...]

PR #4054, opened 2015-10-23
Feature- Show size of download in the Contribution Manager #3926

iamutkarshtiwari:
This patch is feature-enhancement to this request -> https://github.com/processing/processing/issues/3926

With this patch, the user will be able to check the ***DownloadSize*** of the ***selected(and available)*** contribution in the bottom left ***StatusPanel***.

To test this patch, select the contribution from the list in CM whose details you want to see. After ***2-3 seconds(average time taken by HTTP to calculate the file size)*** you will see the ***download size*** below the ***Install button***. :)

### DIFF
--- a/app/src/processing/app/contrib/StatusPanel.java
+++ b/app/src/processing/app/contrib/StatusPanel.java
@@ -2,19 +2,15 @@
 
 /*
   Part of the Processing project - http://processing.org
-
   Copyright (c) 2013-15 The Processing Foundation
   Copyright (c) 2011-12 Ben Fry and Casey Reas
-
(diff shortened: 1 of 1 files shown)


iamutkarshtiwari:
After the above patch, the feature will show up in the CM as show in the below screenshot.
screen

benfry:
[As I mentioned earlier](https://github.com/processing/processing/issues/3926#issuecomment-150262207), this is something that needs to be handled in the contributions file on the server-side, not in the Manager inside the PDE. The result of this patch would be ~100 connections to other servers each time you open the Contribution Manager.

Issue #3706, opened 2015-08-24
CM - Icon instead of an "X" for the "could not connect" message


Akarshit:
@alignedleft 
This is how the error appears when there is no connectivity. 
![screenshot from 2015-08-24 19 30 44](https://cloud.githubusercontent.com/assets/7762605/9442019/f77c0672-4a96-11e5-9220-f664b4168a43.png)  
The closing button contains an "x" right now. I think there should be a icon rendered there. Can you please provide the icon for the button.

iamutkarshtiwari:
With the above PR, the CM "Could not connect" message will show the following icon in the below image.
![cross](https://cloud.githubusercontent.com/assets/6258810/10706712/d40b320e-7a06-11e5-9e75-f45bf913874a.png)

benfry:
Incorporated for 3.0.2.

PR #4055, opened 2015-10-23
CM - Icon instead of an "X" for the "could not connect" message #3706

iamutkarshtiwari:
This patch is a fix to this issue -> https://github.com/processing/processing/issues/3706

With this patch, the CM "Could not connect" message will dispaly the cross icon as shown in the image below.
cross


build/build.xml
```diff
@@ -628,6 +628,9 @@
       <!--<fileset dir="shared" includes="launch4j/**" />-->
       <fileset dir="shared" includes="lib/**" excludes="lib/fonts/**" />
       <fileset file="shared/revisions.txt" />
+
```
iamutkarshtiwari: I have added two spaces here to avoid merge conflicts with my other pending Pull Requests which modifies this same file.

build/build.xml
```diff
@@ -774,6 +777,9 @@
       <fileset dir="shared" includes="lib/**" excludes="lib/fonts/**" />
       <fileset dir="shared" includes="modes/**" />
       <fileset file="shared/revisions.txt" />
+
+
```
iamutkarshtiwari: I have added ***two spaces*** here to ***avoid merge conflicts*** with my other pending Pull Requests which modifies this same file.

build/build.xml
```diff
@@ -1062,6 +1068,9 @@
...
```
iamutkarshtiwari: I have added ***two spaces*** here to ***avoid merge conflicts*** with my other pending Pull Requests which modifies this same file.

Akarshit:
It would be awesome if @alignedleft could provide icon for this.

alignedleft:
I'd like to make sure @jamesjgrady reviews and gives the OK for this approach first. Then I believe he already has an icon that could be adapted here.

benfry:
We can't use this PR as-is, it installs the image in a different location from the other image assets that are used elsewhere in the Contribution Manager, and also doesn't have retina support. To implement this properly, it should use the same methods as all the other GUI elements in the CM.

iamutkarshtiwari:
@benfry  How to implement the retina support ?

benfry:
Look at the methods used for all the other GUI images in the Contribution Manager code.

iamutkarshtiwari:
@benfry https://github.com/processing/processing/pull/4096

PR #4073, opened 2015-10-27
Add Turkish to the list of languages

crazymaster:


benfry:
Thanks for the catch.

PR #4079, opened 2015-10-30
Workaround for JRE bug freezing the PDE

JakubValtar:
Because of a JRE bug, code completion will occasionally freeze the PDE. Never seen this one before, however, it happened to me three times in a row while I was giving a lecture today and I had to kill the PDE and write all the code again. I can reproduce it back to beta 7. It is caused by JIT compiler, workaround is to disable it for this particular method. See https://bugs.openjdk.java.net/browse/JDK-8060036 and http://kingsfleet.blogspot.com.br/2014/11/but-thats-impossible-or-finding-out.html

### DIFF
--- a/build/windows/config.xml
+++ b/build/windows/config.xml
@@ -43,10 +43,10 @@
     <opt>-Djna.nounpack=true</opt>
     <!-- starting in 3.0, require Java 8 -->
     <minVersion>1.8.0_51</minVersion>
-	<!-- TODO: remove the line below as soon as we are on 1.8.0_60 or newer,
-	 see https://bugs.openjdk.java.net/browse/JDK-8060036 and
-	 http://kingsfleet.blogspot.com.br/2014/11/but-thats-impossible-or-finding-out.html -->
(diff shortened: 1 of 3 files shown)


benfry:
Is this only on Windows?

JakubValtar:
@benfry JDK bug is os generic. It's difficult to reproduce on demand, because it depends on when JIT compiles `javax.swing.text.GlyphView#getBreakSpot` method.

I can sometimes reproduce it by typing (not pasting) this line:
```processing
PFont.list();
```
then starting the sketch and then clicking between `l` and `i` and invoking suggestions with CTRL+Space. If the bug was triggered, no suggestions will show and PDE will start freezing and glitching out in a few seconds. Sometimes you will get just
```
Exception in thread "AWT-EventQueue-0" java.lang.ArrayIndexOutOfBoundsException
```
and PDE will remain operational.

When the bug happens, it looks like this:
glitch

JakubValtar:
Friend was asking me if it's normal that Processing is freezing regularly. Replacing processing.exe launcher by the one which uses this workaround solved it for her. She also uses Windows.

Issue #4082, opened 2015-10-31
CM - After removing a contribution the updates tab becomes empty.


Akarshit:

benfry:
Incorporated your fix for the next release (whatever comes after 3.2.2).

PR #4083, opened 2015-10-31
Removing contribution does not removes all from the updates tab

Akarshit:
Fix for #4082


Akarshit:
This will be a temporary fix.
Will have to go deeper into the CM maze to solve this.

Issue #4084, opened 2015-10-31
Up-to-date status disappears after filter is removed.


Akarshit:
To reproduce - 
1. Search a library which needs update using the filter.
2. Update the library, the status is changed.
3. Clear the filter, the status is changed back to needs-update.

benfry:
Changes merged for the release after 3.2.2.

PR #4086, opened 2015-10-31
Fixes the removal of redundant contribution and update related issues

Akarshit:
FIx for #4082 , #4084


benfry:
But from the comments, this appears to introduce a new (major?) bug:
```
 // TODO: Every contribution is getting added twice
// and nothing is replaced ever.
```
or was that there before?

Akarshit:
No that was already happening before, I just added the comment so that it can be looked into.

Issue #4085, opened 2015-10-31
setting a bad font/size in preferences.txt causes a crash on startup


jsphsnkr:
I am unable to open any version of the Processing IDE without encountering the error message in the attached image.

The problem started after I tried to change the editor font in the 2.2.1-windows64 version, a string of error messages were printed to the console and the editor pane froze up. Since closing this instance I have not been able to open any version of the IDE.

Apologies for any lack of clarity in my description! 

Joe
![seriousproblem](https://cloud.githubusercontent.com/assets/12673391/10866201/0e956d60-801d-11e5-9936-b86430086dc0.png)

Akarshit:
@NervousBearKing somehow you seem to have set the font size to 0. A possible solution is to find the `preferences.txt` file on your system and set the `editor.font.size=12`. This will fix the issue. Let me know if you need any help.

This has been fixed for future in #4087

NervousBearKing:
@Akarshit I found the preferences.txt; editor.font.size was set to 12, but the editor.font.family was set to the mysterious "White Rabbit" font I had tried to select (somewhat poetic), set it back to Monospaced and the problem has gone away.

Thanks for your help!

NervousBearKing:
@Akarshit PDE sucessfully opened once; upon closing, editor.font.family was set back to "White Rabbit" in preferences.txt, presumably by PDE itself. Now when I change it back the original issue persists.

NervousBearKing:
@Akarshit Issue is solved, I must have messed something up in preferences.txt, deleting it has solved the issue. Thanks again.

benfry:
Can't reproduce with the current code; hopefully we've long since fixed this.

PR #4087, opened 2015-10-31
Font size cannot be zero

Akarshit:
Fix for #4085


benfry:
Need to fix the source of this problem; rather than hiding the effects of it.

PR #4095, opened 2015-11-02
Cleaned the CM code a little bit and did some of the TODOs

Akarshit:


app/src/processing/app/contrib/ContributionTab.java
```diff
@@ -92,10 +89,9 @@ public boolean matches(Contribution contrib) {
       }
     };
 
-    contribListing = ContributionListing.getInstance();
     statusPanel = new StatusPanel(this, 650);
     contributionListPanel = new ListPanel(this, filter);
-    contribListing.addListener(contributionListPanel);
+    ManagerFrame.contributionListing.addListener(contributionListPanel);
```
benfry: This excessive de-referencing is bad form. We're not going to trade a single declaration for static access to a double-dereferenced method, requiring even more typing.

benfry:
Please do these in smaller pieces. There are issues throughout, but a huge patch like this is a mess to understand and incorporate.

PR #4096, opened 2015-11-02
CM - Icon instead of an "X" for the "could not connect" message #3706 #4055

iamutkarshtiwari:
@benfry  I placed the icons in their correct contribution assets location and it uses the same methods as all the other GUI elements in the CM. If there are any other shortcomings please let me know.


benfry:
Thanks for updating this patch, much appreciated.

Issue #4094, opened 2015-11-02
spacing of the updates number/circle in the lower right is off


jamesjgrady:
@benfry can the orange circle be slightly bigger or moved so the update number does not hit the edge of the circle. (see screen shots with suggested revision)

Current
![current-update](https://cloud.githubusercontent.com/assets/2600893/10887094/4df9c74e-8153-11e5-9316-1ff6b6439482.png)

Revised
![revised-update](https://cloud.githubusercontent.com/assets/2600893/10887095/4dfdb89a-8153-11e5-9c89-7333d052a5c5.png)

PR #4097, opened 2015-11-02
Added a padding to the text along with spaces.

Akarshit:
Fix for #4094 
screenshot from 2015-11-02 22 47 23


benfry:
Wound up checking in a variant of this over in https://github.com/processing/processing4/issues/201, though we're likely to go a different direction for the 4.x interface.

Issue #4098, opened 2015-11-03
processing-java violates "Rule of Silence" by printing debug information to stdout


tyukiand:
Processing is great if one wants to take some input data and transform it into an image or an animation. Moreover, I think it can be also very useful for the other direction, namely: take an image as input, analyse it, print the result as text. However, `processing-java` always produces some debug output, which has to be removed either by stream redirection or some other tools like `awk` or `sed`.

### Example
```java
void setup() {
  /* load some images, grab pixels, do some calculations */
  println("Very useful result = " + 42);
  exit();
}
```

If I call it as follows:

```bash
processing-java --sketch="nameOfTheSketch" --run
```
the expected output would be:
```bash
Very useful result = 42
```
but the actual output is:
```bash
Listening for transport dt_socket at address: 8651
Very useful result = 42
Finished.
```

#### Problem
There are two unexpected lines printed to stdout: the first line is the cryptic message "Listening for transport dt_socket at address:" followed by a random 4-digit number. The last message is always trivially true. 

### Proposed solutions

**Short solution** 
Eliminate both messages altogether. In `processing/java/src/processing/mode/java/Commander.java` delete the line
```java
  systemOut.println("Finished.");
```
and in `processing/java/src/processing/mode/java/runner/Runner.java`
replace the line
```java
String jdwpArg = "-agentlib:jdwp=transport=dt_socket,address=" + portStr + ",server=y,suspend=y";
```
by 
```
String jdwpArg = "-agentlib:jdwp=transport=dt_socket,address=" + portStr + ",server=y,suspend=y,quiet=y";
```

This solution might break some existing code (maybe someone already uses this in a script, and drops the first line of the output).

**Long(er) solution**
Add a  `--quiet` flag to `processing-java`. This is not complicated either, I've implemented it here:

https://github.com/tyukiand/processing/commit/3cfa89c4e360711dd4ac665f224771648744f309 

(link to commit)

Now, one can call `processing-java` as follows:
```bash
processing-java --sketch="nameOfTheSketch" --quiet --run
```
and it would produce only:
```bash
Very useful result = 42
```

This does not break any existing scripts that use `processing-java`, but it would make the STDOUT more useful than it is right now.

benfry:
Fixed for 3.0.2, see discussion in #4103.

PR #4099, opened 2015-11-04
Implemented '--quiet' flag for processing-java.

tyukiand:
(see issue #4098)

The --quiet flag suppresses the output of the dt_socket port number.
It also suppresses the final "Finished."-message.
The default behaviour of processing-java has not been altered.
A brief "--help"-entry for "--quiet" has been added.
All other components have not been altered.


tyukiand:
I've just noticed that changes to JavaEditor.java were not really necessary. I have another, more "environment-friendly" proposal that achieves the same, but does not change signatures of any existing methods.

PR #4103, opened 2015-11-04
Implemented '--quiet' flag for processing-java (2nd attempt, fewer changes).

tyukiand:
Implemented a `--quiet` flag for processing-java.
Issue: #4098.

If processing-java is called with the `--quiet` flag, the dt_socket port number and the "Finished." message are not printed to STDOUT.

I added an optional `boolean quiet` parameter to three methods in `JavaMode.java` and `Runner.java`, as well as few lines to `Commander.java`, everything else is left untouched (in particular, this time the code in JavaEditor.java is not affected).
All methods that existed previously still have exactly the same signature as before.
The default behaviour of processing-java has not been altered.

### DIFF
--- a/java/src/processing/mode/java/Commander.java
+++ b/java/src/processing/mode/java/Commander.java
@@ -58,6 +58,7 @@ public class Commander implements RunnerListener {
   static final String noJavaArg = "--no-java";
   static final String platformArg = "--platform=";
   static final String bitsArg = "--bits=";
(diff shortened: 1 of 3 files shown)


benfry:
Simpler, I'm just making it run quiet all the time for the next release (3.0.2). No reason to have that debug info in there. Thanks for finding how to shut it off and submitting the patch.

PR #4108, opened 2015-11-06
Add warning when drawing with invalid shape constant.

### LINKED ISSUE #4107: Remove unused PConstants
`LINE_STRIP` and `LINE_LOOP` are no longer valid inputs to `beginShape()`. Do they serve another purpose within Processing, or are they safe to remove?

The following snippet draws nothing to screen, where leaving out the `LINE_STRIP` parameter draws as expected. It would be nice to prevent that mistake by removing the option.
```processing
  beginShape(LINE_STRIP);
  vertex(0, 0);
  vertex(50, 50);
  vertex(50, 100);
  endShape();
```

sansumbrella:
Previously, Java2D renderer silently fails to render `LINE_LOOP` and `LINE_STRIP` shapes. This adds a runtime warning when an invalid shape type is specified for the renderer.

Closes #4107

### DIFF
--- a/core/src/processing/awt/PGraphicsJava2D.java
+++ b/core/src/processing/awt/PGraphicsJava2D.java
@@ -864,7 +864,12 @@ public class PGraphicsJava2D extends PGraphics {
         gpath.lineTo(x, y);
       }
       break;
+
+    default:
+      showWarning("Invalid shape type. Try calling beginShape() without a parameter.");
+      break;
     }
+
   }
 
 


sansumbrella:
Things that would improve this:
- a compile-time check warning of the line number (hard).
- More details in the run-time error message (will look at converting the PConstant to string).

JakubValtar:
Hi David, thanks for looking into this. To answer your improvement points:
- If you `throw IllegalArgumentException(String message)` instead of warning, sketch will crash and PDE will highlight the line and display the message. We do both at different places, any preference @benfry?
- You can switch on `LINE_STRIP` and `LINE_LOOP` to give customized message for these two and handle unknown constant in the `default` with a general error message.

Also, we might just want to implement these two instead of printing errors. It's fairly easy on the user side, should be easy on our side too. @benfry?
```processing
// LINE_STRIP
noFill();
beginShape();
vertex(...);
endShape();

// LINE_LOOP
noFill();
beginShape();
vertex(...);
endShape(CLOSE);
```

benfry:
Both `LINE_STRIP` and `LINE_LOOP` should work with the default renderer. Are they not working?

JakubValtar:
They are not.

benfry:
Scratch that; they were removed. They're not even in the [reference](https://processing.org/reference/beginShape_.html) and should probably be taken out of `PConstants`.

Issue #4090, opened 2015-11-01
Linux: Cannot run any sketch


GKFX:
On Linux x64, current version on GitHub, running any sketch in Java mode gives
````
Annotation processing got disabled, since it requires a 1.6 compliant JVM
/tmp/SVGTesting4582962241611943128temp/SVGTesting.java:1: error: The type java.util.Map$Entry cannot be resolved. It is indirectly referenced from required .class files
	import processing.core.*; 
	^
1 problem (1 error)
````
and nothing displays.

GKFX:
This issue persisted back to at least 935dddfbd00088bcdbe226f01d5e068672478775 from 3 July, so I assumed it wasn't a real regression. (It did not affect version 2.21 though.) I therefore deleted `~/.processing` and thoroughly cleaned the git repository, which fixed it.

PR #4114, opened 2015-11-08
Minor fixes to Java Mode

GKFX:
ASTGenerator.java: Stops there being loads of red text (null pointer exceptions) when you open the context menu with the compiler broken as I reported in #4090. (I'm still not sure what caused #4090, but since it went away when I deleted my configuration files I assumed that it wasn't a real bug.)

JavaTextAreaPainter.java: I just noticed the very odd code that had two different checks for the empty string (with one unreliable and using `==`) and thought that this made more sense.


benfry:
Thanks

PR #4134, opened 2015-11-11
Update JNA to 4.2.1

gohai:
JNA 4.2.0 incorporates a fix for Linux multi-arch systems, such as found on Raspbian (Raspberry Pi). The proposed 4.2.1 has an additional fix that is relevant to ARM. JNA 4.2.0 is currently used in the released processing-video, on all platforms.

With the old JNA I am getting the following error when trying to run the OpenKinect library:

    java.lang.UnsatisfiedLinkError: JNA native support (com/sun/jna/linux-arm/libjnidispatch.so) not found in resource path (...)

I *could* fix this by including an updated version of JNA with the very library (that's definitely an option), but I now have the feeling that it might be beneficial to fix this and potentially other libraries by updating core instead. (Updating lib/jna.jar to the proposed version indeed makes the library work.)


gohai:
@benfry Not sure how risky this is, and what exactly in core is using JNA - curious about your opinion.

benfry:
I'm fairly nervous about it... Doesn't 4.x change API versus the 3.x version? I thought I'd try to make this switch once and it broke all our `WindowsPlatform` code.

gohai:
I have a Windows machine that I can test this on early next week - the only source files using JNA are
app/src/processing/app/Platform.java
app/src/processing/app/platform/DefaultPlatform.java
app/src/processing/app/platform/WindowsPlatform.java
app/src/processing/app/platform/WindowsRegistry.java
so should be fairly straightforward to test

gohai:
@benfry I just tried on Windows 10, and it indeed throws an error, so please ignore this PR.

I am getting a `NoClassDefFoundError` on startup for `com.sun.jna.platform.win32.Shell32Util`. This class is now part of the `jna-platform.jar`, a separate (optional) part of JNA, and (no longer?) in the main `jna.jar`. Dropping `jna-platform.jar` into `lib` didn't do anything. If you have any hint as to what to do to get Processing to look for the class in this file I'd be happy to try - otherwise let's just close this on for now!

Best

gohai:
Will revisit some later time, closing

benfry:
If it's just a matter of adding `jna-platform.jar` as well, that's fine. This may be what I ran into last time... 

You just have to modify `build\build.xml` as well as `build\windows\config*.xml` to add the `.jar` for use with ant.

gohai:
Followed up in #4443!

Issue #4135, opened 2015-11-12
Show the current line number and tab location of the cursor in the Editor


iamutkarshtiwari:
There should be a feature in the Editor to display the line number and tab index of the current cursor location.

iamutkarshtiwari:
Working on this.

benfry:
This is the sort of thing we leave out of the PDE on purpose. Advanced users who would want something like this can use another editor (i.e. Sublime) or a full-fledged IDE (Eclipse, IntelliJ).

PR #4136, opened 2015-11-12
Show the current line number and tab location of the cursor in the Ed…

iamutkarshtiwari:
This PR adds a feature to the Processing Editor -> https://github.com/processing/processing/issues/4135

With this feature, the user will be able to see the current line and tab index location of the cursor right below the text editor area in the EditorStatus bar as show in the screenshow below.
scrren

### DIFF
--- a/app/src/processing/app/ui/Editor.java
+++ b/app/src/processing/app/ui/Editor.java
@@ -297,6 +297,9 @@ public abstract class Editor extends JFrame implements RunnerListener {
       String lastText = textarea.getText();
       public void caretUpdate(CaretEvent e) {
         String newText = textarea.getText();
+        //Updates the Cursor Index Location values in the EditorStatus Bar
+        status.updateCursorIndexValues();
+
         if (lastText.equals(newText) && isDirectEdit()) {
           endTextEditHistory();
         }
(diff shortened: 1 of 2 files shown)


iamutkarshtiwari:
@alignedleft  What are your suggestions on this?

alignedleft:
@iamutkarshtiwari I think it needs to be part of a larger conversation about the UI, including the rest of the core team. I like the idea, but I'm not sure this belongs in the IDE. This is a power user feature; will it appeal to beginners (our primary audience)? If someone wants that level of detail, can't they use Eclipse or another editor?

iamutkarshtiwari:
@benfry What do you think about this feature?

Akarshit:
@alignedleft I personally don't think that this feature would be needed as the line number is already being displayed next to the line and also the tab index is rarely needed while developing.

alignedleft:
@Akarshit Thanks. I think I agree. I am not sure how this information is useful for beginners. Isn't the line/tab index only useful when referring a collaborator to a specific point in the code? We are not trying to address that audience here. (And, as you point out, we already have line numbers, which should be enough for our audience.)

iamutkarshtiwari:
@alignedleft It was when I was trying to fix a bug..which required me to track the cursor and caret location. I felt the need to display the line/tab value. It would be useful for developers in case they want to  fix any editor issues.

alignedleft:
@iamutkarshtiwari Of course! Makes sense. But our primary audience for Processing is not developers (who can use Eclipse or other external editors), but beginners.

benfry:
This is the sort of thing we leave out of the PDE on purpose. Advanced users who would want something like this should use another editor (i.e. Sublime) or a full-fledged IDE (Eclipse, IntelliJ).

PR #4140, opened 2015-11-13
Fix typo in PVector.java docs

GKFX:
Fix typo in docs. I don't know exactly how the documentation system works; will this update the website?


shiffman:
@benfry this is correct and can be merged.  @GKFX yes this will update the docs the next time they are generated for the site.

Issue #3996, opened 2015-10-14
OSX: processing-java stealing focus even with --build flag


myrjola:
I want to use processing-java with the --build flag for on-the-fly syntax checking https://github.com/ptrv/processing2-emacs/pull/8.

The problem I'm facing on OSX is that every time processing-java is run it steals focus from the editor, even though no window is created. This makes editing the code as syntax checking is running a frustrating experience. I can reproduce this by running processing-java from the terminal as well.

I found out a fix, which is to add `-Djava.awt.headless=true` to the java run options in /usr/local/bin/processing-java. I didn't notice any issues when running my sketches, but does anyone more knowledgeable in Java settings know what that headless-option could break?

myrjola:
I guess one solution to this problem would be to add a conditional to processing-java, that adds `-Djava.awt.headless=true` to the java command line parameters if `--build` has been used. This will marginally increase the complexity of processing-java, but would otherwise be a simple enough solution.

benfry:
Appreciate you looking into it and finding a fix.

neilpanchal:
```/usr/local/bin/processing-java: line 8: [: too many arguments
/usr/local/bin/processing-java: line 8: [: too many arguments
Listening for transport dt_socket at address: 8439```

I get this error since this change in sublime text build console. Is this related?

**Comment by myrjola** [comment]
@neilpanchal, interesting find. I think I found the explanation and solution for the problem http://stackoverflow.com/questions/13781216/bash-meaning-of-too-many-arguments-error-from-if-square-brackets.

Could you try by applying this diff to `/usr/local/bin/processing-java`

```
@@ -5,7 +5,7 @@
 OPTION_FOR_HEADLESS_RUN=""
 for ARG in "$@"
 do
-    if [ $ARG = "--build" ]; then
+    if [ "$ARG" = "--build" ]; then
         OPTION_FOR_HEADLESS_RUN="-Djava.awt.headless=true"
     fi
 done
```

So just put double quotes around `$ARG`.

neilpanchal:
@myrjola Works perfectly, I just tested it. Thanks! Closing #4119

Issue #4119, opened 2015-11-08
Processing-java stdout fails to output (intermittently)


neilpanchal:
Hi all,

I have found an issue with running processing-java via Sublime Text 3. In version 3.0, everything works as expected and I get the following output:

```
Listening for transport dt_socket at address: 8143
This is Processing version 3.0
Finished.
[Finished in 3.2s]
```

In version 3.0.1, the stdout fails to output intermittently. I can't seem to find any pattern.

Here is the output if it works:

```
/usr/local/bin/processing-java: line 8: [: too many arguments 
/usr/local/bin/processing-java: line 8: [: too many arguments 
Listening for transport dt_socket at address: 8143
This is Processing version 3.0.1
Finished.
[Finished in 3.2s]
```

For now, I have rolled back to version 3.0 as it kind of breaks the build process in sublime text. Any help appreciated :)

Thank you.

PR #4145, opened 2015-11-14
Small fix to processing-java generation

myrjola:
Related to #3996.

Now it doesn't give an error when there is quotation in the command line args.

Related comments:
https://github.com/processing/processing/issues/3996#issuecomment-154859783
https://github.com/processing/processing/issues/3996#issuecomment-155404233

I have tested that this fixes the issue reported in the first comment.


myrjola:
Also fixes the problem described in #4119.

benfry:
Cool, thanks for the help and testing it out.

PR #4155, opened 2015-11-17
Flesh out the Javadoc for PMatrix.

GKFX:
I know that previously you've said that this is an advanced-users-only thing, but it would be nice to make it (reasonably) accessible.  I've ignored the actual translate/rotate/stretch etc. methods, since they're well documented in the normal web reference.


benfry:
Thanks!

PR #4168, opened 2015-11-24
Extended support of SVG format

poqudrof:
Hello Processing maintainers !

I work on a side project called [Soby](https://github.com/poqudrof/Soby) in which I do slideshows based on SVGs, it is based on an older [Sozi](https://github.com/senshu/Sozi). 
I extended a little bit the support of SVGs and now with Java 8 and stuff like the Base64 decoder I have a rather clean solution. 
#### Here are the good parts : 
- Support of embedded images in SVGs, linked images are supported but should not be avoided if possible.
- Text and image locations are identical to SVG visualized in Inkscape. 
  When the font works, it fits perfectly. 
- Low modification of current code except for new parsing functions. 
#### Here are the bad parts : 
- I introduce some code duplication in :

``` java
private void loadBase64Image() 
```
- I have no idea how to handle text quality in a clean way... :

``` java
 public static float TEXT_QUALITY = 1;   
```

Although is seems `2` is a good «guess» for most applications. 
- Inline styles are not supported.
- Not all fonts are supported (depending on your OS).
- Tested only on Inkscape SVGs.

### DIFF
--- a/core/src/processing/core/PShape.java
+++ b/core/src/processing/core/PShape.java
@@ -22,8 +22,23 @@
 
 package processing.core;
 
+import java.awt.Image;
+import java.awt.color.ColorSpace;
+import java.awt.image.BufferedImage;
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Base64.Decoder;
+import java.util.logging.Level;
+import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


poqudrof:
This is now a library : https://github.com/Rea-lity-Tech/SVGExtended

jeremydouglass:
@poqudrof thanks for sharing this! Consider announcing this on the [new forum](https://discourse.processing.org/) -- and if you can, please submit it as a confirming library to the PDE Contribution Manager

https://github.com/processing/processing/wiki/Library-Guidelines

Issue #1492, opened 2013-02-10
clean up 'ant doc' target to remove warnings


processing-bugs:
_Original author: f...@processing.org (December 08, 2012 17:26:38)_

Because we're using a custom doclet, running javadoc on our source with the standard doclet causes hundreds of warnings. Need to sort out a way to ignore these so that the useful warnings can come through.

_Original issue: http://code.google.com/p/processing/issues/detail?id=1454_

benfry:
Fixed for 3.1.1 by https://github.com/processing/processing/pull/4191

PR #4191, opened 2015-12-06
Fix error messages from Javadocs

GKFX:
I'm trying to significantly reduce the number of error messages from the `ant doc` command, so that the genuine errors can be seen and dealt with. Currently, for example, there's an unclosed bold tag over a lot of an online Javadoc page. (It seems to have been fixed in the original files now though.)

This isn't finished yet; I'd like some advice. This first commit was just running [GenerateDescriptions.py](https://github.com/processing/processing-docs/blob/master/java_generate/GenerateDescriptions.py) with some patching/manual editing of files where it was broken. This was so that if I needed to modify and rerun the script or XMLs later, the later commits' diffs wouldn't contain ordinary updates to XML files.

The second commit fixes most of the errors except those in the HTML and syntax categories of Javadoc's Doc-lint tool, and suppresses the production of those messages by the Javadoc tool. It also makes the `@usage` and `@brief` tags display in the proper Javadoc, since that was an option I had when stopping them from being errors; I can hide them if preferred.

This just leaves the syntax errors. As far as I can tell, there's no way to stop Javadoc moaning about `<br />` but keep getting the useful messages about unclosed tags and other real errors. And there are *loads* of messages about `<br />`, making Javadoc's output almost unreadable if you don't suppress them. So: can I
* modify all the XML files to make them say `<br>` as preferred?
* modify the script to convert `<br />` to `<br>`?
* or just leave it alone and keep all syntax errors suppressed?

I'm asking because I don't know what else uses those files and how strict their XML needs to be. I'd also need to wrap code samples in the XMLs in `{@code }` tags where they contain angle brackets (which escapes them) to avoid filling the doc comments in the .java files with `&lt;&lt;=`; can all other users of the XML files handle that sort of Javadoc syntax?

### DIFF
--- a/app/src/processing/app/contrib/ExamplesContribution.java
+++ b/app/src/processing/app/contrib/ExamplesContribution.java
@@ -52,7 +52,7 @@ public class ExamplesContribution extends LocalContribution {
    * exampleLocation directory is compatible with the current mode.
    *
    * @param base
-   * @param exampleFolder
+   * @param props
    * @return true if the example is compatible with the mode of the currently
    *         active editor
    */
--- a/app/src/processing/app/syntax/SyntaxStyle.java
+++ b/app/src/processing/app/syntax/SyntaxStyle.java
@@ -30,9 +30,9 @@ public class SyntaxStyle {
   /**
    * Creates a new SyntaxStyle.
    * @param color The text color
-   * @param italic True if the text should be italics
    * @param bold True if the text should be bold
    */
+// * @param italic True if the text should be italics
 //  public SyntaxStyle(Color color, boolean italic, boolean bold) {
   public SyntaxStyle(Color color, boolean bold) {
     this.color = color;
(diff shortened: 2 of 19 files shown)


benfry:
If you'll remove the 'regenerate javadocs' commit from this PR, then I can merge it. The other stuff might be fine, but re-running those scripts needs to be handled by @REAS, so that's something we should coordinate separately with him.

GKFX:
Okay, I've done that.

REAS:
@GKFX All of the <br> in the XML files are <br /> to work with the system, but nothing has been automated. We've put so much time and energy into getting this to work over the years, but as you've seen it's not finished.

benfry:
@GKFX Thanks for your follow-up. Can you redo your branch so that it no longer has conflicts? Sorry to have missed the window to get it merged earlier.

GKFX:
@benfry I've rebased it.

benfry:
Thanks again!

benfry:
Merged for 3.1.1, fixes https://github.com/processing/processing/issues/1492

Issue #3543, opened 2015-08-06
Font sadness when running PDE from folder with accents


JakubValtar:
1. Rename folder containg processing.exe (or any parent folder) to something containing `ě`
2. Run processing.exe
3. Font sadness happens, PDE does not launch

benfry:
Appears to be a problem with characters outside [this range](https://en.wikipedia.org/wiki/Windows-1252) (on English Windows).

benfry:
Confirmed this to be a launch4j problem... When running from a folder with accents, it cannot find the local JRE located in the "java" folder. 

The specific error is a red herring—it's launching with the JRE installed on the system (if available) or if that's not available, it'll give users this message:

<img width="512" alt="launch4j-problem" src="https://cloud.githubusercontent.com/assets/1623101/9119358/44ab1304-3c44-11e5-86ee-6947b0d8fabd.png">

benfry:
New error message to give folks a hint:

<img width="411" alt="launch4j-problem-improved" src="https://cloud.githubusercontent.com/assets/1623101/9120866/d5bc63f4-3c4c-11e5-95ad-4fd9fc10fc9c.png">

benfry:
Removing the 3.0 final milestone because this is out of our hands. The warning message should at least help people know what's going on and give them a workaround.

benfry:
Adding “known” tag because this is listed in the “known issues” section [in the wiki](https://github.com/processing/processing/wiki/Changes-in-3.0). When this bug is fixed, it should be removed there.

benfry:
To clarify [this comment](https://github.com/processing/processing/issues/3543#issuecomment-128410440) the issue seems to be happening where the character is *outside* the 8-bit version of the platform's default character set. 

That is, if running on a cp1252 machine (i.e. US English), then characters that still fit into 128-255 (for instance, ü) will work fine. But [ě](http://www.fileformat.info/info/unicode/char/11b/index.htm) will not, since it's still outside the 128-255 (non-ASCII) range. 

I reference ASCII in the error message (which only covers 0-127) since it's simpler (some might say more appropriate) than providing the user with a short lecture about ASCII, the evolution of character coding systems, a rant about internationalization, a brief history of MS-DOS code pages inherited by Windows, and so on.

PR #4204, opened 2015-12-16
Fix crashing bugs that happen when the user's name has non-ASCII characters on Windows.

jdf:


benfry:
Is this just when running from Eclipse on Windows?

jdf:
There are two commits; one addressed an Eclipse-on-Windows bug (which could potentially fail on other platforms, too), and the other addresses a crashing bug when running a release build of the PDE.

jdf:
Ping?

benfry:
Thanks for the fix. Does this also cover https://github.com/processing/processing/issues/3543?

jdf:
I can't say offhand; it depends on the codepath font loading uses.

benfry:
(For the record, this doesn't fix #3543, which is based on a launch4j bug.)

jdf:
argh, and thank you!

Issue #4105, opened 2015-11-05
contribution-manager: No visual feedback of update progress


mark-orion:
The contribution-manager should give visible feedback of the update progress. Currently updates appear in the "update" tab, but when clicking "update all" nothing seems to happen although in fact the download and installation runs hidden in the background.

iamutkarshtiwari:
@alignedleft How to reproduce the situation when a update  is available in Contribution Manager provided my installed contributions are already up-to-date ? I need to do this to add the above "visual update progress" features.

alignedleft:
@iamutkarshtiwari Good question. Can you save an older copy of a contribution (like a library) to keep around just for testing purposes?

iamutkarshtiwari:
@alignedleft I dont have any older copy of any contribution because I recently setup my Processing build. Where can I find an older copy?

alignedleft:
@iamutkarshtiwari I don't think we archive contributions anywhere. You would need to check the websites of various contributors to see if they archive older versions. For example, here are older copies of the official video library:

https://github.com/processing/processing-video/releases

joelmoniz:
@iamutkarshtiwari

Another quick way to simulate this is to install a contribution, and simply
change the integer value of the "version" field to a smaller integer in the
.properties file of the installed contribution.

iamutkarshtiwari:
@alignedleft @joelmoniz  thanks ! :)

Akarshit:
@alignedleft Can we make some space to show a progress bar for all the contributions when the `Update All` button is clicked as some contributions are huge and take a lot of time to download. So i think a progress bar would be better than a spinner.  
Currently the progress bar of update is shown when the contribution is selected in its respective tab after `Update All` button is clicked. Perhaps we can display it in the left most corner of the status panel.

alignedleft:
@Akarshit This is a great idea. I can't trigger any updates on my end. If it's convenient, could you post a screenshot capturing the scenario you describe above?  Then I can propose a more specific UI treatment. Thanks.

iamutkarshtiwari:
@alignedleft @Akarshit If "download bar" sounds more interesting, I can implement that as well.

mark-orion:
I want to throw my 5p in for the progress bar. As it has already been said: Downloads can take time - specially on a slow connection and I appreciate the work that is going on here!

Akarshit:
@alignedleft I was working on this for a while and came up with a couple of designs

1. Simply display the progressbar in the bottom left of the window.
[...]

PR #4229, opened 2016-01-12
progress bar in the status column while installing/removing/updating

Akarshit:
Partial fix for #4105 
This shows progressbar for all other tab except Update tab.
A few bugs still need to be fixed but the branch is useful for testing.
A screenshot of the progressbar while installing is provided below
screenshot from 2016-01-12 17 43 09


alignedleft:
@Akarshit, in the Status column, it looks like the progress bar is going right to left. Either that, or the progress bar is white with a gray background. Either way, it doesn't seem to match the progress bar below right (which is gray, and moves left to right). Can you confirm?

Akarshit:
@alignedleft The progressbar in the status column goes from left to right, with a white color and gray bakground, and the one in the description panel goes from left to right, with a gray color and background same as that of the panel.
I made it this way to be approximately consistent with the mockup below 
This behaviour if you are finding inconsistent could be easily changed. Would you like the progress bars to be as in the status column or as in the description pane?

alignedleft:
@Akarshit Ah, I see! I find this confusing. In the "Status" column in the mockup above, light blue indicates progress, where gray is "not yet complete". So, as the installation progresses, the blue area increases and the gray area decreases. Correct?

I would recommend changing this, so the "progress" area is gray (instead of blue) and the "not yet complete" area is just the background color of the row — white, if the row is not selected, or blue, if the row is selected.

What do you think? I can mock that up if needed.

Akarshit:
@alignedleft I have changed the progressbar to look like the figures below
screenshot from 2016-01-20 18 17 17
screenshot from 2016-01-20 18 16 42
I think this is what you intended ?

benfry:
Going with a simpler solution from here https://github.com/processing/processing/pull/4154 that allows us to see that something is happening, but doesn't duplicate the progress indicator from the bottom panel.

Issue #4237, opened 2016-01-21
CM-dealy when CM is opened even though list is downloaded beforehand


Akarshit:
Even after the number of updates available is displayed in the editor(that means that the list has been downloaded) there is still delay when opening CM

PR #4238, opened 2016-01-21
Moved the stuff around to make things work faster

Akarshit:
Fix for #4237 
1. Instead of downloading the contribution file 2 times, it is done only a single time now
2. There is no loader displayed when the CM is opened after the list has been downloaded. The list is populated instantaneously.

Made some of the variables and function static and moved the list downloading function to the ManagerFrame class as they are singular for the Manager.

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -343,8 +343,10 @@ public class Base {
     // check for updates
     new UpdateCheck(this);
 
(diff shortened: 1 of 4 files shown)


benfry:
This code has several style/practice issues. Making so many things `static` is not a solution, it's a hack. A single list download is a good idea, but this introduces a lot of bad practice into the code.

PR #4241, opened 2016-01-23
Added scroll funtionality to the library menu

Akarshit:
This is the fix for the issue(which i was not able to find) in which the list of libraries were too long to fit into the screen in the `Import Libraries...` menuItem.
The default no. of items has been set to 20.
The menu looks like below now(Sorry for the poor quality of image)
img_20160123_162511546_hdr
Functionalities - 
- The menu is scrollable now
- The menu opens to the last scrolled location
- The menu can be scrolled using mouse wheel
  Further features - 
- Open the menu to the full height of the screen(for which the scrollCountForScreen has already been implemented)
- Fix the first item in the menu ( Add Libraries...)
  Though there is a little bug in this, the down arrow shift upward by 1 when the menu is scrolled down by more than 2 units. The screenshot of it is posted below
  img_20160123_162539315_hdr

### DIFF
--- a/app/src/processing/app/Mode.java
+++ b/app/src/processing/app/Mode.java
@@ -484,6 +484,7 @@ public abstract class Mode {
   public void rebuildImportMenu() {  //JMenu importMenu) {
     if (importMenu == null) {
       importMenu = new JMenu(Language.text("menu.library"));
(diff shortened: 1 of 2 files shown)


benfry:
Closing since this since it's a very complex solution (3000+ lines of code!) for something that we wouldn't recommend anyway (installing more than 20 libraries at a time). If folks need this, they should really be using a full IDE.

PR #4243, opened 2016-01-24
Message pops up instead of blank window when sketchbook is empty

PARAG00991:
Presently when the File->sketchbook button is clicked , window that appears is(if the sketchbook is empty)-
screenshot from 2016-01-24 19 37 14

Instead of showing this blank window , I thought that it would be better if it shows some message like shown below-
screenshot from 2016-01-24 19 29 26
I've sent a pull request for the changes required. Please tell me if something other than this is required.


benfry:
If anything, we should just do grayed-out text that says "Empty sketchbook". A pop-up window should be reserved for something on which the user actually needs to be interrupted or needs to take action. This would hit every single Processing user on their first use.

Issue #4254, opened 2016-01-28
Implement click functionality on the whole width of selection in sketchbook


Akarshit:
For the items in the sketchbook, they only open when the name is double clicked. Instead of this they should open whenever the user double clicks anywhere on the selection(the horizontal highlighted area)
For example 
![screenshot from 2016-01-28 17 59 51](https://cloud.githubusercontent.com/assets/7762605/12644581/cbd43408-c5e9-11e5-96cb-99a02dbdbbc9.png)
In the screenshot above only if the area in which the text is written is double clicked then the sketch opens. Clicking outside the area(the area of the test which is highlighted with a dashed boundary) will not do anything.

benfry:
That's non-standard UI behavior. It's even weirder on OS X where trees like this aren't single-touch like they are on Windows and Linux.

PR #4255, opened 2016-01-28
Click working on the whole row in sketchbook

Akarshit:
Fix for #4254
* Removed the highlighting box
* Single/Double click working on the whole row
* Toggle working on the whole row


benfry:
Mmm.. this is non-standard UI. Not something we want to do.

Issue #4272, opened 2016-02-06
'Background Color when Presenting' not visible on Preferences window


PARAG00991:
When we select the background color used when using present from color selector in preferences window, it doesn't get displayed in the color selector icon. This could be illustrated from the following snapshots:
![screenshot from 2016-02-06 19 03 20](https://cloud.githubusercontent.com/assets/8044561/12866978/1da4981c-cd07-11e5-972b-397e7e913c8e.png)
![screenshot from 2016-02-06 19 03 56](https://cloud.githubusercontent.com/assets/8044561/12866979/2015f55a-cd07-11e5-8c34-6fac989eed44.png)

benfry:
Fixed in https://github.com/processing/processing/pull/4278

PR #4273, opened 2016-02-06
background color when presenting is now visible on preferences window

PARAG00991:
Fix for #4272 
Now the background color used when using Present is getting visible on the Preferences window as shown in the snapshot below:
screenshot from 2016-02-06 19 29 47


benfry:
Fix already incorporated from here https://github.com/processing/processing/pull/4278

PARAG00991:
@benfry, but I sent PR earlier than the fix in #4278 .

PR #4278, opened 2016-02-09
 'Background Color when Presenting' not visible on Preferences window

iamutkarshtiwari:
This patch fixes the issue -> https://github.com/processing/processing/issues/4272

***Color selector icon*** in the ***Preferences*** window will now be able to update it's color depending upon user's color selection in Color Selector panel.


benfry:
Good catch; thanks for the fix.

Issue #3921, opened 2015-09-30
implement 2x versions of the icons for the debugger window/variable inspector


benfry:
Didn't make the cut for 3.0.

benfry:
Resolved in 4.0 beta 9, where 1x and 2x versions of the icons are generated from SVG files.

PR #4291, opened 2016-02-10
implement 2x version of the icon of variable inspector

PARAG00991:
Fix for #3921


benfry:
That's not a real implementation—we need to use the 1x version on 1x displays, the 2x version on 2x displays. See elsewhere in the code that these icons are implemented.

PR #4297, opened 2016-02-14
IO: Use GPIO numbers in examples

gohai:
Even the Raspberry Pi Foundation used GPIO numbers over (congruous) physical pin numbers:
https://www.raspberrypi.org/learning/introduction-to-processing/worksheet-2/

Switch our examples as well, so that it they're more clear and hardware-agnostic.


benfry:
Works for me; thanks!

PR #4298, opened 2016-02-14
IO: PR #4297, plus Fritzing drawings

gohai:
They are about 190kb each. I think documenting the hardware-aspect in such a visual way, close to the sketch (which then also includes the link to the datasheet) is a nice touch, but I am biased :)


benfry:
I think that's fine. For future use, I recommend using `optipng -o7` to make 'em as small as possible. But it doesn't make sense to do that now—we'd save 20% on the current size, while adding 80% of the image size to the repo (since the old files will still be part of the history).

PR #4299, opened 2016-02-14
implement 2x version of the icons for debugger window/ variable inspector

PARAG00991:
Fix for #3921


benfry:
Have you tested this on a machine with a retina display?

PARAG00991:
No, as I don't have any machine with a retina display. But I think it would work fine.

benfry:
Closing since this doesn't cover the base issue of new icons and proper device handling

Issue #4308, opened 2016-02-17
ant deb no longer works for non-arm based processors.


kfeuz:
Commit 8f30d42672734466ef1d8f4a3814cd36c9cec454 adds conditions that are always true.  i.e. "linux-arm32" always equals "linux-arm32".  I am guessing that one of those is supposed to be a property/variable instead.  Something along the lines of `<equals arg1="${linux-arm32}" arg2="linux-arm32" />` .  A pull request will follow shortly.

benfry:
Incorporated for 3.0.3 with https://github.com/processing/processing/pull/4309.

PR #4309, opened 2016-02-17
Fixes non-arm linux deb build process

kfeuz:
As discussed in Issue #4308 adding arm support for .deb broke the .deb build process for non-arm devices.  I've tested it and this commit allows .deb to be built on non-arm devices again.  I have not tested to verify if .debs can still be built on the raspberry pi.  Perhaps @gohai can verify this?


gohai:
Sorry for breaking this, this is obviously correct. Thanks!

PR #4310, opened 2016-02-17
Fix a few undo-related bugs

### LINKED ISSUE #4280: Undo misbehaving
to undo replacing text one needs to undo at least twice, as in the case of copy pasting undoing once only deletes the text pasted.

### LINKED ISSUE #4302: Undo is a little "jerky" in insert mode
Steps to reproduce the error:

1. Enter some text
2. Make the editor in Insert mode (by hitting the `insert` key)
3. Type to overwrite text entered in `1.`
4. Undo

PS: I think I've identified the root of the issue. I should hopefully have a fix in soon.

### LINKED ISSUE #4303: "Replace" and "Replace All" do not undo in a single step
When using the `Replace` and `Replace All` options in the `Find and Replace` dialog, and then subsequently undoing these changes, the `Undo` does not take place in a single step.

This issue, #4280 and #4302 are along the same lines, I think.

joelmoniz:
This PR attempts to resolve #4280, fix #4302 and close #4303.

#4303 was mostly an issue with not considering `Replace All` as a compound edit. I have left `Replace`s to be individually undo-able, since technically each `Replace` i a distinct operation (unlike `Replace All`s, which all happen at once).

I had to modify the `JEditTextArea` a little for #4280, because the replacement happens from within there, and there's no `Editor` object accessible (and hence, the `Editor`'s compound edit methods can't be called). The `SyntaxDocument` has compound edit methods, but they were left blank, and never overridden, so I now have.

#4302 was a little tougher to crack, but primarily boiled down to each insert and deletion being considered a separate compound edit, with replacements (i.e., entries in `Insert` mode) being modeled as a delete+insert, which kept triggering `endCompoundEdit()` and starting a new one. Another issue was that a compound edit ends whenever the text in the Editor is the same before and after a key press, which could happen when the user replace a line like `int k = 4;` with `int x = 5;' (several character overlap here, and the set of characters between each overlapping "block" would be undone at a go, and a block could be as small as a single character).

@benfry There are changes in here to a few of the methods, and how they are called, but I have tried to keep things as neat as possible. Could you please let me know if these changes are ok, if any modification to how changes have been incorporated need to be made, or if further changes are required? Thanks!

### DIFF
--- a/app/src/processing/app/syntax/JEditTextArea.java
+++ b/app/src/processing/app/syntax/JEditTextArea.java
@@ -1429,7 +1429,6 @@ public class JEditTextArea extends JComponent
     }
   }
 
-
   /**
    * Replaces the selection with the specified text.
    * @param selectedText The replacement text for the selection
--- a/app/src/processing/app/syntax/JEditTextArea.java
+++ b/app/src/processing/app/syntax/JEditTextArea.java
@@ -1597,11 +1597,9 @@ public class JEditTextArea extends JComponent
     int caretLineEnd = getLineStopOffset(getCaretLine());
     if(caretLineEnd - caret <= str.length())
     {
-      setSelectedText(str);
+      setSelectedText(str, false);
       return;
     }
-
-    document.beginCompoundEdit();
 
     try
     {
@@ -1611,10 +1609,6 @@ public class JEditTextArea extends JComponent
     catch(BadLocationException bl)
     {
       bl.printStackTrace();
-    }
-    finally
-    {
-      document.endCompoundEdit();
     }
   }
 
--- a/app/src/processing/app/syntax/JEditTextArea.java
+++ b/app/src/processing/app/syntax/JEditTextArea.java
@@ -1429,15 +1429,30 @@ public class JEditTextArea extends JComponent
     }
   }
 
+
   /**
    * Replaces the selection with the specified text.
    * @param selectedText The replacement text for the selection
    */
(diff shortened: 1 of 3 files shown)


benfry:
@JakubValtar Can you review?

JakubValtar:
Reviewing right now.

JakubValtar:
#4280 is not fixed

Other two are fixed, but I find Undo still terribly broken when more tabs are involved.

When Undo modifies more tabs, the code in Editor is updated, but they are not marked as modified (orange strip), and when I save without visiting the tabs, changes made by undo are not saved (something to do with SketchCode having both `Document` and `String program`). This is quite dangerous and confusing as hell.

I was also fighting with this with Rename and I temporarily fixed it in some places (JavaBuild, PreprocessingService) by re-reading documents from all tabs instead of using SketchCode's `program`.

Also when Undo affects more tabs, you can undo only from one of them, others don't have any Undo available.

I think Undo needs to be fixed before 3.1. The situation with rewriting text from code is already quite convoluted and maybe it would be better to clean it up instead of adding more complexity by merging this. If you want I can look at it tomorrow.

benfry:
Hm, so that would delay 3.1 for a week. Are things any worse with Undo than they have been in the past? Or is it just a matter of it being a high priority?

JakubValtar:
@benfry Let's merge this and I'll work on Undo and look at #4280 when it's merged

benfry:
Ok, done.

PR #4313, opened 2016-02-18
Uncomment identation fix #4249

### LINKED ISSUE #4249: Comment/uncomment removes indenting
Version 3.0.1, running on Mac Mini El Capitan.
I have not tried to reproduce this on any other platform.

I have the processing preference editor.tabs.size set to 4.

When I comment/uncomment, each time one space is removed from the indentation of curly brace contents until such is eventually aligned with the braces. It does not proceed further. In the case of deeply nested code, the issue gets a bit more interesting.

I can submit my preferences and theme files as well, because there may be some ancillary interaction responsible for this issue.

DarkPrince304:
Hello,

With this I think the issue #4249 is fixed and uncommenting the lines wont reduce the identation, and it will remain the way it was before uncommenting. :)

Would like to have some review on this!

### DIFF
--- a/app/src/processing/app/ui/Editor.java
+++ b/app/src/processing/app/ui/Editor.java
@@ -2052,11 +2052,7 @@ public abstract class Editor extends JFrame implements RunnerListener {
         continue; //ignore blank lines
       if (commented) {
         // remove a comment
-        if (lineText.trim().startsWith(prefix + " ")) {
-          textarea.select(location, location + prefixLen + 1);
-        } else {
-          textarea.select(location, location + prefixLen);
-        }
+        textarea.select(location, location + prefixLen);
         textarea.setSelectedText("");
       } else {
         // add a comment


JakubValtar:
@DarkPrince304 Thanks!

@benfry Let's merge this for 3.1.1

benfry:
@JakubValtar also see note I just added [here](https://github.com/processing/processing/issues/4249#issuecomment-217892419) as you're looking into this further.

JakubValtar:
@benfry There was no space being added. However, if there was a space after the slashes, it got deleted (fixed by this PR).

I'm personally used to slashes being added to the left border (zero line offset), now they are added after indent and it makes the code messy. Do you think it would be better idea?

Also the keyboard shortcut doesn't work on my keyboard, I will have to fix it too.

benfry:
I was used to the slashes going at the beginning of the line, but relented because we got so many requests to change it. 

When the new version was contributed, an additional space was added (since that's now the style of many editors... perhaps Sublime in particular?) I removed that with the later commit that was referenced.

JakubValtar:
Okay, now it should be fully functional, I'm just going to fix the keyboard shortcut.

JakubValtar:
Shortcut follow up: https://github.com/processing/processing/pull/4457

Issue #4322, opened 2016-02-22
3.0.2 IDE  - code disappears


zdjurisic:
I am not sure I can reproduce the problem but this is the second time it happened in a week, so here it goes:

While editing code in 3.0.2 IDE, the code window goes completely white, but as I press the up/down arrow to move cursor across lines, the lines of code appear. Here is the console contents that appeared at the moment the code window went white:

Error repainting line range {80,121}:
java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
	at java.util.ArrayList$Itr.next(ArrayList.java:851)
	at processing.mode.java.JavaEditor.findError(JavaEditor.java:2532)
	at processing.mode.java.pdex.JavaTextAreaPainter.paintErrorLine(JavaTextAreaPainter.java:359)
	at processing.mode.java.pdex.JavaTextAreaPainter.paintLine(JavaTextAreaPainter.java:217)
	at processing.app.syntax.TextAreaPainter.paint(TextAreaPainter.java:490)
	at processing.mode.java.pdex.JavaTextAreaPainter.paint(JavaTextAreaPainter.java:587)
	at javax.swing.JComponent.paintChildren(JComponent.java:889)
	at javax.swing.JComponent.paint(JComponent.java:1065)
	at javax.swing.JComponent.paintChildren(JComponent.java:889)
	at javax.swing.JComponent.paint(JComponent.java:1065)
	at javax.swing.JComponent.paintToOffscreen(JComponent.java:5210)
	at javax.swing.RepaintManager$PaintManager.paintDoubleBuffered(RepaintManager.java:1579)
	at javax.swing.RepaintManager$PaintManager.paint(RepaintManager.java:1502)
	at javax.swing.RepaintManager.paint(RepaintManager.java:1272)
	at javax.swing.JComponent._paintImmediately(JComponent.java:5158)
	at javax.swing.JComponent.paintImmediately(JComponent.java:4969)
	at javax.swing.RepaintManager$4.run(RepaintManager.java:831)
	at javax.swing.RepaintManager$4.run(RepaintManager.java:814)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:76)
	at javax.swing.RepaintManager.paintDirtyRegions(RepaintManager.java:814)
[...]

PR #4324, opened 2016-02-23
Fix concurrency issues with errorPoints

qiubit:
This pull request addresses #4322

It is caused by using enhanced for-loop when iterating through `synchronizedList` (details here https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#synchronizedList(java.util.List) and here http://stackoverflow.com/questions/6596673/java-concurrentmodificationexception-while-iterating-over-list) and one unfortunate line in code that makes `synchronizedList` normal `ArrayList` which breaks thread-safety.

The problem is resolved by wrapping each enhanced for-loop with `synchronized` block (and also fixing one `errorPoints` redeclaration). After doing that, I'm unable to reproduce the bug by code change described here https://github.com/qiubit/processing/commit/62138e78dc75506b690401a027271e7b60907bb8 (the scheduler always schedules list modification after iteration).

### DIFF
--- a/java/src/processing/mode/java/JavaEditor.java
+++ b/java/src/processing/mode/java/JavaEditor.java
@@ -2529,13 +2529,15 @@ public class JavaEditor extends Editor {
   public LineMarker findError(int line) {
     List<LineMarker> errorPoints = getErrorPoints();
     JavaTextArea textArea = getJavaTextArea();
-    for (LineMarker emarker : errorPoints) {
-      Problem p = emarker.getProblem();
-      int pStartLine = p.getLineNumber();
-      int pEndOffset = textArea.getLineStartOffset(pStartLine) + p.getPDELineStopOffset() + 1;
-      int pEndLine = textArea.getLineOfOffset(pEndOffset);
-      if (line >= pStartLine && line <= pEndLine) {
-        return emarker;
+    synchronized (errorPoints) {
+      for (LineMarker emarker : errorPoints) {
+        Problem p = emarker.getProblem();
+        int pStartLine = p.getLineNumber();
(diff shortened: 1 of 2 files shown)


JakubValtar:
Thanks a lot for looking into this.

It makes the problem go away, but synchronization is not the best solution. You end up with more complicated code than needed. (Also, the dispatch thread can be blocked, waiting for synchronization - this removes the benefit of doing things concurrently, doesn't it?)

The errorPoints arraylist should be always accesed from the dispatch thread (EDT), so there is no need for synchronization.

However, there is one SwingWorker which modifies the collection in a background thread. If you instead get rid of the SwingWorker (which is not really needed, the code is not that heavy) and run the code right there, there will be no concurrent modification.

Would you like to try this in a new pull request?

qiubit:
Sure, will try to look into this today. Thanks for the suggestions.

qiubit:
When looking at the code, there is a comment:

> Also, update is in the UI thread via SwingWorker to prevent concurrency issues. [Manindra]

Are you sure it is safe to move this code entirely to EDT anyway? When grepping all `errorPoints` operations, it seems that everything is being run from there (apart from this SwingWorker code), but still this comment made me a little bit worried that there can be a little subtlety hidden somewhere, that I can't see....

JakubValtar:
The updateErrorPoints function always runs on EDT (also known as UI thread)
so it's safe. Done is called on EDT after the background operation is
complete. The code which runs in background thread is fast so we don't need
SwingWorker and it can be moved to EDT altogether.

(Also you don't need to call recalculateMarkerPositions in there, it is
called in repaint.)

JakubValtar:
Effort moved to #4325.

PR #4325, opened 2016-02-23
Move updateErrorPoints background work to UI thread

qiubit:
This pull request addresses #4322 and provides alternative solution of this problem than #4324, as suggested by @JakubValtar.

Instead of using Java synchronization mechanisms, which would sometimes cause UI thread to block until background thread in `updateErrorPoints` finishes execution, the work from background thread is now entirely done on the UI thread, which entirely eliminates synchronization issues (that's why all `synchronized` blocks were removed from `updateErrorPoints`).

Also, `errorPoints` redeclaration is fixed in order to be consistent with the initial declaration (which allows future `errorPoints` references from different threads to be safe - although maybe using `synchronizedList` now is unnecessary, given that all `errorPoints` references are from UI thread now?)

### DIFF
--- a/java/src/processing/mode/java/MarkerColumn.java
+++ b/java/src/processing/mode/java/MarkerColumn.java
@@ -32,7 +32,6 @@ import java.util.Collections;
 import java.util.List;
 
 import javax.swing.JPanel;
-import javax.swing.SwingWorker;
 import javax.swing.text.BadLocationException;
 
 import processing.app.Mode;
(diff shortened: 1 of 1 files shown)


JakubValtar:
Thanks, great job!

If you still want to change it, I think we don't need synchronized list, regular ArrayList will do (it's accessed only from one thread). Also it should be enough to just clear the list instead of creating a new one every time update is called. Otherwise I think we are ready to merge it.

JakubValtar:
@benfry Can you please merge this? Fixes #4322, which is quite nasty.

JakubValtar:
Thanks you @qiubit for fixing this, it's now merged and will be part of the next release.

PR #4343, opened 2016-02-29
Added Ukrainian localization

rzats:
This is based on the v3.0.2 PDE.properties, so all the current strings should be translated.

_Note:_ Processing doesn't appear to have any obvious issues related to displaying cyrillic script.


benfry:
Awesome; thanks!

Issue #2886, opened 2014-10-06
Non-western language hint text for toolbar buttons not correctly displayed


codeanticode:
When switching to Chinese or Korean, the hints that appear next tot he play and stop buttons don't render correctly (Japanese is fine because those strings are not translated):

![bug](https://cloud.githubusercontent.com/assets/62246/4525416/81e3be3c-4d53-11e4-9856-f64a286e033a.png)

crazymaster:
Footer doesn't also render correctly.

![screenshot from 2015-10-05 13 30 55](https://cloud.githubusercontent.com/assets/1528093/10273072/bb88b9f4-6b65-11e5-964c-b7b6c5775517.png)

No More Tofu!!!

satoshiokita:
@crazymaster , I have a question. What your environemnt font setting?
I can be reproduced codeanticode's header. but I cant reproduced your screenshot .

crazymaster:
It is "Takao Gothic" with Japanese on Ubuntu 15.10.

P.S. Those strings are not translated in Japanese.
Search 'FIXME' in https://github.com/processing/processing/blob/master/build/shared/lib/languages/PDE_ja.properties.

satoshiokita:
Thanks, I could find 'FIXME' comments in PDE_ja.properties.
I will send a PullRequest in a few days.

![001](https://cloud.githubusercontent.com/assets/16870334/13486237/8aa0ae60-e152-11e5-9cb3-0880311f4b5f.png)

benfry:
Should be fixed in 3.1 with https://github.com/processing/processing/pull/4348

PR #4348, opened 2016-03-05
fix:load fallback font for CJK(The Chinese, Japanese and Korean).

satoshiokita:
This is a pull request for #2886.
If a language is the CJK(The Chinese, Japanese and Korean), this patch load a system font.
the header and footer could draw correct font.

screenshot from 2016-03-05 13-26-46


crazymaster:
:+1:

benfry:
Thanks for looking into it.

PR #4351, opened 2016-03-07
PApplet: Adds helper methods to reduce code repitition

irisSchaffer:
Adds `setInternalRandom` and `setPerlinRandom` methods and adjusts indentation of comments for a `printArray`.


benfry:
Thanks; though the code repetition is one line, and this adds 10 lines of source (+29, -19) to avoid it, so better to stick with what's there already.

PR #4373, opened 2016-03-24
Added getVertexData() and getMatrix() to PShape

codetorex:
Hello,

In PShape class there is methods for change matrix of the shape; like translate, rotate etc. But there is no function to retrieve this matrix. Which is important for being able to copy shapes manually and create a shape from previous shape by applying some transformations. (There is static copyMatrix function but it also not public and marked as unapproved and only used by createShape that uses another shape which only does automatic copying)

Also another missing functionality was accessing raw float[] per vertex data. It seems possible to add new vertices by giving raw float array (shown in reference but I couldn't find in code, is it removed?) but is not possible retrieve it. Currently it is required to access this data by calling these functions getVertex, getTextureU getTextureV, getNormal etc. Still there is some other data (like per vertex material data) that is not accessible might be lost. It would be great if it is possible access it directly for the purpose of copying with applying different transforms.

If this changes are approved, copyGeometry function could use this getVertexData and copy the geometry by directly copying the float array.

Thanks!

### DIFF
--- a/core/src/processing/core/PShape.java
+++ b/core/src/processing/core/PShape.java
@@ -2158,6 +2158,15 @@ public class PShape implements PConstants {
     return getVertex(index, null);
   }
 
+  public float[] getVertexData(int index)
+  {
+    return vertices[index];
+  }
+
+  public PMatrix getMatrix()
+  {
+    return matrix;
+  }
 
   /**
    * @param vec PVector to assign the data to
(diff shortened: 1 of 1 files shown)


diwi:
still important.

tolkanabroski:
Does this work with Processing ?

micycle1:
@benfry, @jeremydouglass Can this be merged?

JAY-M-ERICSSON:
this is still very important. Setting also a matrix would be needed.  Could this be merged ?

jeremydouglass:
Hi @micycle1 -- sorry for delayed reply. I'm a community member, I don't actually have control over merges to this repository.

My *guess* would be that new features on PShape like this might be proposed as an issue / PR on Processing4 https://github.com/processing/processing4/ and ben and sam would look at it there -- but that's just a guess

PR #4379, opened 2016-03-31
Add two Raspberry Pi related fixes to JOGL

gohai:
This is JOGL 2.3.2 with two fixes I wrote on top:
https://github.com/sgothel/jogl/pull/96
https://github.com/sgothel/jogl/pull/97

Makes the output window centered when using the default closed-source driver on the Raspberry Pi, and adds support for using the new OpenGL driver with the files of the old one still in place. Tested on the Raspbian and OS X.

@benfry @codeanticode 
Let me know if we could carry this slightly modified JOGL file till their next upstream release. Should be functionally equivalent on all platforms besides the Pi.


codeanticode:
If nothing else was changed in JOGL, then I would ahead.

gohai:
I also tested this change on Windows 10.

benfry:
Hm, there's a cost to a big binary like this—how soon is the new JOGL due?

gohai:
There's only been [10 commits](https://github.com/sgothel/jogl/commits/master) since 2.3.2, and the next release is supposed to be 2.**4**.0 - believe this won't come anytime soon :(

benfry:
Ouch, ok. Fingers crossed...

PR #4387, opened 2016-04-04
Update JDT to 4.5.2

JakubValtar:
- updated jars
- removed stuff which does not exist from classpaths


benfry:
This is making me a little nervous since jdi.jar and jdi-model.jar had to be extracted and referenced separately in the past. But I'll trust you that this isn't breaking anything.

PR #4393, opened 2016-04-06
looping & redraw to private, draw

onlylemi:
* define `looping` & `redraw` to private, because there are `loop()` and `noLoop()` function to set it. It will ensure the encapsulation of class. 
* And in `draw()` function, `finished` should be false. Because class extends base and override `draw()` function, it will execute `super.draw()`. If `finshed` is true, it will execute once.

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -709,10 +709,10 @@ public class PApplet implements PConstants {
    */
   public float frameRate = 10;
 
-  protected boolean looping = true;
+  private boolean looping = true;
 
   /** flag set to true when a redraw is asked for by the user */
-  protected boolean redraw = true;
+  private boolean redraw = true;
 
   /**
    * ( begin auto-generated from frameCount.xml )
@@ -1803,7 +1803,7 @@ public class PApplet implements PConstants {
   public void draw() {
     // if no draw method, then shut things down
     //System.out.println("no draw method, goodbye");
-    finished = true;
+    //finished = true;
   }
 
 


GKFX:
I don't understand why you've removed the line from `draw()`. If the user provides a draw function, it means that `finished = true;` doesn't get run because `PApplet.draw` doesn't get run. If they don't provide a draw function, finished is set to true to stop the sketch trying to use draw repeatedly and consuming a lot of processing power.

onlylemi:
hi @GKFX ,

I understand you, but I develope processing by `IDEA`. I mush extends `PApplet` and override `setup()` `draw()` function. If do these, override function means execute `super.draw()`. Then `draw()` will execute once.

Look here,

```java
package com.onlylemi.processing.PTest;

import processing.core.PApplet;

public class PTest extends PApplet {

    @Override
    public void settings() {
        super.settings();
        size(720, 480);
    }

    @Override
    public void setup() {
        super.setup();
        
        ......
    }

    @Override
    public void draw() {
        super.draw();

        ......
    }

    public static void main(String[] args) {
        PApplet.main(new String[]{new PTest().getClass().getName()});
    }
}
```

GKFX:
Then delete the line `super.draw()`. You can't submit a pull request that breaks a feature of Processing just to make it work with your IDE. If you're struggling with Processing, you can ask on the [forum](https://forum.processing.org/two/) or StackOverflow.

benfry:
This is a misunderstanding of the fundamentals for how the software works (removing `finished` from `draw()`) and would break existing code (changing access to `redraw` and `looping`) in a fairly drastic way for fairly limited benefit (greater purity for encapsulation).

Issue #4391, opened 2016-04-04
cast long to int failing for big numbers


MNAF:
`  long a = 1662664642;
  int processingInt = int(a);
  int javaInt = new Long(a).intValue();
  println(a,processingInt, javaInt);`

gives this result:

`1662664642 1662664704 1662664642`

in Processing 3.0.2

JakubValtar:
This is a bug, we are missing parsing functions which accept longs, so it gets converted to float and then parsed to int. Either we should remove the claim that we accept longs from [the reference](https://processing.org/reference/intconvert_.html), or implement long versions for all parseXXX functions.

@REAS @benfry?

REAS:
It's seems easiest to remove the claim(s) from the Reference, but is that what we want to do?

benfry:
No, we should fix it.

PR #4395, opened 2016-04-07
Cast long to int failing for big numbers #4391.

MNAF:
Methods invoked with long parameters where cast to float parameters losing precision.


GKFX:
Thanks for this! Unfortunately, a lot of return types are wrong, such as in `static public final float abs(long n)`, where you're converting a long to a float when you only wanted to find the absolute value.

benfry:
We don't use `long` and `double` in the core API, and have no plans to add them.

PR #4402, opened 2016-04-11
Fix bracket matching bugs.

### LINKED ISSUE #4397: Multi-line comments ending in '**/' cause ArrayIndexOutOfBoundsException
Small repro:

```
/*
 * Hello World
**/

void foo() {
}
```

Pasting that into a sketch gives me exceptions galore when I try to move the cursor around or select a block of text. Same error message as #1940, but I figured I'd open a new issue since that one was closed, and looks like it probably was a different root cause.

GKFX:
Fixes #4397 and  makes bracket matching work immediately after changing tabs. (Previously, you had to make an edit after changing tabs for the bracket matchers "offsets" variable to be invalidated.)
Also undoes one line from #4113:
````
if (offsets == null || offsets.get(offsets.size()-1) >= text.length())
````
If `offsets` contains an invalid offset, then there is a bug that needs fixing, because invalidate() hasn't been called when the document changed. So it doesn't make sense to just check for bad offsets and then call parse(), because it's likely that the bug will also manifest itself in other ways.

### DIFF
--- a/app/src/processing/app/syntax/Brackets.java
+++ b/app/src/processing/app/syntax/Brackets.java
@@ -51,7 +51,7 @@ public class Brackets {
       return -1;
     }
 
-    if (offsets == null || offsets.get(offsets.size()-1) >= text.length())
+    if (offsets == null)
       parse(text);
     
     // find this bracket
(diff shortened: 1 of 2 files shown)


benfry:
@JakubValtar another for you to review.

JakubValtar:
@GKFX Good catch!

@benfry Let's merge this.

benfry:
Many thanks, fellas.

PR #4406, opened 2016-04-14
Ignore memory options when exporting for ARM

gohai:
I had students export a sketch on their laptops and try it out on a Raspberry Pi, but this failed for most, since their memory settings in Preferences were at insanely high values. Perhaps for the time being, we could ignore those for ARM?

### DIFF
--- a/java/src/processing/mode/java/JavaBuild.java
+++ b/java/src/processing/mode/java/JavaBuild.java
@@ -1123,7 +1123,11 @@ public class JavaBuild {
     /// figure out run options for the VM
 
     List<String> runOptions = new ArrayList<String>();
-    if (Preferences.getBoolean("run.options.memory")) {
+
+    // set memory options, except for ARM where we're more memory-constrained compared
+    // to the machine the user might be building the sketch on
+    if (Preferences.getBoolean("run.options.memory") &&
+        !exportVariant.equals("armv6hf")) {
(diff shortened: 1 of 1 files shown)


gohai:
(Tested by exporting on Mac, btw)

benfry:
Hm... what's a better solution here? Do we repeat the memory preference when exporting for ARM, so that folks can turn it off? Without this, people have no control over memory on ARM. 

Or... is it the case that there isn't much memory control anyway, so this solution will work in the short (next 12-18 months) term?

gohai:
Another idea: we could add a little bit of shell magic to the startup script on Linux to make it so that we request _min(memoryRequested, memoryAvailable)_. There is /proc/meminfo, which is pretty easy to parse, and should be available pretty much everywhere.

But you're right, there isn't much memory to control anyway :)

benfry:
I think the benefits here outweigh the potential issues, at least for the next year or three.

festiberico:
I am coding in RPI and exporting for ARM. Precisely because I do not have a lot of memory (and I need to preload a number of images) I need to set it to 512M  before exporting. If the exporter would ignore the settings in preferences that would create a further problem. Then I would regard the exporting as broken and need to avoid it altogether :(.
If it works as it should, why break it on RPI? Users exporting from other machines should check the memory or get a warning instead of crippling the export for everyone, specially those that run the machines target of the expor.
ps. This sounded very much like a non-solution. I even finally opened an account to comment...

PR #4410, opened 2016-04-17
Fix drawing lines with tabs

### LINKED ISSUE #3975: Editing caret not working properly when files have tabs
For some reason my cursor caret is behavnig like that:

caret bug

This started to happen when i've started saving my files with sublime text 2. The same file opens normally on windows notepad and notepad++

**edit**: seens a problem with the tab character. It doesn't seem to happen on lines without it.
Also when i first type on Processing and then load the files on ST or NP++, those text editors shows that the Tab entered in processing is actually four spaces, not a Tab character.

JakubValtar:
Fixes #3975


benfry:
Thanks... I think there are a couple dupes that go with #3975 too.

JakubValtar:
Searching for issues containing "tab", I didn't find anything related.

PR #4416, opened 2016-04-19
Add automatic mipmap support to GLES2

gohai:
glGenerateMipmap is supported on GL ES 2.0 and above. Tested on Raspberry Pi w/ binary GLES2 driver.


gohai:
CC @codeanticode

PR #4426, opened 2016-04-24
Fix file path decoding

### LINKED ISSUE #4417: Processing 3.0.2 Error
Hello, I just got this error when trying to open 3.0.2 for the first time. Windows 10 64-bit.

error

JakubValtar:
URLDecoder was being used for path decoding, even though it can't handle
RFC2396 encoding. This resulted in plus characters being removed and
possibly other weirdness.

See https://docs.oracle.com/javase/8/docs/api/java/net/URL.html

> The recommended way to manage the encoding and decoding of URLs is to use URI, and to convert between these two classes using toURI() and URI.toURL(). The URLEncoder and URLDecoder classes can also be used, but only for HTML form encoding, which is not the same as the encoding scheme defined in RFC2396.

Fixes #4417

### DIFF
--- a/app/src/processing/app/Platform.java
+++ b/app/src/processing/app/Platform.java
@@ -26,6 +26,8 @@ package processing.app;
 import java.io.File;
 import java.io.FilenameFilter;
 import java.io.IOException;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


benfry:
Ok, I'll need to test this on OS X since that's why the decoding is in there in the first place.

JakubValtar:
This is how encoding should be handled according to Java documentation, so I presume it's going to work alright. In any case, URLDecoder is not the right tool for the job.

PR #4443, opened 2016-05-07
Update to JNA 4.2.0

gohai:
Initial motivation for this was to support ARM platforms on libraries that don't come with their own copy of JNA, but depend on the one that comes with Processing. (The Kinect library did that, but I wouldn't be surprised if there were more.)

JNA 4.2.0 and above should work with ARM - this is also what this PR updates to. (The Video library uses the same version since a couple of month with no complaints.)

jna-platform.jar is required for the Win32-specific functionality that the app makes use of. Note: The name of the subfolder containing jnidispatch.dll for 64-bit Windows changed - fixed this in build.xml.

This successfully starts up on Windows 10.


benfry:
I'd rather that the libraries include the version of JNA that they need, otherwise they'll just run into compatibility problems down the road. But I'd rather we be using a fresher version of JNA...

benfry:
Are you sure this works on Windows? When double-clicking processing.exe I now get errors about native support not working properly.

gohai:
@benfry I only tried `ant run` on my jna-update branch w/ Windows 10. Unfortunately I don't have my Windows desktop at hand to try more... perhaps, revert if release is imminent? (Sry sry)

benfry:
Hm, scratch that... processing.exe is working, but it seems that the new JNA just exacerbates https://github.com/processing/processing/issues/3543, so we no longer get the useful error message workaround. I'll keep fiddling.

PR #4446, opened 2016-05-08
Move PDEX listeners from JavaEditor to PDEX

### LINKED ISSUE #4281: handleCtrlClick misbehaving
double clicking a word and copying it (ctrl + c) often throws me to the definition of the word.
it's probably handleCtrlClick(evt) in JavaTextAreaPainter.java on line 86.

playing around I noticed a quick sequence of one left mouseclick -> ctrl + c does the same, but timing is essential.

What I'd expect here is: this should only happen if ctrl is held down before actually clicking.

JakubValtar:
Just a cleanup

Fixes #4281, go to declaration happens only when ctrl was pressed before the click


lpdunwell:
nice! I really had a hard time believing it was my mistake :D

PR #4448, opened 2016-05-08
Fix build

JakubValtar:
My previous pull request was using it


benfry:
Was worried about that...

benfry:
How about `PDEX.editor`?

JakubValtar:
PDEX.editor is not used right now, all inner classes have their own reference.

PR #4451, opened 2016-05-08
PDEX: Add friendlier error message for HidingEnclosingType

### LINKED ISSUE #4228: Throw friendly error message instead of “The nested type cannot hide an enclosing type”
Hi, 

A student got this error because she named the class the same as the filename of the main sketch. Would it be possible to throw a friendlier error message? Like: "A class cannot have the same name as your the name of your sketch".

````
JitterBug bug;

void setup() {
  size(480, 120);
  bug = new JitterBug(width/2, height/2, 20);
}

void draw() {
  bug.move();
  bug.display();
}

class JitterBug {
  float x;
  float y; 
  int diameter;
  float speed = 2.5;

  JitterBug(float tempX, float tempY, int tempDiameter)  
  { 
    x = tempX;
    y = tempY;
    diameter = tempDiameter;
  }

  void move() {
      x += random(-speed, speed);
      y += random(-speed, speed);
  }

  void display() {
      ellipse(x, y, diameter, diameter);
  }
}
````

http://stackoverflow.com/questions/33902511/the-nested-type-cannot-hide-an-enclosing-type-class-java-processing

### PR DESCRIPTION
**Pull request by rzats**
Fixes #4228; is a duplicate of #4337 adjusted for some API changes.

### DIFF
--- a/build/shared/lib/languages/PDE.properties
+++ b/build/shared/lib/languages/PDE.properties
@@ -370,7 +370,7 @@ editor.status.type_mismatch = Type mismatch, "%s" does not match with "%s"
 editor.status.unused_variable = The value of the local variable "%s" is not used
 editor.status.uninitialized_variable = The local variable "%s" may not have been initialized
 editor.status.no_effect_assignment = The assignment to variable "%s" has no effect
-editor.status.hiding_enclosing_type = The class "%s" cannot have the same name as the name of your sketch
+editor.status.hiding_enclosing_type = The class "%s" cannot have the same name as your sketch or its' enclosing class
 
 # Footer buttons
 editor.footer.errors = Errors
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by JakubValtar** [comment]
Hi @rzats, thanks for your pull request.

Could you please remove the changes made in `Problem` and the `args[0].equals(problem.getClassName())` in Simplifier? It is not necessary as it will always return true.

**Comment by rzats** [comment]
@JakubValtar: here's why these are necessary - https://github.com/processing/processing/pull/4337#issuecomment-190280293
Although I could remove the changes if this tweak is not required.

**Comment by JakubValtar** [comment]
However, you are not checking it for the name of the sketch... you are merely comparing `problem.iProblem.getArguments()[0]` with itself (which is the name of the problematic class).

```processing

class C {
  class C { // Error: The class C cannot have the same name as the name of your sketch
  }
}
```

If you want to compare with the sketch name, you would have to pass it into `getSimplifiedErrorMessage`. I would suggest changing the error message to something like 'The class "%s" cannot have the same name as your sketch or enclosing class', because the sketch name is currently not available in the Simplifier.

The simplifier will be changed soon to have more information about the sketch, so we will be able support these messages better.

rzats:
@JakubValtar: thanks - this makes sense :wink: Updated accordingly.

JakubValtar:
@rzats No, thank _you_ for your contribution!

@benfry Let's merge this and add TODO to check for sketch name when I'll be refactoring this.

benfry:
Great, thanks a lot folks.

Issue #2199, opened 2013-11-02
Problems with shortcuts on non-US keyboards (ctrl-/ and others)


designhofmann:
The keyboard shortcut for Comment/Uncomment ( Cmd + / ) in Processing 2.1 doesn't seem to work on Mac OSX (10.8.5). The menu bar gets highlighted when pressed, but no action is performed on the code. 

When I click manually on the menu item it works fine but not via the shortcut. All other shortcuts work. I can confirm that it worked well in version 2.0.3.

kontur:
This still seems to be an issue. Is there a way to overwrite the shortcut combinations for keyboards that don't have these key combinations available in some configuration file for the user?

For me, too, @designhofmann 's suggestion works and when I press `cmd` + `-` on my Finnish keyboard I get the functionality of `cmd` + `/`, because the keyboard positions seem to coincide - but unless I've typed on foreign keyboard layouts before, there is no way of a user knowing or discovering this.

Ghostbird:
I'm in favour of a way to edit keyboard shortcuts too (just a configuration file is fine). The Redo action is mapped to Ctrl+Y for some reason, which is a hard to grab combo. Most of my software uses Ctrl+Shift+Z, which is a very easy and logical grab for QWERTY keyboards.

linco95:
@Ghostbird ctrl + z is actually pretty standard (not saying that what you are used to is not) way of redoing.
"control-Y is often used to redo actions." (source: "ctrl-y" at wikipedia). Just fyi :)

Pablofero:
Commenting on a spanish keyboard in windows 10 doesen't work in Processing 3.3.5! `/` is inaccesible, in other programs `ctrl+ç` is used. I hope that this problem is adressed, it has been in Processing since ages!

DonnyCraft1:
It's been five years since this issue was opened, and it's still not fixed. The majority of programmers use the US or British keyboard and don't have these problems, but the rest of us still can't use many of the shortcuts... It's time to look into this issue now, so we don't have to deal with these problems anymore.

I'm from Norway, and can't use many of the shortcuts, one of them being the `Comment/Uncomment` shortcut (`ctrl + slash`) because we access slash by typing `shift + 7`. I have tried to use `ctrl + shift + 7` but it does not work.

I hope someone can look into this issue.
Anyways, I love what the open source community can do as a whole. Just look at how long processing has come!

Here's how the norwegian keyboard looks like:
[...]

PR #4457, opened 2016-05-09
Make Comment/Uncomment trigger on numpad slash

JakubValtar:


benfry:
I suppose this is a patch, but really this is https://github.com/processing/processing/issues/2199. Are you proposing this as a band-aid until we can get to that one?

JakubValtar:
It's not, they are two different keys. I tested on US layout and on Czech layout. Numpad slash is VK_DIVIDE and regular slash is VK_SLASH. And there is no way known to me in their terrible API to add second accelerator key to the menu item.

benfry:
Heh, I understand that. I'm asking if this a patch to paper over #2199 so that it works on your keyboard (and others) until we have a better way of specifying a better shortcut for non-US keyboards?

JakubValtar:
No, this has nothing to do with locale. It does not work on any numpad, because slash on numpad is considered to be a different key than slash in the letter part on every keyboard. Menu item is set up for VK_SLASH, but that does not cover numpad slash key, which is VK_DIVIDE. I belive that unless we have one listener for each key, it will not work with both keys.

JakubValtar:
Just to make things clear: this is about numpad slash. Comment/Uncomment in other apps is often mapped to Ctrl+Slash and it always works with this key. This PR adds listener for this slash in all layouts.

keyboard2

benfry:
I can't stop laughing that even though I said I understood that these were different keys, you still had to make an image for me. We miss you back here in Boston.

JakubValtar:
Looks like their API is so bad it makes people dumb. I spent an hour with it before I figured it out and look at this.

PR #4480, opened 2016-05-17
IO: Fix drawing for SPIAnalogDigital examples

gohai:


benfry:
Have you run these through [optipng](http://optipng.sourceforge.net/)? Can't recall if we've discussed that before.

gohai:
Yes, not optipng, but something similar. They went down from 200kb to about 70 IIRC.

benfry:
Great; thanks.

Issue #4496, opened 2016-05-21
JRE downloader in build process crashes with NullPointerException


jdf:
Clean build on travis:

   https://travis-ci.org/jdf/processing.py/builds/131919028

```
jre-download:
BUILD FAILED
/home/travis/build/jdf/processing.py/processing/build/build.xml:296: The following error occurred while executing this line:
/home/travis/build/jdf/processing.py/processing/build/build.xml:242: java.lang.NullPointerException
	at Downloader.download(Downloader.java:161)
	at Downloader.execute(Downloader.java:96)
	at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:291)
	at sun.reflect.GeneratedMethodAccessor6.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:483)
	at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.java:106)
	at org.apache.tools.ant.Task.perform(Task.java:348)
```
etc.

jdf:
```java
      List<String> cookies = headers.get("Set-Cookie");
      conn = (HttpURLConnection) new URL(url).openConnection();
      for (String cookie : cookies) {
```
Apparently headers.get can return null.

gohai:
I saw this also when building for the ARM. I believe we didn't get headers because Oracle was redirecting us to a Login page instead..

gohai:
Btw: Some time ago I [came across](https://www.digitalocean.com/community/tutorials/how-to-manually-install-oracle-java-on-a-debian-or-ubuntu-vps) a simple command to download from Oracle:

    wget --header "Cookie: oraclelicense=accept-securebackup-cookie" http://download.oracle.com/otn-pub/java/jdk/8u91-b14/jdk-8u91-linux-arm32-vfp-hflt.tar.gz

Seems to work (well, at this point in time anyway). Perhaps this could simplify the downloader somewhat?

jdf:
I should say, because it's not obvious, that this is a show-stopper bug for Python Mode, because our continuous-build/test system (Travis) is the only QA I've got. :)

jdf:
Thanks very much @gohai for your wget tip! I've worked around the Downloader bug in my travis configuration by downloading the appropriate JRE into build/linux and uncompressing it there. I have a green build again.

[...]

PR #4511, opened 2016-05-27
Fix Downloader

gohai:
Addresses #4496. Enough to get JRE downloads to work again.


benfry:
That's not a superfluous slash... rsync is slash-sensitive.

PR #4547, opened 2016-06-24
Fix run failure when tabs are unviewed

DominusDodo:
This affects Python mode, but the source of the bug is here. When some
tabs have not been viewed after the sketch has been opened, the
'document' of the respective SketchCode is null. This is fine, just
don't throw an error. This fix shouldn't have any negative effects.

### DIFF
--- a/app/src/processing/app/ui/Editor.java
+++ b/app/src/processing/app/ui/Editor.java
@@ -2713,7 +2713,13 @@ public abstract class Editor extends JFrame implements RunnerListener {
     //current.setProgram(editor.getText());
     for (SketchCode sc : sketch.getCode()) {
       try {
-        sc.setProgram(sc.getDocumentText());
+        try {
+          sc.setProgram(sc.getDocumentText());
+        } catch (NullPointerException e) {
+          // sc has no document because the tab has not been viewed;
+          // carry on because String 'program' has been set and no
+		  // document should mean no changes have been made
+        }
       } catch (BadLocationException e) { }
     }
 


jdf:
I don't think this is a good patch, since it papers over an error without addressing the cause.

DominusDodo:
To clarify, the `NullPointerException` is raised because in this section

```java
  private void loadSavedCode() {
    SketchCode[] code = sketch.getCode();
    for (int i=0; i<code.length; i++) {
      if (!code[i].getProgram().equals(code[i].getSavedProgram())) {
        code[i].setProgram(code[i].getSavedProgram());
        /* Wild Hack: set document to null so the text editor will refresh
           the program contents when the document tab is being clicked */
        code[i].setDocument(null);
      }
    }
    // this will update the current code
    setCode(sketch.getCurrentCode());
  }
```

the `document` is set to `null` in each `SketchCode` instance. This means that if the tab has not been viewed, it will have a document of value `null`. However, the value for the program string in the `code[i]` will still be set to the loaded value (from `code[i].getSavedProgram()`). I can't find any other possible reasons for `document` to be null but `code[i]` to have a value without prior errors, so catching and suppressing the `NullPointerException` shouldn't conceal any other errors.

jdf:
Rather than catching the NullPointerException, the code quoted above should not set the document of each tab to ```null```. Whatever the issue was that the hack addressed should be fixed.

jdf:
To clarify, the code that sets each tab to null is a bug. Catching NPE will silently obscure an infinite class of other bugs (current and future). It's not a good idea.

DominusDodo:
So perhaps we should look to instead create a boolean flag for whether the
tab has been loaded or not?

benfry:
@JakubValtar more quirkiness with Document/SketchCode/etc...

JakubValtar:
@benfry Same problem as #4555, I think I'm going to roll back my change which caused both of these and figure out something better.

PR #4557, opened 2016-07-02
removed white space

chhavi1996:


benfry:
Stylistic choice... both versions are legitimate.

Issue #2615, opened 2014-06-27
Wrong mode shows as being selected in Mode Menu pop-up when trying to change mode with unsaved changes in sketch


joelmoniz:
When there are unsaved changes in the Sketch, and the user tries to change the mode, after the dialog that prompts the user to save unsaved changes appears, the Mode Menu popup shows the mode that the user tried to change to as the current mode, not the actual current mode.

Reproducing this issue:
1. Create a new sketch with the following code (for example), don't save: 

```
void setup() {
  size(200,200);
}

void draw() {
  rect(20,20,80,90);
}
```

<ol start="2."><li>Try to change the mode, say from Java to JavaScript</li><br />

<li>The dialog prompting the user to save unsaved changes appears, and the mode now shows as JavaScript, though the user is still on Java</li></ol>

PR #4559, opened 2016-07-03
Show warning message if mode cannot be changed

tyfkda:
If you try to change the mode which is not compatible with the current mode,
the mode change is canceled silently and different mode is checked in mode menu popup.

Reproducing this issue:

  1. Install Python mode.
  2. Select Python mode.
  3. Load any sample sketch.
  4. Click mode menu popup, and select Java.
  5. No message is shown and Java is checked in popup (but keeps Python mode.) 

This change fixes the issue,
if selected mode cannot handle current source,
show warning dialog and reselect old (current) mode.

This change added text strings for English and Japanese only.

Related #2615

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -885,8 +885,9 @@ public class Base {
   /**
    * The call has already checked to make sure this sketch is not modified,
    * now change the mode.
+   * @return true if mode is changed.
    */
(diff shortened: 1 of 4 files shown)


tyfkda:
@benfry I've fixed the conflict, so would you please take a look? Thanks!

benfry:
Thanks, let's give it a try.

Issue #4555, opened 2016-07-01
NullPointerException in SketchCode.getDocumentText()


JJ-Atkinson:
When I run tweak mode all is well. But when I try to run normal after running tweak mode, it throws this error:

```
Exception in thread "Thread-58" java.lang.NullPointerException
	at processing.app.SketchCode.getDocumentText(SketchCode.java:228)
	at processing.app.ui.Editor.prepareRun(Editor.java:2716)
	at processing.mode.java.JavaEditor.prepareRun(JavaEditor.java:1781)
	at processing.mode.java.JavaEditor$34.run(JavaEditor.java:1095)
	at java.lang.Thread.run(Thread.java:745)
```
I'm running processing 3.1.1 on windows 10.

Please note this only happens in sketches with multiple files.

galsasson:
Can you please provide an example sketch that produces this error?

JJ-Atkinson:
@galsasson  Sure, here you go:

### Tab 1

```Java
void setup() {
  size(300, 300);
}

void draw() {
  otherFunc();
}
```

JJ-Atkinson:
I went poking around in the source, and found [`JavaEditor.stopTweakMode()`](https://github.com/processing/processing/blob/0cd98fcc0288dbaceb2e949108274472c42e692d/java/src/processing/mode/java/JavaEditor.java#L2703) and [`JavaEditor.setTabContents()`](https://github.com/processing/processing/blob/0cd98fcc0288dbaceb2e949108274472c42e692d/java/src/processing/mode/java/JavaEditor.java#L1681), both of which set the document to null.

```Java
  protected void stopTweakMode(List<List<Handle>> handles) {
    tweakClient.shutdown();
    getJavaTextArea().stopTweakMode();

    // remove space from the code (before and after)
    //removeSpacesFromCode();

    // check which tabs were modified
    boolean[] tweakedTabs = getTweakedTabs(handles);
    boolean modified = anythingTrue(tweakedTabs);

    if (modified) {
      // ask to keep the values
      if (Messages.showYesNoQuestion(this, Language.text("tweak_mode"),
                                     Language.text("tweak_mode.keep_changes.line1"),
                                     Language.text("tweak_mode.keep_changes.line2")) == JOptionPane.YES_OPTION) {
        for (int i = 0; i < sketch.getCodeCount(); i++) {
          if (tweakedTabs[i]) {
            sketch.getCode(i).setModified(true);

          } else {
[...]

PR #4563, opened 2016-07-05
Fix Null Pointer Exception

tyfkda:
`SketchCode` might have no `document`.
In the case, calling `getDocumentText()` raises NPE.
This might happen when `Editor#prepareRun()` is called.

Java mode works fine,
but I've encountered this exception
when I try to implement custom mode.

### DIFF
--- a/app/src/processing/app/SketchCode.java
+++ b/app/src/processing/app/SketchCode.java
@@ -225,6 +225,8 @@ public class SketchCode {
 
 
   public String getDocumentText() throws BadLocationException {
+    if (document == null)
+      return null;
     return document.getText(0, document.getLength());
   }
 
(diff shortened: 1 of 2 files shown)


tyfkda:
This exception is happening on Python mode, too.

rwelti:
Re: V 3.1.1

I get it in Java mode, just by editing .pde with external editors (Mac). 
TextEdit and vim cause this upon reload of .pde:

Exception in thread "Thread-17" java.lang.NullPointerException
	at processing.app.SketchCode.getDocumentText(SketchCode.java:228)
	at processing.app.ui.Editor.prepareRun(Editor.java:2716)
	at processing.mode.java.JavaEditor.prepareRun(JavaEditor.java:1781)
	at processing.mode.java.JavaEditor$34.run(JavaEditor.java:1095)
	at java.lang.Thread.run(Thread.java:745)

But reverting to V 3.0.1 I do **NOT** have this problem.

Something to do with newlines I think.

tyfkda:
Related #4555

benfry:
We need to figure out why this is actually happening, rather than papering over the NPE. This solution would just push the problem to later in the process. We'll use #4555 to track progress.

PR #4593, opened 2016-08-01
Set text color for InputMethod

tyfkda:
It is difficult to see the editing text,
because default text color might be similar to the background color:

ss

### DIFF
--- a/app/src/processing/app/syntax/JEditTextArea.java
+++ b/app/src/processing/app/syntax/JEditTextArea.java
@@ -80,6 +80,7 @@ public class JEditTextArea extends JComponent
   private InputMethodSupport inputMethodSupport = null;
 
   private Brackets bracketHelper = new Brackets();
+  private TextAreaDefaults defaults;
 
 
   /**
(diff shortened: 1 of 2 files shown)


benfry:
What's normal for this sort of thing in other editors? The 'before' seems like a bug, but the 'after' doesn't feel any more standard. What's the best way to handle this?

tyfkda:
There is underline, so we can recognize editing text and I think this is standard way (and I don't change anything other than text color in this PR).

Other screenshot: Notepad (Windows):
notepad

tyfkda:
Google Chrome textarea (Windows):
chrome_textarea

benfry:
Ok, so this solution is a good start, though we should also set the background highlight color as well to the light blue.

tyfkda:
I see.
I have changed to use same foreground/background color for editor.

benfry:
Thanks for this. I've now incorporated your changes using a slightly different method, but they're now in there. Please give them a check to make sure I integrated them properly.

PR #4594, opened 2016-08-01
Insert characters by InputMethod at one time

tyfkda:
I think calling `insertString` for each character is a bit wasteful,
so do it for committed characters at one time.


benfry:
Thanks, though the impact here is minimal—I think I'm more comfortable with inserting by characters and having the editor update properly than this sequence. However, I've done a little cleanup in the code, since I've not touched it in the past and it doesn't follow our conventions.

tyfkda:
I want to reduce confusion for newbies about why cannot input comment using InputMethod.

I have create another pull request to enable input method support as default only for Japanese/Korean/Chinese, would you please consider about it? Thanks!

#4598

PR #4595, opened 2016-08-01
Support input method as default

tyfkda:
It is kind to support input method as default for such environment.
And if a user is in an environment what doesn't use input method,
enabling input method is no harm, I think.


benfry:
Because we don't have a consistent maintainer, and support for this hasn't been great, we need to leave it turned off by default. Can't risk a bug that would affect the vast majority of our users.

tyfkda:
I want to reduce confusion for newbies about why cannot input comment using InputMethod.

I have created another pull request to enable input method support as default only for Japanese/Korean/Chinese, would you please consider about it? Thanks!

#4598

PR #4596, opened 2016-08-01
Prevent NPEs when tab document is null

### LINKED ISSUE #4555: NullPointerException in SketchCode.getDocumentText()

JakubValtar:
JavaMode visits all tabs on sketch load to extract breakpoints, so doc
is always set (except when running Tweak mode; it sets docs to null to
force reload). Other modes have null docs until the tabs are visited
manually.

Fixes #4555

### DIFF
--- a/app/src/processing/app/ui/Editor.java
+++ b/app/src/processing/app/ui/Editor.java
@@ -1371,10 +1371,13 @@ public abstract class Editor extends JFrame implements RunnerListener {
       redoAction.updateRedoState();
       if (sketch != null) {
         sketch.setModified(!getText().equals(sketch.getCurrentCode().getSavedProgram()));
+        // Go through all tabs; Replace All, Rename or Undo could have changed them
         for (SketchCode sc : sketch.getCode()) {
-          try {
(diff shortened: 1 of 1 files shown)


benfry:
Thanks; maybe we'll do a 3.1.3 with this.

JakubValtar:
There are two oher places where null doc can be a problem. One is main error checker loop, it uses docs to get tab contents. It will spit an exception into console, but keep running. Other place is Rename, it will also spit an exception and do nothing. I need to revise this later and make sure there are some clear rules how document works.

PR #4598, opened 2016-08-01
Support input method as default for Japanese/Korean/Chinese

tyfkda:
It is kind to support input method as default for such environment.

Newbies try to put comment on their own language here in Japan
and get confused why cannot on PDE.

### DIFF
--- a/app/src/processing/app/Language.java
+++ b/app/src/processing/app/Language.java
@@ -236,6 +236,16 @@ public class Language {
   }
 
 
+  /**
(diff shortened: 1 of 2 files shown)


benfry:
Great; this is a much better solution than #4595, thank you. I've implemented it slightly differently, but it's now in the code. 

For future PRs, please also take a look at the [Style Guidelines](https://github.com/processing/processing/wiki/Style-Guidelines) so I can merge patches without additional modifications.

PR #4599, opened 2016-08-02
Set sketch as modified when any character committed using input method

tyfkda:
Currently, when a user input any character using input method support, sketch is not marked as modified.

So no dialog popped up when the user tried to close the window or quit the application and lost the changes.


benfry:
Ouch, that's serious. Given this, I'll also incorporate a version of  https://github.com/processing/processing/pull/4594 so that we're not firing that callback incessantly.

benfry:
Merged this and multiple other edits; please give it a try.

PR #4602, opened 2016-08-04
Insert string when it is committed

tyfkda:
`inputMethodTextChanged` is called each time
when a user pressed key and keeps modifying a text on input method support.
In the time, code hasn't been modified yet,
but calling `handleInputMethodCommit` makes sketch as modified.

Code is modified when the user pressed enter key to decide the input,
and then `committedCount` becomes non zero.


benfry:
Thanks!

PR #4607, opened 2016-08-06
Fix resizing targets for async save

### LINKED ISSUE #4578: Saving a PGraphics results in black image 50% of the time
Saving PGraphics appears to be unreliable in Processing 3.1.1. However, since the chance of failure is lower than 100%, it took some time to track down the exact issue.

I have created a test sketch that creates a PGraphics 7 times and saves each created PGraphics 3 times. So a total of 21 images are saved. When I run this test sketch, results vary, but on average about 50% of the saved PGraphics are incorrect, meaning they are fully black instead of the actual image. It really is a game of chance, so which ones are incorrect differs each time. However once the initial PGraphics is created 'wrong', then all saves are black. But in the P2D renderer, even when the first save of a created PGraphics is succesful, there is still a chance that subsequent saves might fail.

Additional findings:
- The problem occurs in Processing 3.1.1 (it does not occur in Processing 2.2.1)
- It is not caused by smooth (in the test sketch smooth is not even used)
- It is not caused by too large PGraphics dimensions (the test sketch' PGraphics dimensions are very small)
- It may be related to multiple instantiations, especially when the dimensions vary (when the PGraphics dimensions of subsequent instantiations are the same, the problem does not occur)
- It is not only OpenGL-related (the problem occurs in both the JAVA2D and P2D renderers)
- The problem occurs when similar code is run from keyPressed, setup() or draw().
- Displaying the created PGraphics seems to work consistently,  but even after correct display, saving may fail.
- There is no indication of a problem, apart from the black image, so there are no NPE's, error messages or anything to indicate saving was unsuccessful. Literally the saving _WAS_ succesful (there is a readable image file), except it just doesn't show the actual image that was in the PGraphics.

System specs: Win 8 64-bit. NVIDIA GeForce GTX 770

**Example Sketch:**
```java
PGraphics pg;

void setup() {
  size(200, 200, P2D);

  for (int i=0; i<7; i++) {
    createCanvas(100 + i * 10);
    for (int j=0; j<3; j++) {
      saveCanvas("canvas" + i  + "-save" + j);
    }
  }
  exit();
}

void createCanvas(int d) {
  // create PGraphics
  pg = createGraphics(d, d, P2D);

  // draw to PGraphics
  pg.beginDraw();
  pg.background(255, 0, 0);
  pg.fill(0, 255, 0);
  pg.ellipse(pg.width/2, pg.height/2, pg.height/3, pg.height/3);
  pg.endDraw();
}

void saveCanvas(String name) {
  //save PGraphics
  pg.save(name + ".png");
}
```

JakubValtar:
Got burned by not setting pixelWidth and pixelHeight properly.

Fixes #4578

### DIFF
--- a/core/src/processing/core/PGraphics.java
+++ b/core/src/processing/core/PGraphics.java
@@ -8364,12 +8364,10 @@ public class PGraphics extends PImage implements PConstants {
           targetsCreated++;
         } else {
           target = targetPool.take();
-          if (target.width != requestedWidth ||
-              target.height != requestedHeight) {
-            target.width = requestedWidth;
-            target.height = requestedHeight;
+          if (target.pixelWidth != requestedWidth ||
+              target.pixelHeight != requestedHeight) {
             // TODO: this kills performance when saving different sizes
-            target.pixels = new int[requestedWidth * requestedHeight];
+            target = new PImage(requestedWidth, requestedHeight);
           }
         }
         target.format = format;


benfry:
Thanks!

PR #4618, opened 2016-08-16
Fix loop maximum value for X and Y axis

tyfkda:
Loop variable `j` tracks 0 to `WIDE`, but it is used for Y axis.
Also variable `i` is used for X axis, but it is limited with `HIGH`.
This will break screen if `WIDE` will be different from `HIGH`.

This change fixes the misuse,
and swaps loop varaible `i` and `j`.


benfry:
Your fix isn't quite right: `j` should be the outer loop, but it should be looping over `HIGH`. I'll leave it for you to fix since you complained last time I fixed your patch myself.

tyfkda:
Thanks for your review.
I've swapped back the loop variable.

Issue #4407, opened 2016-04-16
loadPixels throws an exception when the window is resized


lbraginsky:
This script starts and lets you resize the window but if you stretch it enough the script crashes with ArrayOutOfBoundsException: Coordinate out of bounds!.

    void setup() {
      surface.setResizable(true);
    }
 
    void draw() {
      loadPixels();
    }

josephhenrymurphy:
This is an error being thrown by PGraphicsJava2D.java:2752:

`raster.getDataElements(0, 0, pixelWidth, pixelHeight, pixels);`

Stacktrace:

```
java.lang.ArrayIndexOutOfBoundsException: Coordinate out of bounds!
	at sun.awt.image.IntegerInterleavedRaster.getDataElements(IntegerInterleavedRaster.java:264)
	at processing.awt.PGraphicsJava2D.loadPixels(PGraphicsJava2D.java:2755)
	at processing.core.PApplet.loadPixels(PApplet.java:10687)
	at sketch_160815a.draw(sketch_160815a.java:23)
	at processing.core.PApplet.handleDraw(PApplet.java:2409)
	at processing.awt.PSurfaceAWT$12.callDraw(PSurfaceAWT.java:1540)
	at processing.core.PSurfaceNone$AnimationThread.run(PSurfaceNone.java:316)
```

I believe we can just move forward after this exception, so I will submit a fix.

benfry:
That would hide the error, but doesn't fix the underlying problem. The rendering context shouldn't be getting resized midway through `draw()` by an event happening on the event thread. `loadPixels()` is the most pronounced way to reproduce the issue, since it's time-consuming, but the real problem is that the threads are interacting in a bad way (which will give rise to other problems later).

PR #4620, opened 2016-08-16
Issue 4407: catches exception from WritableRaster.getDataElements.

josephhenrymurphy:
#4407 

Catches exception thrown by `.getDataElements` call when window is rapidly resized.

issue_4407


benfry:
Not a sufficient patch; see comments here: https://github.com/processing/processing/issues/4407#issuecomment-240195175

PR #4624, opened 2016-08-18
Scroll textarea to the left when caret is outside

tyfkda:
Left gutter is added in Java mode,
and `0` doesn't mean the left limit for `x` position.
So Scrolling to the left is not started
even if caret is hidden with left gutter.

This change calculates leftmost position of textarea,
and when caret moves out of it,
change horizontal scroll position.


benfry:
This is probably code that's still in `JavaTextArea` that needs to be moved to the superclass, rather than this sort of specific fix (assuming it was working before 3.2).

tyfkda:
The behavior come from left gutter, and not related to the code refinement for textarea for version 3.2.
I remember it has happened before 3.2.

PR #4639, opened 2016-08-28
Simplify font situation to make it possible to use vanilla JRE trees

gohai:
This has only been lightly tested on OS X and ARM Linux. The createFont function could be also simplified in the future by mandating all custom files to be placed in shared/fonts.

Note: this still prints "Source Code Pro not available, resetting to monospaced" (and will change preferences.txt accordingly). This seems to be because the first `createFont("SourceCodePro-Regular.ttf")` only comes after the call to `updateAppearance()`. (The Source _Sans_ Pro fonts are instead created & registered _before_ the `updateAppearance()`).

Will be busy with real life the next week, but I can look some more into this next weekend!

### DIFF
--- a/app/src/processing/app/ui/Toolkit.java
+++ b/app/src/processing/app/ui/Toolkit.java
@@ -975,16 +975,24 @@ public class Toolkit {
    * the Preferences window, and can be used by HTMLEditorKit for WebFrame).
    */
   static private Font createFont(String filename, int size) throws IOException, FontFormatException {
+    boolean registerFont = false;
+
     // Can't use Base.getJavaHome(), because if we're not using our local JRE,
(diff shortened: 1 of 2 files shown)


benfry:
Hopefully that "resetting" message is what I ran into before, and not realizing the source of the problem, I just backed out the `registerFont()` change. Working on a fix for that next...

gohai:
Fingers crossed!

benfry:
Just checked in a change to ensure that the mono font loads. Can you try it and see if that gets rid of the "resetting" message?

gohai:
It does indeed. Thanks a ton!

benfry:
Great; that's a good sign, at least...

PR #4641, opened 2016-08-28
Cleanup createFont

gohai:
Wanted to at least remove the misleading comment that the last fallback is for Linux only, but ended up removing the other cases too (which are hopefully not needed anymore).

### DIFF
--- a/app/src/processing/app/ui/Toolkit.java
+++ b/app/src/processing/app/ui/Toolkit.java
@@ -929,7 +929,7 @@ public class Toolkit {
 
 
   /**
-   * Get a font from the JRE lib/fonts folder. Our default fonts are also
+   * Get a font from the lib/fonts folder. Our default fonts are also
    * installed there so that the monospace (and others) can be used by other
    * font listing calls (i.e. it appears in the list of monospace fonts in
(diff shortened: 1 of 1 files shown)


benfry:
Can't incorporate this until someone tests the change across all platforms: Windows, macOS, Linux, and when running from `ant run`, from Eclipse (at least on OS X), and in a build from `ant dist`.

gohai:
Will do!

(BTW: edd9ea36bb5551bce55d85350116efc23a987b96 already stopped copying the fonts into the jre font directories - so this approach better works or it's probably best to revert completely)

benfry:
Yup, saw that—just need to make sure that we're not missing anything with the extra logic here. Or do you think we're safe to just incorporate this, since if things are broken, it's gonna be a lot of rollback?

gohai:
Either way is fine, but I better test before you release even without it :)

gohai:
@benfry I tested Mac, Windows (10), and (ARM) Linux - `ant run` and `ant dist`, with a fresh `preferences.txt`, while watching the terminal output. Looked all good!

benfry:
Thanks for looking into all that; much appreciated.

Issue #4615, opened 2016-08-13
Function result is ignored


AppChecker:
Hello!

Code in the method [UrlHelper.getFileURI](https://github.com/processing/processing/blob/aa214287e33ffec29ee47bb5c518cee8301861b7/build/macosx/appbundler/src/com/oracle/appbundler/BundleDocument.java#L54):
```
        extensions = extensionsList.split(",");
        for (String extension : extensions) {
            extension.trim().toLowerCase();
        }
```
ignores the return value of the String.toLowerCase method.

Probably, is should be:

```
extensions = extensionsList.split(",");
for (int i = 0; i < extensions.length; i++) {
    extensions[i] = extensions[i].trim().toLowerCase();
}
```

This possible defect found by [static code analyzer AppChecker](http://cnpo.ru/en/solutions/appchecker.php)

PR #4645, opened 2016-09-03
Fix set extensions MAC OSX CFBundleDocument

amrufathy:
Updated extensions with the returned value from trim() and toLowerCase() methods.
Fix [#4615](https://github.com/processing/processing/issues/4615)

### DIFF
--- a/build/macosx/appbundler/src/com/oracle/appbundler/BundleDocument.java
+++ b/build/macosx/appbundler/src/com/oracle/appbundler/BundleDocument.java
@@ -51,7 +51,7 @@ public class BundleDocument {
         
         extensions = extensionsList.split(",");
         for (String extension : extensions) {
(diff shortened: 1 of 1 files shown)


benfry:
Correct fix incorporated separately.

amrufathy:
@benfry 
I appreciate your work on processing. I've looked the fix you incorporated and it basically applies the same approach in the PR. I would've preferred if you let me know beforehand so I could incorporate the fix in my PR. I was looking for a way to contribute and this behavior discourages anyone from contributing.

benfry:
Your PR was incorrect and repeated the exactly the same problem as was in the source. I incorporated the fix that was proposed in the bug report, which was correct. If you'd submitted that as a PR, I would have incorporated it. The personal insult about "behavior" is way out of bounds.

PR #4661, opened 2016-09-11
Chinese translation

guoyunhe:
Completed most strings. Since the property file bundle is not synced, here still might be some missing things.


benfry:
Thanks!

PR #4672, opened 2016-09-20
more flexible interpreter path

tpltnt:
Hi there,

I made the shebang for the python interpreter more flexible. This helps running processing on FreeBSD / PCBSD without impacting any current user.
Cheers,
tpltnt


benfry:
Ok, all for incorporating; though that code is only run on macOS.

PR #4697, opened 2016-10-09
Update spanish translation

federicobond:


benfry:
Thanks for the updates; much appreciated.

federicobond:
@benfry You can direct any translators to this script I made a year ago to find missing translations for their language. It's not without its flaws but ended up being very useful for this kind of work.

https://gist.github.com/federicobond/741d5c7df61191e13408

If you call it like this:

````
./missing.sh
````

It will show strings that are missing from the default English properties file (indented lines are those found in the .properties file but not in the codebase)

If you call it with a language code as an argument:

````
./missing.sh es
````

It will compare the keys for that language against the keys in the default language .properties file. Again, the behavior with regards to indented lines is similar: those lines appear in the corresponding language file but were removed from the default one.

benfry:
That's awesome! Actually, would you be interested in writing up a wiki page to introduce people to how to do the translations? And you could provide a link (or simply include) this as well. You've really been spearheading the translation process for us, and folks would benefit from your explanation.

federicobond:
Sure! I could expand [this one](https://github.com/processing/processing/wiki/Localization).

benfry:
Yes, please!

federicobond:
Done. I don't have more time to elaborate on it, but I am happy to respond to any feedback.

benfry:
Excellent! Thanks once again for your help.

Issue #4668, opened 2016-09-17
Add Jump to Declaration to PDE right-click menu (enhancement)


jeremydouglass:
Could the PDE right-click menu include an option for Jump to Declaration / jump to definition?

This is highly useful, and there is a keyboard shortcut for it, but no menu items or contextual menus expose the fact that this is possible. Like "Find in Reference" it might be nice if users could discover this functionality.

(enhancement)

benfry:
Added for 3.3.1.

PR #4707, opened 2016-10-19
PDEX: Add 'Jump to Declaration' to right-click menu (#4668)

rzats:
### Contains

A simple implementation for #4668 - adds a "Jump to Declaration" option to the PDEX context menu with the same functionality as ctrl-clicking on a variable (jumps to the declaration of the variable if it was declared within the sketch; otherwise shows its' usage instead).

image
### How to test

In a Processing sketch, right-click on a variable, select "Jump to Declaration" and verify that the declaration is indeed being jumped to 😄 
### Outstanding before merging
- [ ] When a variable not defined within the sketch is ctrl+clicked, the "Show Usage" window is displayed instead. Should the menu option behave in the same way?

### DIFF
--- a/build/shared/lib/languages/PDE.properties
+++ b/build/shared/lib/languages/PDE.properties
@@ -312,6 +312,7 @@ editor.header.delete.warning.title = Yeah, no.
 editor.header.delete.warning.text = You cannot delete the main tab of the only open sketch.
 
 # PopUp menu
+editor.popup.jump_to_declaration = Jump to Declaration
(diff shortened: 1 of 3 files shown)


rzats:
@JakubValtar: this has been pending for quite a while 😃 Can you please review?

benfry:
Just not a high priority and I've not had time… Now merged.

benfry:
And yes, this menu option should behave the same way.

rzats:
@benfry: thanks and understood :)

PR #4710, opened 2016-10-21
The previous translation would actually stand for "Open sketches folder", as if this…

jordirosa-p5:
… menu option opened the generic folder where all the sketches are stored by default. But what it actually does is opening the folder of the sketch we are working on. That's why I think this translation is incorrect, and the correct one would be "Mostrar carpeta del sketch".


benfry:
Thanks; I'll have to trust you on this…

Issue #4696, opened 2016-10-07
Version listed as 'null' with PeasyCam


benfry:
See https://github.com/processing/processing-docs/issues/478. It's a problem with PeasyCam's contrib file, but we shouldn't be showing 'null' in the UI.

benfry:
Fixed for 3.2.2 or 3.3.

benfry:
Actually fixed for 3.2.3 or 3.3.

PR #4712, opened 2016-10-23
CM: Fix null versions in the contribution manager (#4696)

rzats:
### Contains

A fix for #4696 that adds some checks to avoid displaying `null` in the contribution manager. Note that `prettyVersion` becomes a `"null"` **string** when the base version is invalid - this is reflected in the checks.

### How to test

Check that PeasyCam, a library that currently has an invalid contribution file, has no `null` versions displayed in the contribution manager:

image

### DIFF
--- a/app/src/processing/app/contrib/DetailPanel.java
+++ b/app/src/processing/app/contrib/DetailPanel.java
@@ -461,7 +461,7 @@ class DetailPanel extends JPanel {
     desc.append("</font> ");
 
     String version = contrib.getPrettyVersion();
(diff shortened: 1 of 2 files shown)


benfry:
Thanks… I've removed the check for `"null"` (the `String` version) since if that's happening, we should fix the source of the bug, not hide it like this. But with that check removed, PeasyCam looks correct so I think we're set.

benfry:
Found the problem with the "null" entries, so that's now fixed too. Thanks again for your help.

Issue #4732, opened 2016-11-06
proper error handling when downloading contribs listing


benfry:
As of 3.2.2, a 500 error (or timeout, or whatever) will just spew errors into the console. If this is running in the background, then we can avoid it or just put something semi-useful in the console. 
If it happens when loading the manager, however, we need to put useful errors out there. 
```
_ crash when the contribs script is down
java.io.IOException: Server returned HTTP response code: 500 for URL: http://download.processing.org/contribs
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at sun.net.www.protocol.http.HttpURLConnection$10.run(HttpURLConnection.java:1926)
	at sun.net.www.protocol.http.HttpURLConnection$10.run(HttpURLConnection.java:1921)
	at java.security.AccessController.doPrivileged(Native Method)
	at sun.net.www.protocol.http.HttpURLConnection.getChainedException(HttpURLConnection.java:1920)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1490)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1474)
	at processing.app.contrib.ContributionManager.download(ContributionManager.java:92)
	at processing.app.contrib.ContributionListing$2.run(ContributionListing.java:424)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.io.IOException: Server returned HTTP response code: 500 for URL: http://download.processing.org/contribs
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1876)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1474)
	at sun.net.www.protocol.http.HttpURLConnection.getHeaderField(HttpURLConnection.java:2982)
	at java.net.URLConnection.getHeaderFieldLong(URLConnection.java:629)
	at java.net.URLConnection.getContentLengthLong(URLConnection.java:501)
	at java.net.URLConnection.getContentLength(URLConnection.java:485)
[...]

PR #4767, opened 2016-12-10
Not printing exception to the console

Akarshit:
Fix for #4732 
Completely removes the network error stack from Processing console.


benfry:
Looking for a more complete fix than this, rather than just hiding all exceptions. See my notes in the bug report.

PR #4774, opened 2016-12-14
Make color, fill, stroke et al. thread-safe: #3077

### LINKED ISSUE #3077: color() function is not thread-safe
The following code:

```
    for(int j=0; j < dimention; j++) {
      color current = ledImage[i].pixels[j];
      ledImage[i].pixels[j] = color(red(current), green(current), blue(current)); //multiply color correction here
    }
```

Was acting as a place holder. The plan was to multiply the Red, Green, Blue by color correction values for each pixel of an LED Wall. This function occasionally returned the wrong color value when threading (each thread handles a different i). I reimplemented it using bit shifting and it worked fine.

GKFX:
`calc***` color variables deleted; `colorCalc` now returns a `float[] {r,g,b,a}` array, with values from 0 to 1, like the old calc[RGBA] variables. The gray `color` methods don't bother with colorCalc at all.
The fill and stroke methods were also non thread-safe and values could easily pass between them.
I ran a number of tests with random inputs to the new `color` functions and they differed by one (eg. `ffE5E5E5` → `ffE6E6E6`) from the previous version in a tiny number (seven in 10 million) of cases due to different rounding but were otherwise identical.
Fixes #3077.

### DIFF
--- a/core/src/processing/awt/PGraphicsJava2D.java
+++ b/core/src/processing/awt/PGraphicsJava2D.java
@@ -2460,8 +2460,8 @@ public class PGraphicsJava2D extends PGraphics {
 
 
   @Override
-  protected void strokeFromCalc() {
-    super.strokeFromCalc();
+  protected void strokeFromCalc(float[] c) {
+    super.strokeFromCalc(c);
     strokeColorObject = new Color(strokeColor, true);
     strokeGradient = false;
   }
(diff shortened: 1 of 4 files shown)


benfry:
This runs counter to an intentional design decision in the code because it would mean a lot of allocation and garbage collection for that array as it's re-created and destroyed inside tight `draw()` loops. You'll notice that all the low-level functions in core are designed to avoid allocation wherever possible because they can lead to stuttering. A proper solution for this issue needs to minimize the use of `new` for a bunch of temporary variables.

GKFX:
@benfry That's fine; I'd like to replace the floats `fillR` etc with a `final float[] fillColor` that can be passed to calcColor, which would also return the color as an int in 0xAARRGGBB form, so that the color() functions would not need to pass the `float[]` array. Should I delete `fillR` etc., tag them depreciated but keep filling them, or leave them alone entirely? The same applies to stroke and others.

benfry:
Messing with `fillR` would break all renderer subclasses (OpenGL in particular). The solution would need to work across all renderers and not break existing code.

PR #4797, opened 2016-12-29
Unify mouse pressed/released events across renderers

### LINKED ISSUE #4361: Renderers but JAVA2D have problem for mouseButton under P3
while left mouse button is plessed, mouseButton ignores that right mouse button plessed or released.

```
void setup() {
  size(100, 100);
  //size(100, 100, P3D);

  fill(255);
}

boolean LMB, RMB;

void draw() {
  if (LMB) {
    background(0);
    if (RMB) {
      rect(30, 30, 40, 40);
    }
  } else {
    background(255);
  }
}

void mousePressed() {
  if (mouseButton==LEFT) {
    LMB=true;
  } else {
    RMB=true;
  }
}

void mouseReleased() {
  if (mouseButton==LEFT) {
    LMB=false;
  } else {
    RMB=false;
    if (LMB) {
      println("success");
    }
  }
}
```

for more details here: https://forum.processing.org/two/discussion/15531/how-to-detect-right-mouse-button-press-release-while-left-mouse-button-is-plessed-in-p3d-mode

JakubValtar:
OPENGL and FX renderers will now correctly report button which triggered
this pressed/released event (same as JAVA2D).

Previously:
- OPENGL would report currently pressed buttons in order LEFT, CENTER,
RIGHT regardless of which button triggered the event. E.g. when holding
LEFT and presing RIGHT, LEFT would be reported. When holding CENTER and
pressing LEFT, LEFT would be reported
- FX would report only first button which is down, so in RELEASE event
button would be missing

Now:
- event contains only button which triggred this event (button just
pressed or just released)

Fixes #4361

### DIFF
--- a/core/src/processing/javafx/PSurfaceFX.java
+++ b/core/src/processing/javafx/PSurfaceFX.java
@@ -806,11 +806,6 @@ public class PSurfaceFX implements PSurface {
     int count = fxEvent.getClickCount();
 
     int action = mouseMap.get(fxEvent.getEventType());
-    //EventType<? extends MouseEvent> et = nativeEvent.getEventType();
-//    if (et == MouseEvent.MOUSE_PRESSED) {
-//      peAction = processing.event.MouseEvent.PRESS;
-//    } else if (et == MouseEvent.MOUSE_RELEASED) {
-//      peAction = processing.event.MouseEvent.RELEASE;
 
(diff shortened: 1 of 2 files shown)


benfry:
Oh, how I love it when you fix these cross renderer, cross platform headaches like this. 😍

PR #4806, opened 2017-01-01
Replace keyChar with keyCode to prevent new line error when debugging

### LINKED ISSUE #4804: IDE debug stepping key short-cuts not working
Using the key short-cuts to debug (Ctrl+J and Ctrl+Shift+J) put newlines into the code.  Ctrl+Alt+J does not.  The debug command does run but it litters the code with newlines.

Processing 3.2.3, on Windows 10.

joelmoniz:
This PR replaces the keyChar check used to trigger auto-indent with a keyCode check, and fixes #4804. `Ctrl+J` happens to be an escape sequence representing new line (and their keyChar values are therefore both 10/`\n`), so `Ctrl+J` and `Ctrl+Shift+J` created a new line to appear when stepping. The keyCode for the Enter Key, however, is different, and hitting the Enter Key would be when the auto-format is triggered.

### DIFF
--- a/java/src/processing/mode/java/JavaInputHandler.java
+++ b/java/src/processing/mode/java/JavaInputHandler.java
@@ -164,7 +164,7 @@ public class JavaInputHandler extends PdeInputHandler {
         event.consume();
       }
 
-    } else if (c == 10 || c == 13) {  // auto-indent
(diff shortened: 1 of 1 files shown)


benfry:
Ah! Nice simple fix. Thank you.

PR #4808, opened 2017-01-02
Avoid missing check when gaining focus from a Sketch Disappeared popup

### LINKED ISSUE #4805: "Sketch disappeared" infinite pop up dialogs
When you open a sketch from removable media (SD card), eject the card, and then attempt to save, a dialog box will pop up saying that the sketch folder has disappeared. When you close that dialog, it will instantly pop back up, along with a dialog that says that the sketch could not be properly re-saved and to copy and paste the code into another text editor. When you close that dialog, it closes, but the first dialog remains. Closing the first dialog starts the cycle all over again. You can not recover your code by copy-paste due to the infinite dialogs. To close Processing, you either have to re-mount the removable media and successfully save or kill the process with the task manager and lose your work.
bad processing

joelmoniz:
This PR resolves #4805, which was happening because the `ensureExistence()` method was called each time the window gained focus: since closing the popup causes the window gains focus, this check and the accompanying warning dialogs would repeat ad infinitum.

In detail, what caused the bug is as follows:

1. The "Sketch Disappeared" popup appears and is closed.
2. Closing it gives the sketch window the focus, calling `ensureExistence()`
3. In the meanwhile, Processing realizes that it can't save the sketch, throwing the "Could not re-save sketch" popup
4. Closing the popup opened in 3 and then the one opened in 2 gives the sketch the focus again...

PS: I used the name of the JDialog since it seemed liked the simplest option- to avoid making changes to the structure of `ensureExistence()` (such as by returning false or re-throwing the error if the sketch does not exist), since several other methods call it, and I'll be happy to update this PR if another way of doing this might be better.

### DIFF
--- a/app/src/processing/app/ui/ChangeDetector.java
+++ b/app/src/processing/app/ui/ChangeDetector.java
@@ -14,8 +14,10 @@ import java.util.concurrent.ForkJoinPool;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import javax.swing.JDialog;
 import javax.swing.JOptionPane;
 
+import processing.app.Language;
 import processing.app.Messages;
 import processing.app.Preferences;
 import processing.app.Sketch;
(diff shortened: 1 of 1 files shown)


benfry:
@JakubValtar another to review for the `ChangeDetector`.

benfry:
Fixed with https://github.com/processing/processing/commit/b3763a8f8072d93178cd435ebc6f4a3a0f94adb9

PR #4834, opened 2017-01-17
Fix #4826. PShape in Java2D wasn't respecting kind.

### LINKED ISSUE #4826: PShape::beginShape() does not accept a shape type as an argument
Not sure if this is true in P3D, but in the P2D engine, the following code does not work as specified:

```
void display() {
    plus = createShape();
    plus.beginShape(LINES);
      // horizontal line
      plus.vertex(0, size/2);
      plus.vertex(size, size/2);
      // vertical line
      //plus.vertex(size/2, 0);
      plus.vertex(size/2, size);
    plus.endShape();
    
    shape(plus);
  }
```

`size` is passed into the class which this display method is inside of. In this case I am using `LINES`, which should create two lines which intersect at their midpoint. Instead I am getting the following, which is 3 lines, which is the expected behavior if there is no `shape type` passed in as an argument.

image

If I change the `shape type` to `POINTS`, it also does not work, again, defaulting to the typical `PShape::beginShape()` without an argument output.

Of course, this is an issue only if the `PShape::beginShape` method does not support arguments, but this documentation implies that it should behave like the `beginShape` function 

https://processing.org/reference/beginShape_.html
https://processing.org/reference/PShape_beginShape_.html - note the `syntax` section.

GKFX:
Fix #4826. I think the bug reporter is confused, since I can't reproduce the bug in P2D but I can in normal Java 2D. It's fixed in Java 2D now anyway.

### DIFF
--- a/core/src/processing/core/PShape.java
+++ b/core/src/processing/core/PShape.java
@@ -1604,9 +1604,9 @@ public class PShape implements PConstants {
     } else if (family == PRIMITIVE) {
       drawPrimitive(g);
     } else if (family == GEOMETRY) {
-      // same as path
-      drawPath(g);
-//      drawGeometry(g);
+      // Not same as path: `kind` matters.
+//      drawPath(g);
+      drawGeometry(g);
     } else if (family == PATH) {
       drawPath(g);
     }


benfry:
Thanks for your help!

PR #4848, opened 2017-01-23
Improvements to sketch launching and stopping

JakubValtar:
- Put stuff on correct threads and added proper synchronization
- Run button stays active after restarting sketch
- Run and Stop buttons now work reliably
- Restarting sketch with Run button while debugging is fixed


benfry:
Awesome, thank you.

PR #4849, opened 2017-01-23
Fix change detection when adding files to a sketch

### LINKED ISSUE #4713: Problem when we add tab in sketch folder
When a tab is add or remove from the desk in the sketch folder, Processing ask to update, if that it agree and we forget the the work code not saved, may be that can be good if Processing do an autosave before reloade ?

GKFX:
Change Detector now differentiates between a merge conflict (same file changed; will load the external changes if accepted by user; discards editor content) or a non-conflicting change (will load the external changes if accepted by user; nothing is lost). This is done per file so that if one tab does conflict and the next doesn't, the non-conflicting tab is not discarded for no reason.
Fixes #4713.

### DIFF
--- a/app/src/processing/app/Sketch.java
+++ b/app/src/processing/app/Sketch.java
@@ -758,6 +758,16 @@ public class Sketch {
 
 
   /**
+   * Ensure that all SketchCodes are up-to-date, so that sc.save() works.
+   */
+  public void updateSketchCodes() {
+//    if (current.isModified()) {
+    current.setProgram(editor.getText());
+//    }
+  }
+
+
+  /**
    * Save all code in the current sketch. This just forces the files to save
    * in place, so if it's an untitled (un-saved) sketch, saveAs() should be
    * called instead. (This is handled inside Editor.handleSave()).
@@ -767,9 +777,7 @@ public class Sketch {
     ensureExistence();
 
     // first get the contents of the editor text area
-//    if (current.isModified()) {
-    current.setProgram(editor.getText());
-//    }
+    updateSketchCodes();
 
     // don't do anything if not actually modified
     //if (!modified) return false;
@@ -911,9 +919,7 @@ public class Sketch {
 
     // grab the contents of the current tab before saving
     // first get the contents of the editor text area
-    if (current.isModified()) {
-      current.setProgram(editor.getText());
-    }
+    updateSketchCodes();
 
     File[] copyItems = folder.listFiles(new FileFilter() {
       public boolean accept(File file) {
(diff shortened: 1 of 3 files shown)


benfry:
@JakubValtar Can you review this? You were last in the change detection code…

JakubValtar:
@benfry Yes, sure thing.

JakubValtar:
@GKFX I like your idea and it works as expected, I just think the dialogs are way to complicated and as a user I'm really confused. This makes my brain melt :)

image

@benfry How do you think this should work? It's more of a design problem.

Current approach is to either
- scrap all changes in PDE and reload the sketch from the disk,
- or keep stuff in PDE and just mark appropriate tabs as modified.

Problem is when you have modified tabs and you change one file externally, you either have to scrap changes in all tabs and reload the sketch, or ignore the external changes.

The approach proposed in this PR is to detect tabs which were modified by external program (timestamp changed), and either
- save contents of all tabs modified only in PDE and reload the sketch (this is because there is no code to reload only some tabs so we have to save what we want to keep and reload the whole sketch),
- or keep stuff in PDE and just mark appropriate tabs as modified.

The idea is good because if you modify something externally, you probably want to reload just that and keep other tabs intact. The problem is that we have to save all other tabs in order to reload only one file, and it's really hard to explain what is going on.

I think to address the problem we should rather write code to reload separate tabs and show one dialog per tab (like other editors):
- modified: `File "file2.pde" has been modified by another program, do you want to reload it?`
- modified both externally and in PDE: `File "file3.pde" has been modified by another program, do you want to reload it and lose changes from Processing?`
- deleted: `File "file1.pde" was deleted, keep it in editor?`
- added: Add file to sketch automatically. If it's in the folder, it's part of the sketch and will appear anyway when you open the sketch next time.

benfry:
That sounds right with minor changes: 

* modified externally & no modifications in PDE – just grab the new code from the disk. I've noticed that Eclipse has started to simply reload files that are changed externally without asking, which I think is good behavior—why pester the user? 
* modified externally & unsaved changes – “This file was modified by another program. Would you like to keep this version or load the new changes?” … whichever version loses can be saved as `file3.pde.unsaved` or something like that, just in case
* deleted – `"file1.pde" has disappeared from the sketch folder, would you like to re-save it or remove it from your sketch?` with buttons "Re-save" and "Remove"
* added – just add immediately, no dialogs

GKFX:
@benfry Sounds good, I will implement if you guys haven't done it already.

JakubValtar:
@GKFX Thank you, we're looking forward to it!

GKFX:
Just to say that while I've left this inactive for a while, I am working on it again now.

benfry:
@GKFX Thanks, let us know when it's ready for review!

GKFX:
@benfry I've sent a new PR.

benfry:
Ok, thanks! Closing this one.

PR #4862, opened 2017-01-29
Fix a number of memory leaks.

jdf:
These leaks don't matter--could never be noticed--in Java Mode, which just System.exit()s when the sketch is done. But the Python Mode sketch runner sticks around between invocations, which makes it susceptible to severe memory loss when, for example, many PShapes are created.

This patch does a few things.

Most noticeably, it refactors PGraphicsOpenGL's resource disposal technique to remove a ton of duplicated code. But it also throws away the "RefList" which not only didn't do a desirable thing, but kept resources from ever being released (since there was no way for resources to be removed from the RefList other than by their being disposed() which could only happen if they were released from the RefList!).

It also terminates an exception-catching thread at shutdown time (which otherwise keeps running forever, and never lets go of its closure on PGraphicsOpenGL).

It clears a closure on PApplet from the ThinkDifferent QuitHandler.

This patch fixes https://github.com/jdf/processing.py/issues/233.

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -1194,7 +1194,7 @@ public class PApplet implements PConstants {
   *
   */
   public void pixelDensity(int density) {
-    println(density + " " + this.pixelDensity);
+    //println(density + " " + this.pixelDensity);
     if (density != this.pixelDensity) {
       if (insideSettings("pixelDensity", density)) {
         if (density != 1 && density != 2) {
@@ -3482,6 +3482,27 @@ public class PApplet implements PConstants {
   }
 
 
+  /**
+   * Pass a set of arguments directly to the command line. Uses Java's
+   * <A HREF="https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#exec-java.lang.String:A-">Runtime.exec()</A>
+   * method. This is different from the <A HREF="https://processing.org/reference/launch_.html">launch()</A>
+   * method, which uses the operating system's launcher to open the files.
+   * It's always a good idea to use a full path to the executable here.
+   * <pre>
+   * exec("/usr/bin/say", "-v", "Pipe Organ", "welcome to the command line");
+   * </pre>
+   * Or if you want to wait until it's completed, something like this:
+   * <pre>
+   * Process p = exec("/usr/bin/say", "waiting until done");
+   * try {
+   *   int result = p.waitFor();
+   *   println("the process returned " + result);
+   * } catch (InterruptedException e) { }
+   * </pre>
+   * You can also get the system output and error streams from the Process
+   * object, but that's more that we'd like to cover here.
+   * @return a <A HREF="https://docs.oracle.com/javase/8/docs/api/java/lang/Process.html">Process</A> object
+   */
   static public Process exec(String... args) {
     try {
       return Runtime.getRuntime().exec(args);
@@ -5927,6 +5948,8 @@ public class PApplet implements PConstants {
 
   /**
    * @webref output:files
+   * @param json the JSONObject to save
+   * @param filename the name of the file to save to
    * @see JSONObject
    * @see JSONArray
    * @see PApplet#loadJSONObject(String)
@@ -5938,7 +5961,7 @@ public class PApplet implements PConstants {
   }
 
   /**
-   * @nowebref
+   * @param options "compact" and "indent=N", replace N with the number of spaces
    */
   public boolean saveJSONObject(JSONObject json, String filename, String options) {
     return json.save(saveFile(filename), options);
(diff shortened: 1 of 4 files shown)


benfry:
Awesome; this is great work, thank you. 

@codeanticode Can you review the parts of the OpenGL parts? If they look good, we're good to merge.

codeanticode:
yep, will review asap! Thanks for these fixes, @jdf

jdf:
Don't thank me until it's been proven not to crash. :)

codeanticode:
@jdf The refactoring to reduce code duplication looks good to me. However, the use of the RefList is simply following the technique described in the [original article](http://www.oracle.com/technetwork/articles/java/finalization-137655.html):

_"Notice that reference objects are discovered by the garbage collector and added to their associated reference queues only if they are reachable themselves. Otherwise, they are simply reclaimed like any other unreachable object. This is why you add all NativeImage3 instances to the static list -- actually, any data structure will suffice -- to ensure that they remain reachable and processed when their referents become unreachable. Naturally, you also have to make sure that you remove them from the list when you dispose of them. This is done in the dispose() method."_

jdf:
Understood. Will ping again when I've corrected it.

jdf:
OK; please take another look.

JakubValtar:
I tihnk it's fixed. However, you also need to check type of argument in equals(), since you are mixing all the subclasses in one list, otherwise you get ClassCastException in `reachableWeakReferences.remove(this);`

JakubValtar:
If in doubt, run this stress test with `-Xmx32m` and print debug message in dispose()
```java
  public void settings() {
    size(640, 480, P3D);
    smooth(2);
  }

  public void setup() {
  }

  public void draw() {
    PGraphicsOpenGL pg = (PGraphicsOpenGL)createGraphics(width, height, P3D);
    pg.beginDraw();
    pg.background(random(255), random(255), random(255));
    pg.endDraw();
    image(pg, 0, 0);
  }
```

JakubValtar:
Looks good, thanks @jdf. Let's wait what @codeanticode thinks.

However, it does the same thing as before (just with less code), so you have to check if it fixes your problem. The code worked well previously (we tested it with @codeanticode before 3.0 came out), so I'm not sure how the leak could occur. Everything should be released in next garbage collector run, unless there is some other sneaky problem.

jdf:
@JakubValtar First, thanks very much for your help.

The test you suggested worked well to verify disposals taking place. (Although for Python Mode I had to give it 48M to run at all :} ).

I think the real problem was the non-exiting thread that waited forever for an unhandled exception to be thrown, along with the Apple-specific one.

JakubValtar:
@jdt You're welcome :) I ran it throught VisualVM (great for heap analysis) and I think you also need to clear the exception handler int PAppletJythonDriver at line 628.

jdf:
@JakubValtar I don't know who @jdt is, but I believe they spend a lot of time working in Eclipse. :)

Yes, I know about the PAppletJythonDriver fix; I've got that going in my own repo. Again, thank you.

For what it's worth, I use YourKit for memory debugging (and other profiling tasks). I'll also download VisualVM and give it a try.

JakubValtar:
Oops, sorry :)

VisualVM is part of JDK, at least on Windows (jviasualvm.exe). I never used YourKit but I think it's better than whatever ships with JDK.

jdt:
I have no idea what this project is, but I wholeheartedly support the elimination of memory leaks.

codeanticode:
Yes, I think the problem was the exception-catching thread, seems to me that now is properly terminated when the animation thread stops.

@benfry I think the GL fixes are ready to be merged.

benfry:
Great; merged.

codeanticode:
Just noticed that the System.err.println() line is still present in Disposable.dispose() https://github.com/processing/processing/blob/master/core/src/processing/opengl/PGraphicsOpenGL.java#L82

@jdf Shouldn't be removed?

jdf:
Yes.

neilcsmith-net:
@jdf I've just been looking at this pull request and testing the latest version of the Python mode, and there is definitely a memory leak still there!  Praxis LIVE actually does a similar thing to the Python mode here, and I need to update my fix hence looking at this.  I'm using reflection to clean out the reference lists and remove the PGL references in the resource classes at the end of each sketch run.

One problem is that the reference list is a static field - when I suggested using this method of cleanup to @codeanticode I suggested using instance fields, perhaps on PSurfaceJOGL, and explicitly disposing.  The second issue is that the resource classes have a strong reference to PGL in them which creates a circular reference that stops a lot of things being cleaned up.  Of course, if the second issue didn't exist, you'd end up cleaning up resources that no longer exist because the JOGL context has gone, so in some ways it's a memory leak masking something worse.

Issue #4499, opened 2016-05-21
Processing 3.1.1 with P2D/P3D on raspberry pi: mouseButton is "0"


tibfox:
under raspberry pi with the current version of processing (3.1.1) and enabled P2D (no experimental open gl) 

Now I have an issue with mouseButton and its value. All I get is "0" :(  

```
void setup(){
  //size(200,200);
size(200,200,P2D);
}

void draw(){
}

void mousePressed(){
   println(mouseButton);
}
```

prints out "0". Without "P2D" in size() i got the correct value "37".

After activating the experimental open gl driver via raspi-config -> "37" as mouseButton LEFT

thanks in advance

gohai:
Thanks for the helpful bug report @tibfox. I'll look into it.

gohai:
Confirmed. Looks like a problem either in the Pi-specific backend in JOGL, or in the Broadcom GLES2 driver itself. I will report to the JOGL team.

All mouse mouse events lack the button information, like this one:

````MouseEvent[EVENT_MOUSE_CLICKED, [117]/[64], button 0, count 1, rotation [0.0, 0.0, 0.0] * 1.0, pointer<1>[0/Mouse: 117/64, p[0.0/1.0=0.0]], InputEvent[modifiers: [], NEWTEvent[source:jogamp.newt.driver.bcm.vc.iv.WindowDriver, consumed false, when:1463935695926 d 1ms]]]````

gohai:
https://jogamp.org/bugzilla/show_bug.cgi?id=1309

gohai:
@tibfox This will finally be fixed in the upcoming release of Processing. Thanks for the report!

tibfox:
Super thanks a lot !

benfry:
@gohai which release are you referring to? Is this part of #4911?

gohai:
@benfry Yes! (meant: an upcoming release)

@tibfox You could download an updated [jogl-all.jar](https://github.com/gohai/processing/raw/094f95e0537cac8fdd91f91351c618272acb761c/core/library/jogl-all.jar) already and replace the existing one in `core/libraries`, if you wanted.

Issue #4807, opened 2017-01-02
Might be a text spacing bug in retina mode


ioeb:
MacBook Pro (Retina, 13-inch, Late 2013)
Intel Iris 1536 MB
OS X El Capitan - 10.11.6 (15G1212)
Processing 3.2.3

These two test-programs running

Test1
```
void setup() {
  size(480,320,P3D);
  pixelDensity(1);  
}
 
void draw() {
  text("this is text",10,10);
}
```
<img width="481" alt="test1" src="https://cloud.githubusercontent.com/assets/6170045/21589161/5417ba4a-d0ef-11e6-9d95-d1dd2d227abe.png">

Test2
```
void setup() {
  size(480,320,P3D);
  pixelDensity(2);  
}
 
void draw() {
  text("this is text",10,10);
}
```
<img width="482" alt="test2" src="https://cloud.githubusercontent.com/assets/6170045/21589165/5a9f968a-d0ef-11e6-9435-85452803d506.png">

jeremydouglass:
I can confirm that I'm also seeing the same strange spacing on Test2 above ( `pixelDensity(2)` ) on a 2016 MacBook Pro (13-inch, 2016) running macOS 10.12.1.

codeanticode:
ok, will try to fix asap.

Issue #4728, opened 2016-11-05
P3D fails on MBP 2014 when running on 4k external at 4k


IanBellomy:
The following one line sketch:

    size(100,100,P3D);

crashes when connected to external Sony Bravia (or using it as the sole/primary screen) while using a resolution higher than 1080p on the Sony.

Processing provides the error: `"Could not run the sketch (Target VM failed to initialize),"`

If the TV's resolution is set to 1080p or less, no problem.
If I use only the laptop's built in (Retina) screen, I can do `size(3840,2160,P3D);` and higher with no problem.
I can also run gigantic P2D sketches on the external (at 4K) without this issue. 

I understand that the error message means that there is not enough available memory, but I'm not sure why this would be the case even when the sketch size is so small. 

**Specs:**
Processing 3.2.2 
OSX 10.11.6 
Macbook pro Retina, Mid 2014 (16GB RAM, NVIDIA GeForce GT 750M 2048 MB)
Sony Bravia 4K XBR49X830C (2015)
Connected via HDMI

benfry:
What about just `size(100, 100)` without the P3D?

IanBellomy:
No problem.

IanBellomy:
Possibly relevant?

. . .

Crashed Thread: 4

Exception Type: EXC_BREAKPOINT (SIGTRAP) Exception Codes: 0x0000000000000002, 0x0000000000000000 Exception Note: EXC_CORPSE_NOTIFY

Application Specific Information: *** CFEqual() called with NULL first argument ***

. . .

[...]

PR #4911, opened 2017-02-16
Add more Raspberry Pi related fixes to JOGL

gohai:
This hopefully addresses #4499 and works around the long-standing dual driver stack issue on Raspbian.


benfry:
Is that a separate JOGL build you created?

gohai:
@benfry Yes. JOGL 2.3.2 plus the [following patches](https://github.com/gohai/jogl/commits/bcm-test) on top. Hope to at least fix one more bug, before we grow the repo size because of it. I'm in touch with @xranby about incorporating those in some form or another into upstream JOGL eventually also.

gohai:
@benfry I added one more fix, so from my perspective that'd be ready to go in. What this is improving is listed in the commit description.

benfry:
And no activity on the JOGL front, huh? So unlikely that this will be coming in an official build anytim soon? 

@codeanticode Any other JOGL tweaks we need before incorporating a new binary like this?

gohai:
I wouldn't hold my breath for an upcoming JOGL release ... Xerxes is in the know about those patches though

codeanticode:
I don't have any JOGL tweaks at this time. I will look at some GL issues as soon as I can (for example #4807), but I don't those require any fixes in JOGL

I've been following reports about problems connecting to HDMI/Hires displays, e.g.: #4728 #4954, and my concern is that those might be caused by some upstream problem in JOGL...

benfry:
#4728 and #4954 probably need some attention, since two similar reports likely means several people having issues. Though that code would be outside the jogl-all.jar that's being updated here, right? So maybe we go ahead with this stuff.

codeanticode:
I'm pretty confident that the issues with HDMI/hi-res output are entirely separate from the RPi fixes from Gottfried.

benfry:
I'm sure of that. The point was that this is an update to `jogl-all.jar` (the non-native part of JOGL), and those other issues all live inside the native code for macOS, right?

codeanticode:
Ok, got it. Let me look at the hi-res issues first, because I'm not certain whether those are due to some problem with the native libs, or in the platform-specific Java code.

benfry:
Great; thanks.

codeanticode:
I would go ahead with the merge as, regarding the issues I mentioned before, one appears to be caused by a [hardware failure](https://github.com/processing/processing/issues/4728#issuecomment-288917615) and the other can be solved with a [resolution-setting tool](https://github.com/processing/processing/issues/4954#issuecomment-288708794).

PR #4912, opened 2017-02-16
Attempt to ease `processing.java` argument passing

kritixilithos:
It should be noted that I speak for Processing on macOS.

In command-line Processing, `processing-java`, the argument for the sketch folder `--sketch=` has to be the full path to that directory, and this can get quite cumbersome. Digging in, I found that the reason was this [line] where the program changes directory and the origin of the directory passed in as an argument is lost.

I'll show you what I mean. Say you're in a directory `~` and you run `processing-java --sketch=test_sketch --run`, you get an error message saying that `test_sketch` cannot be found. Even when you prepend `~/` to the sketch name, you still get a similar error message. But only when you type everything out `/Users/user_name/test_sketch` will you be able to run the program. My Pull Request aims to solve this issue.

First of all, I replace all `~`s with `$HOME`s. Then I check if the sketch directory is valid. If it is, no problem. Otherwise, we prepend the directory from which the command was called.

You can see that this also modifies the `output`. The reason is similar. If you don't specify the full directory for the output folder, it will end up being wherever your [`javaRoot`](https://github.com/processing/processing/blob/a395991e7d9036f979dd77267a641f467cdfd93b/app/src/processing/app/tools/InstallCommander.java#L99) is. So I apply a similar, but simpler, logic for `--output`.

> Note: I have tested this by editing the `processing-java` directly in Bash and it has worked successfully (or at least as far as I can tell). But I haven't tested this Java code in `InstallCommander.java` (because I don't know how).

### DIFF
--- a/app/src/processing/app/tools/InstallCommander.java
+++ b/app/src/processing/app/tools/InstallCommander.java
@@ -94,7 +94,9 @@ public class InstallCommander implements Tool {
                    "    if [ \"$ARG\" = \"--build\" ]; then\n" +
                    "        OPTION_FOR_HEADLESS_RUN=\"-Djava.awt.headless=true\"\n" +
                    "    fi\n" +
-                   "done\n\n");
+                   "done\n" +
+                   "args=($(pwd) $@)\n" +
+                   "set -- ${args[*]}\n\n");
 
       String javaRoot = Platform.getContentFile(".").getCanonicalPath();
 
--- a/build/linux/processing
+++ b/build/linux/processing
@@ -13,6 +13,11 @@
 # Thanks to Ferdinand Kasper for this build script. [fry]
 
 
+# Make pwd as the first arg
+args=($(pwd) $@)
+set -- ${args[*]}
+
+
 # JARs required from JDK (anywhere in/below the JDK home directory)
 JDKLIBS="rt.jar"
 
--- a/java/src/processing/mode/java/Commander.java
+++ b/java/src/processing/mode/java/Commander.java
@@ -210,6 +210,11 @@ public class Commander implements RunnerListener {
       } else if (arg.startsWith(outputArg)) {
         outputSet = true;
         outputPath = arg.substring(outputArg.length());
+        outputPath = outputPath.replaceAll("~", System.getProperty("user.home"));
+        // Relative paths
+        if (!outputPath.matches("^(\\$|\\/).*")) {
+          outputPath = pwd + "/" + outputPath;
+        }
 
       } else if (arg.equals(forceArg)) {
         force = true;
--- a/java/src/processing/mode/java/Commander.java
+++ b/java/src/processing/mode/java/Commander.java
@@ -101,6 +101,14 @@ public class Commander implements RunnerListener {
     int task = HELP;
     boolean embedJava = true;
 
+    // Contains the current path
+    String pwd = "";
+    // Only for MacOS and Linux
+    if (!Platform.isWindows() && args.length > 0) {
+      pwd = args[0];
+      args[0] = "";
+    }
+
     try {
       if (Platform.isWindows()) {
         // On Windows, it needs to use the default system encoding.
@@ -173,9 +181,22 @@ public class Commander implements RunnerListener {
 
       } else if (arg.startsWith(sketchArg)) {
         sketchPath = arg.substring(sketchArg.length());
+        sketchPath = sketchPath.replaceAll("~", System.getProperty("user.home"));
         sketchFolder = new File(sketchPath);
+        // If the full path does not exist
         if (!sketchFolder.exists()) {
-          complainAndQuit(sketchFolder + " does not exist.", false);
+          if (Platform.isWindows()) {
+            complainAndQuit(sketchFolder + " does not exist.", false);
+          } else {
+            // Try the relative path
+            if (!new File(pwd + "/" + sketchPath).exists()) {
+              complainAndQuit(sketchFolder + " does not exist.", false);
+            } else {
+              // And then update sketchFolder and sketchPath
+              sketchPath = pwd + "/" + sketchPath;
+              sketchFolder = new File(sketchPath);
+            }
+          }
         }
         File pdeFile = new File(sketchFolder, sketchFolder.getName() + ".pde");
         if (!pdeFile.exists()) {


gohai:
As long as this doesn't break passing an absolute path (i.e. the current behavior), I am in favor of this change. The way I do on Linux normally: `cd` into the sketch folder and then simply run ``processing-java --sketch=`pwd` --run`` - but I also once wondered why this wasn't working with relative paths.

@kritixilithos Could the patch also handle arguments like like this `--sketch=./sketch`? I like this form because it does signal to the reader that the sketch folder is supposed to be present in the current directory.

kritixilithos:
I am going to repeat this again. This patch only changes `processing-java` for macOS (as you can tell by the call to `osascript` somewhere else in the file). I don't know what I would have to do for Linux or Windows.

@gohai `./sketch` notation  works just as well for me. If you want the pure bash for testing it out, you can see it [here](https://gist.github.com/kritixilithos/5c31a1a96b90f35ff4f0844a112f7b8d). (you may have to change the path to the Processing app since it only shows what it should look like for me)

gohai:
@kiwistrongis Since this equally affects macOS and Linux, I think both should be addressed, since we'd want the same behavior on each. (I don't think there is a Windows version of `processing-java`, but I might be mistaken.) Thanks.

kritixilithos:
I assume the @-ping was directed to me?

gohai:
@kritixilithos Indeed. Sorry for that!

gohai:
Another idea would be to make `--sketch` just take the following argument as the sketch folder. So instead of

    --sketch=foo --run

one would do

    --sketch foo --run

and the shell would take care of turning this into an absolute path for us.

kritixilithos:
Nice idea! Sure, I will have to modify how the shell works slightly, but I think I will also have to work on [`Commander.java`](https://github.com/processing/processing/blob/master/java/src/processing/mode/java/Commander.java). 

@gohai Do you know if there is a way for me to test the Java code once I modify it? I can test the shell script easily, but I also want to  be able to test the Java so I can see how it works when I run it.

kritixilithos:
Never mind, I forgot to read `README.md`. I found the build instructions and ran my original clone. Apparently there's a problem with (un)escaped slashes. Am working on it at the moment.

gohai:
@kritixilithos Was about to send the link to the build instructions - it's surprisingly painless to build!

Good point regarding `Commander.java`. I was wondering whether resolving this could be as straightforward as leaving the current behavior with `--sketch=` as-is, and simply changing `Commander.java` to:

* look whether the `--sketch` argument is preceded by a `=`, and - if not - treat the following argument as the path, as well as skip the following argument from evaluation inside the loop

kritixilithos:
@gohai Just to clarify what you have said: if `--sketch` is preceded by a `=`, then treat it as normal, otherwise, treat the next argument as the path. Does that sound right?

Also, I believe that the current behaviour with `--sketch=` should be changed to accommodate what you just said about the `=` because as of now, it **only** changes the path if the argument is in `--sketch=` format. But you're right, most of the work will have to be done in `Commander.java`

benfry:
The = thing is *way* too subtle. Most apps do = or not as a fairly arbitrary thing.

kritixilithos:
Ah, okay then. I haven't started working on the `=` yet, so I will just not implement it now.

Now all that's left for me is to make sure this (the directory name argument) is implemented in other systems as well as macOS

kritixilithos:
Now I made some big changes to the original PR so that the changes can be easily ported to other operating systems and not just macOS. (along the way I accidentally messed up indentation because of `gg=G`)

Now all that's left for me is to port it to Linux.

@gohai You said you have Processing on Linux. Could you please confirm whether your `/usr/local/bin/processing-java` (or wherever else it is located) looks like [this](https://github.com/processing/processing/blob/master/build/linux/processing)? If not could you `cat` it? I want to have an idea of what I'm working with so that I know where to make changes.

gohai:
@kritixilithos Yes, that's it. (see [build.xml](https://github.com/processing/processing/blob/master/build/build.xml#L795))

kritixilithos:
@gohai All that's left now is to test my commit for Linux. Could you please confirm that it works for the following paths:

```Bash
processing-java --sketch=/full/path/to/test_sketch --run
processing-java --sketch=~/path/to/test_sketch --run
processing-java --sketch=test_sketch --run
processing-java --sketch=$HOME/path/to/test_sketch --run
```

gohai:
@kritixilithos I get this:

    /usr/local/bin/processing-java: 17: /usr/local/bin/processing-java: Syntax error: "(" unexpected

I saw that your git history contains a lot of unrelated, e.g. white-space, changes that got inadvertently committed, but which makes it harder to figure out what's going on. You might consider copying the files you changed to a separate location on your machine before running

    git reset --hard 1c218e2f2d4e31614411b3aef5ef5b79a00a923f

This resets your tree to the most recent commit that you share with the main Processing tree. Then you can put your files back into your working directory and commit only the relevant changes with

    git add -p

This asks you for each chunk that differs whether you want to add them to the commit or not (`y` or `n`).

After you have committed your change, pushing your branch to GitHub will fail, since you have "rewritten" its history, and its record of whats happened doesn't match the one on GitHub. Use this command to erase the history on GitHub in favor of your cleaned up one (this is fine in this context, and will automatically update this PR as well, no need to open a new one or close the current one:

    git push -f

kritixilithos:
Ok, so I redid everything. Now before going into Linux, I must first ensure that `--output=`'s behaviour is similar to that of `--sketch=` instead of them being different

kritixilithos:
Now I have to identify whether the output path is the full path or the relative path. Hmm... I think that if the path begins with `/` or `~` or `$some_variable` it will be a full path. Otherwise, it's a relative path.

@gohai Do you think that the above is enough for the output directory?

kritixilithos:
@gohai I had a look at line 17.

```sh
args=($(pwd) $@)
```

I have no idea why there was an error on this line. I pushed again, and I hope it works this time. Could you please test it again?

gohai:
@kritixilithos Downloaded and tested again, same error

    pi@raspberrypi:~/sketchbook/foo $ processing-java --sketch=`pwd` --run
    /usr/local/bin/processing-java: 17: /usr/local/bin/processing-java: Syntax error: "(" unexpected

kritixilithos:
@gohai That is such a weird error. I have tested that line of code multiple times in the shell and it has given me no such error. Just to be sure, what is on line 17 of your `/usr/local/bin/processing-java`?

```Bash
cat -n /usr/local/bin/processing-java | grep "^\s\+17"
```

gohai:
pi@raspberrypi:~ $ cat -n ~/Downloads/processing-master/build/linux/work/processing-java | grep "^\s\+17"
    17	args=($(pwd) $@)

kritixilithos:
I have no idea on what is going wrong here. It should work, but it doesn't. I have tested this numerous times yet I still am clueless. I tried running the file on mac (using `./processing`) but it gives me no syntax errors.

<sub>(Maybe it's vim's linefeed at the end of the file. But it shouldn't matter.)</sub>

alphaCodeo:
Since you seem to know about using processing-java, can you help with this? I can run and compile everything fine using processing-java.exe. But when I try to run the .class(had used --build) it says "Error: Could not find or load main class [MySketchName]". It runs fine using --run.

benfry:
Doing some repo cleanup so I'm closing since it's been a year w/o updates. Generally speaking, relative paths might be fine for someone who wants to get those working and test them across all three of our platforms, but `~` or variable expansion isn't something we should be getting into. 

This is a lower priority, though… The command line version is really there for external editors or scripts; typing these things out on the command line manually seems tedious whether or not you have relative path support.

PR #4920, opened 2017-02-18
Report more error conditions in validateFramebuffer

gohai:
I don't know what any of this means, but hope the gesture will be apprechiated :)


JakubValtar:
@gohai Good idea! Since we are messing with this, can you please look at [glCheckFramebufferStatus docs](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glCheckFramebufferStatus.xhtml) and add all the codes which are documented there? Please don't delete the ones which are not on the page, I think they might be encountered when running older GL versions. If the constants are not in `com.jogamp.opengl.GL`, they will be in `com.jogamp.opengl.GL4`.

gohai:
@JakubValtar Should be done. I picked `com.jogamp.opengl.GL3` since this was also was JOGL was using for those constants.

codeanticode:
Yes, looks good.

JakubValtar:
Nice work, please merge

Issue #4919, opened 2017-02-18
Framebuffer issues on Mesa VC4 graphics


gohai:
Out of the P3D examples, fifteen currently fail with the open source Mesa driver for the Raspberry Pi.

The output (with Mesa debugging enabled) is always similar to this:
````
Framebuffer error (unknown error), rendering will probably not work as expected Read http://wiki.processing.org/w/OpenGL_Issues for help.
Mesa: User error: GL_INVALID_FRAMEBUFFER_OPERATION in glClear(incomplete framebuffer)
Mesa: User error: GL_INVALID_FRAMEBUFFER_OPERATION in glDrawArrays(incomplete framebuffer)
Mesa: User error: GL_INVALID_FRAMEBUFFER_OPERATION in glClear(incomplete framebuffer)
OpenGL error 1286 at bot beginDraw(): invalid framebuffer operation
Mesa: User error: GL_INVALID_FRAMEBUFFER_OPERATION in glBlitFramebuffer(incomplete draw/read buffers)
Mesa: User error: GL_INVALID_FRAMEBUFFER_OPERATION in glDrawArrays(incomplete framebuffer)
Mesa: User error: GL_INVALID_FRAMEBUFFER_OPERATION in glBlitFramebuffer(incomplete draw/read buffers)
````

This is the stack trace of the first error that occurs with the `Basics/Form/Primitives3D` example:
````
Framebuffer error (unknown error), rendering will probably not work as expected Read http://wiki.processing.org/w/OpenGL_Issues for help.
Mesa: User error: GL_INVALID_FRAMEBUFFER_OPERATION in glClear(incomplete framebuffer)
com.jogamp.opengl.GLException: Thread[main-FPSAWTAnimator#00-Timer0,5,main] glGetError() returned the following error codes after a call to glClear(<int> 0x4500): Unknown glGetError() return value: ( 1286 0x506), 
	at com.jogamp.opengl.DebugGL4bc.writeGLError(DebugGL4bc.java:31803)
	at com.jogamp.opengl.DebugGL4bc.glClear(DebugGL4bc.java:1251)
	at processing.opengl.PJOGL.clear(PJOGL.java:1823)
	at processing.opengl.PGL.createFBOLayer(PGL.java:975)
	at processing.opengl.PGL.beginRender(PGL.java:727)
	at processing.opengl.PGraphicsOpenGL.beginOnscreenDraw(PGraphicsOpenGL.java:6567)
[...]

PR #4921, opened 2017-02-18
Retry with multisampling disabled if creating a framebuffer fails because of INCOMPLETE_MULTISAMPLE

gohai:
No idea if tearing down the FBO half-way and setting it back up again is recommended practice, but this does make the `Primitives3D` example work for me with the Mesa VC4 driver (see #4919).


gohai:
This depends on #4920 for the new constant.

codeanticode:
I think is a good fallback, go ahead and merge.

PR #4928, opened 2017-02-21
Fixed #4600: Adding an option for tab size to the preference UI

### LINKED ISSUE #4600: add option for tab/indent length
Some people just like to have more spaces.
Secondary request: use 'hard' tabs instead of spaces.

meer1992:
The [current situation](https://github.com/processing/processing/issues/4600) is letting the user to manually change the tab size in the preferences.txt file.
To enhance the usability of our users, an option is added in the preference UI.  The user can change it within the range of 1 to 10.


meer1992:
@benfry What do you think of the enhancement? One note to mention is that I have not translated the label yet to other languages.

benfry:
We try to keep as few options as possible in the Preferences dialog—it's not like we don't know how to write the code for a UI for this feature. I think the question would be whether it would make sense to remove something else if we're going to add this. 

For instance, is this a more useful feature, for more people, than enabling/disabling smoothing in the editor? At this point, maybe we could swap them. But it's a very low priority.

meer1992:
I agree with you :) I will look for a high priority issue which I could help

benfry:
Thanks—and I'm happy to consider arguments that this is more useful than the text anti-aliasing at this point, but it seems like another option is that we need something like `about:config` from Firefox or `about:flags` from Chrome that makes all the prefs available for people who want to tweak (and to help prevent breaking their `preferences.txt` file).

PR #4935, opened 2017-02-27
Fix #4825: Out of memory exception when using a Timer class in EditorConsole 

### LINKED ISSUE #4825: Array print issue
When Processing tries to print out 200 elements in 3 arrays, Processing itself freezes and must force quit the application. macOS 10.12.2

meer1992:
The issue #4825 describes that having too many outputs in the console can sometimes freeze the program and lead to out of memory exception. An example as @Lord-of-the-Galaxy pointed out is:
```
void setup(){
  size(100, 100);
}

void draw(){
  for(int i = 0; i < 10000; i++){
    for(int a = 0; a < 10000; a++){
      print(i + " " + a);
    }
    println();
  }
}
```

The **main reason** is because of the synchronisation structure in the [EditorConsole](https://github.com/processing/processing/blob/master/app/src/processing/app/ui/EditorConsole.java). The _appendString()_ keeps appending string and the _insertAll()_ method fails to sync, and as a result we get out of memory exception.

I thought of having a more **concurrent structure** rather than **complete synchronization**. This will provide us scalability and stability. I used _ScheduledExecutorService_ which can achieve the same behaviour as the timer but it is more manageable and it's making sure that the memory is available.

Tested on Mac 10.11.6 | java version "1.8.0_92"


benfry:
Thanks for looking into it… will take a look when I have a moment to test on Windows and Linux as well. (Or if someone else can do the same, please have at it…)

meer1992:
Thank you. I will also do a test on Linux and Windows

JakubValtar:
@meer1992 Thanks for identifying this problem and submitting a solution, it's really appreciated. However there are some threading bugs. We can't accept this PR because almost every sketch uses console and it is critical for console to work correctly at all times.

- You use `SheduledExecutorService`. The difference is that `Timer` runs the task on EDT - Event Dispatch Thread (the only thread which should update UI), but the `ScheduledExecutorService` runs it on it's own thread. This is a threading bug (because `flush()` should be called only on EDT) and the `Timer` should stay there.
- All the `Atomic*` objects and `ConcurrentLinkedQueue` give you a false sense of security. You don't use any advanced methods (like `getAndSet()` or `compareAndSet()`) so it is no better than using `volatile` primitives. Which means that you just removed all locking (all threads can touch anything anytime), because you also removed `synchronized` keyword. This is another threading bug which will leave you with corrupted state.

The whole problem can be summarized as single reader thread and multiple writer threads. Right now, all methods are `synchronized` on a single instance lock, which means that you can not read and write at the same time. If you write too much then read might be waiting for the lock forever and you end up out of memory.

To fix this, you need to separate read lock and write lock. This is a great moment to use a `BlockingQueue`, because you can add from multiple threads and also drain the whole queue in one atomic thread-safe operation.

If you are interested in making a new PR (and I would be happy if you do), here are some steps which should get you through:
- Keep using the `Timer`.
- Use `LinkedBlockingQueue<ElementSpec>` instead of the `ArrayList`. Leave `synchronized` on `appendString()` because you need to protect the fields from concurrent access when multiple threads write to the console at the same time.
- Remove `synchronized` keyword from `insertAll()` and use `queue.drainTo(someList)` to drain the queue as a single atomic operation. This separates read and write lock and makes sure that `insertAll()` is never waiting and queue is drained regularly.
- Remove `hasAppendage` because it would be modified from multiple threads (reading and writing thread) and instead every time check `queue.isEmpty()` which is properly synchronized by the queue.

meer1992:
@JakubValtar Thanks for the feedback. Yes. I am looking forward to improve the solution.

benfry:
@meer1992 Do you plan to update this fix or should we close this PR?

meer1992:
@benfry You can go ahead and close the PR.
 
I have been struggling solving the issue, and so far no updates.

Issue #4540, opened 2016-06-19
Add resize() method


cchan:
This probably has been mentioned before, but I couldn't find any similar github issues and I thought I would mention it.

The following variable-sizing code seems to work fine. I'd assume there's some significant overhead associated with surface.\* functions, but since it's only supposed to be used on setup anyway it's alright.

```
void size2(int w, int h){
  surface.setResizable(true);
  surface.setSize(w, h);
  surface.setResizable(false);
}

void setup() {
  int w = round(random(200, 500));
  int h = round(random(200, 500));
  size2(w, h);
}
```

DominusDodo:
A quick sketch cycling through 100 of your setup calls a frame takes ~300ms or less a frame. That's about 3ms per resize, which seems perfectly reasonable for a single call. I don't know if there are any reasons against implementing a function like this. Perhaps a function like this:

```java
public void resize(int w, int h) {
  surface.setResizable(true);
  surface.setSize(w, h);
  surface.setResizable(false);
}
```

usser123:
Using setResizable() and setSize() did not work with P3 renderer. I get NullPointerException error after resizing. This occurs when I try to access pixels array. It seems like after resizing, the pixels array is not accessible because the program fails at location 0.

benfry:
This issue is resolved with `windowResize()` and the other `windowXxxx()` methods in Processing 4.

Download it here: https://github.com/processing/processing4/releases

PR #4951, opened 2017-03-08
Added resize() method for PApplet

bergvel1:
Issue [#4540](https://github.com/processing/processing/issues/4540) suggests the addition of a resize() method to change the sketch window size with a single line of code. The current way to do this seems to be

`surface.setResizable(true);`
`surface.setSize(w, h); `
`surface.setResizable(false);`

The proposed change allows changing the window size with:

`resize(w, h);`

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -1230,6 +1230,21 @@ public class PApplet implements PConstants {
   }
 
 
+  /*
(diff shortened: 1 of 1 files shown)


benfry:
Unfortunately, making this work consistently across all renderers is significantly more work than just these three lines of code, otherwise we'd have done it already.

Issue #4944, opened 2017-03-06
Arabic characters not supported in default fonts used in the PDE


11maxed11:
Hey,
I am working on adding Arabic translation to Processing. Most things get viewed correctly. However, some of them (as viewed in the picture) are not. Any idea on the reason behind that, and whether a quick fix exists for that?

![capture](https://cloud.githubusercontent.com/assets/3680307/23592954/b694b192-0208-11e7-9833-2a36316a7d93.PNG)

![untitled2](https://cloud.githubusercontent.com/assets/3680307/23592955/b696b398-0208-11e7-95db-2cb57c323f6c.png)

Any notes on the support for Right-to-Left languages in Processing?

11maxed11:
@benfry Hello, do you have any thoughts on this? If the reason cannot be pointed out, I can do a partial translation that involves all the parts that work and leaves out the errors, in addition to providing a full translation document to be used later on --if the issue gets fixed.

benfry:
Thanks for looking into it. The issue with the boxes is that the built-in font (Source Sans Pro) doesn't have Arabic characters, so we need to figure out how to set a [fallback](https://docs.oracle.com/javase/8/docs/technotes/guides/intl/font.html) font when the characters are out of range. It looks like it involves messing with these [configuration](https://docs.oracle.com/javase/8/docs/technotes/guides/intl/fontconfig.html) files, though it'd be nice to figure out if we can handle it from code.

Assuming you're not looking to dive into that part of it, though, some ideas to get things moving:

* If you modify `lib/theme.txt` and swap `processing.sans` with `Dialog`, more of those characters should start working
* If some, but not all, are working, please report back and I'll make sure that all of them can be edited from that file so you can continue your work

Longer-term, I'm not sure what sort of right-to-left issues we might have, since we use many custom components in the interface, but I'd love to see how far we can get without requiring a huge rewrite. (The editor text area, I can almost guarantee, is unlikely to be happy about RTL rendering, for instance.)

Let me know how it goes!

11maxed11:
Hey!
Thanks for the support. I am willing to look further into this.
I'll do the experimentation advised by you, and try to get some clear results while solving any issues along the way. I'll also try to check how well RTL languages are handled and highlight the result, so we'll see how well can this go with the current codebase.

11maxed11:
I went ahead and submitted a PR with the translation file in PR #4970 . This problem can be addressed later in a separate PR.
[...]

PR #4970, opened 2017-03-20
Arabic translation

11maxed11:
Hi,

This PR adds Arabic translation to Processing.
All lines have been translated with the except of the longest one, which has HTML markup in it. It's quite annoying to mix-up HTML with RTL languages and it doesn't get displayed well when tested; so I left it as it is.

The problem mentioned in issue #4944 still remains. 

I'd like to keep translating anything that might be added later. Is there any way one can be listed as an official translator and be notified on that, or should I find a way to receive notifications on file edits? i.e. if the English language file changed, I'll know new stuff should be translated to Arabic..


benfry:
Very cool, thank you! Sorry it's taken so long for me to get to this. 

We don't currently have a useful way to keep track of translators, but that's a good idea. @REAS have you made any progress with web translation that might help here?

I think there are ways to set up a watch for files here on Github, but I don't know offhand. Let us know what you find out, because we should also add that to the translation page.

REAS:
We plan to start working on the translation systems again in June of this year. We'll likely be discussing over at /processing/processing-docs/

PR #4981, opened 2017-03-26
Created bug list and reported a dangerous bug

NateUniAccount:


benfry:
This is not the way to submit issues or bug reports. However, the issue in that file is unlikely to be addressed: the `launch()` command is a low-level feature that carries with it the responsibility of the creator of a sketch (and user of that same sketch) to be sensible about this sort of thing.

NateUniAccount:
Thanks for the reply, it's been added as an issue now. I agree it's the user's responsibility to be sensible about what code they run, on the other hand there's no need to make it so easy for a user to shoot themselves in the foot, especially when there's a much safer way of doing the same thing.

PR #5011, opened 2017-04-17
Window placement and pixel density cleanup

JakubValtar:
TODO:
- pixelDensity() needs JNA to do anything on Windows
- test/fix things on Linux
- implement Present mode for FX
- overhaul window-out-of-visible-area detection


benfry:
Kaboom!

PR #5014, opened 2017-04-20
Unblock hardware-accelerated P3D on ARM Mali devices

gohai:
This works around a JOGL bug that affects ARM Mali devices, such as the PocketCHIP. (reported by @xranby)

### DIFF
--- a/core/src/processing/opengl/PSurfaceJOGL.java
+++ b/core/src/processing/opengl/PSurfaceJOGL.java
@@ -273,6 +273,15 @@ public class PSurfaceJOGL implements PSurface {
       } else if (PJOGL.profile == 2) {
         try {
           profile = GLProfile.getGL2ES2();
+
+          // workaround for https://jogamp.org/bugzilla/show_bug.cgi?id=1347
+          if (!profile.isHardwareRasterizer()) {
+            GLProfile hardware = GLProfile.getMaxProgrammable(true);
+            if (hardware.isGL2ES2()) {
+              profile = hardware;
+            }
+          }
+
         } catch (GLException ex) {
           profile = GLProfile.getMaxProgrammable(true);
         }


xranby:
This is a workaround for ARM Mali devices when both Mesa 3D software rendered OpenGL and ARM OpenGL ES drivers are installed at the same time on the same system.

core/src/processing/opengl/PSurfaceJOGL.java
```diff
@@ -273,6 +273,15 @@ protected void initGL() {
       } else if (PJOGL.profile == 2) {
         try {
           profile = GLProfile.getGL2ES2();
+
+          // workaround for https://jogamp.org/bugzilla/show_bug.cgi?id=1347
+          if (!profile.isHardwareRasterizer()) {
+            GLProfile hardware = GLProfile.getMaxProgrammable(true);
```
xranby: if i am not mistaken you should use
GLProfile.getMaxFixedFunc(true);
here.
see: https://jogamp.org/bugzilla/show_bug.cgi?id=1347#c2

gohai: Hm, I don't understand - I thought fixedFunc means GL2ES1, whereas we want GL2ES2, which is also what the hardware is capable of?

xranby: please double check that GLProfile hardware = GLProfile.getMaxProgrammable(true); gives you GLES2 since I think you may en up with the Mesa 3D software renderer because Mesa 3D do support a very high Programmable profile using software renderer.

gohai: Seems to give me the hardware profile - here are my debug printlns while I was testing the code above:

    Trying GL2ES2
    libGL error: MESA-LOADER: malformed or no PCI ID
    libGL error: unable to load driver: sun4i-drm_dri.so
    libGL error: driver pointer missing
    libGL error: failed to load driver: sun4i-drm
    Got GLProfile[GL2/GL2.sw]
    Not hardware
    Max programmable is GLProfile[GLES2/GLES2.hw]
    Picking max programmable over software

xranby: OK, good!

PR #5021, opened 2017-04-22
Make the change detector not reload the sketch

### LINKED ISSUE #4713: Problem when we add tab in sketch folder

GKFX:
Fixes #4713, fixes #4849. I've added a method to Messages that accepts custom buttons, and Messages.showYesNoQuestion is now based on it for Mac OS.

### DIFF
--- a/app/src/processing/app/ui/ChangeDetector.java
+++ b/app/src/processing/app/ui/ChangeDetector.java
@@ -3,7 +3,6 @@ package processing.app.ui;
 import java.awt.EventQueue;
 import java.awt.event.WindowEvent;
 import java.awt.event.WindowFocusListener;
-import java.nio.file.Files;
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
@@ -30,7 +29,7 @@ public class ChangeDetector implements WindowFocusListener {
   private final Sketch sketch;
   private final Editor editor;
 
-  private List<String> ignoredAdditions = new ArrayList<>();
+  //private List<String> ignoredAdditions = new ArrayList<>();
   private List<SketchCode> ignoredRemovals = new ArrayList<>();
   private List<SketchCode> ignoredModifications = new ArrayList<>();
 
@@ -103,12 +102,12 @@ public class ChangeDetector implements WindowFocusListener {
         .collect(Collectors.toList());
 
     // Added files that are actually candidates for a new tab
-    List<String> addedTabsFinal = addedFilenames.stream()
-        .filter(f -> !ignoredAdditions.contains(f))
-        .collect(Collectors.toList());
+    //List<String> addedTabsFinal = addedFilenames.stream()
+    //    .filter(f -> !ignoredAdditions.contains(f))
+    //    .collect(Collectors.toList());
 
     // Take action if there are any added files which were not previously ignored
-    boolean added = !addedTabsFinal.isEmpty();
+    boolean added = !addedFilenames.isEmpty();
 
 
     // REMOVED FILES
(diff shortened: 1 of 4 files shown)


benfry:
@JakubValtar have a moment to review?

JakubValtar:
@benfry I'm travelling, but I should have time today evening.

benfry:
Thanks!

build/shared/lib/languages/PDE.properties
```diff
@@ -445,6 +445,18 @@ ensure_exist.messages.unrecoverable.description = Could not properly re-save the
 # Check name
 check_name.messages.is_name_modified = The sketch name had to be modified. Sketch names can only consist\nof ASCII characters and numbers (but cannot start with a number).\nThey should also be less than 64 characters long.
 
+# External changes detector
+change_detect.reload.title=Tab modified externally
+change_detect.reload.question="%s" was modified by another program.
+change_detect.reload.comment=Would you like to keep this version or load the new changes?\nEither way, the version you discard will be saved to your sketch folder.
```
JakubValtar: @benfry I know you proposed this message before, but I think it's confusing. 

image

Something along these lines seems more clear:

image

app/src/processing/app/ui/ChangeDetector.java
```diff
@@ -150,39 +168,111 @@ private synchronized void checkFiles() {
...
+                  Messages.showWarning("Move failed",
+                      "Could not move the external editor's"
+                    + " version of " + scKeep.getPrettyName() + " to a safe"
+                    + " location; make a copy of it before saving the sketch"
+                    + " if you need it.", e);
+                  ignoredModifications.add(scKeep); // No infinite loops.
+                  return;
+                }
+                try {
+                  scKeep.save();
```
JakubValtar: @benfry Do we want to save the tab right here if the user decides not to reload external changes? Other text editors don't save. Maybe updating the timestamp of the SketchCode to the timestamp of the file (so it won't pop up again until the file is changed externally again) and marking the SketchCode as modified is more what users are used to?

benfry: Yeah, that latter sounds correct.

build/shared/lib/languages/PDE.properties
benfry: The language needs to be rewritten in the first, but it's mostly in the direction we need to go. I think dropping the final sentence would get rid of most of the awkwardness. It could even be shown in a second dialog box (or just in the console as a message). 

Your suggested version works great for people already familiar with IDEs or advanced text editors that have this kind of feature. For the majority of our audience, it's not clear what "reload" really means in this context: do I lose my work? Does it save something? What's happening?

JakubValtar: Ok, understood. I think my problem was with 'keep this version'. What is a version? Does *this* refer to the externally modified version from the previous sentence?

Something like 'Do you want to keep what is in the *test5* tab or reload the tab from the file?' may be better?

Just a proposal, it's your call.

benfry: Right, I think it needs to be rewritten, “this version” is confusing.

GKFX:
Okay, I've made the changes. That's excluding the UI text, since it makes more sense for you guys to tweak that, so `change_detect.reload.comment` is now inaccurate.

JakubValtar:
@GKFX Thanks, looks great!

@benfry It's merge time! When it's done I'll push some changes on top (run it on EDT and nudge the error checker).

benfry:
Great, thanks guys.

JakubValtar:
@benfry Follow-up here: #5075

PR #5027, opened 2017-04-24
Improve loadBytes performance

gohai:
I was surprised that loading a 54 MB file with `loadBytes` took as long as it did (2.7 seconds). Hit the code with a hammer a little bit, and now it's down to 0.4 seconds. And seems to works for URLs still...

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -7174,14 +7174,15 @@ public class PApplet implements PConstants {
    */
   static public byte[] loadBytes(InputStream input) {
     try {
-      BufferedInputStream bis = new BufferedInputStream(input);
       ByteArrayOutputStream out = new ByteArrayOutputStream();
+      byte[] buf = new byte[4096];
 
-      int c = bis.read();
-      while (c != -1) {
-        out.write(c);
-        c = bis.read();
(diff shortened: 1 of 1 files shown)


benfry:
Hm, this is literally what `BufferedInputStream` should be doing, which IIRC is why it was originally written in this goofy manner. 

Actually, I think the performance problem may be that `createInput()` now (3.3.1 vs 3.3) creates a `BufferedInputStream` by default, so this is now doubling that… What happens if you just remove the extra `BufferedInputStream` wrapper and keep reading byte by byte?

gohai:
When I get rid of the `return new BufferedInputStream(input)` in both `createInput` variants, and replace them by a simple `return input`, this still takes the same time as before for me.

benfry:
Alright, thanks for checking. (Having done various rounds of optimization on those functions, I'm surprised that it was using the previous method…) Also, did you try 8192 or other buffer sizes? (Usually 8k is more of the sweet spot…)

gohai:
I just tried - they seem to be about the same as 4k on my machine, but hard to measure with page caching and all.

benfry:
Ok, thanks. Also checked in some additional optimized versions that will be used when loading from a file or URL (where we can get the size beforehand).

PR #5042, opened 2017-05-02
Revert updatePixels() to set modified flag to true

### LINKED ISSUE #5040: tint(rgb, alpha) bug
I think this is a **tint(rgb, alpha)** error (processing 3.3.2 windows10):
The same code works well with 3.3 version.
```
PGraphics pg0;
float t0=128;

void setup() 
{
  size(200, 200);
  pg0 = createGraphics(200, 200);
  pg0.smooth();
  pg0.beginDraw();
  pg0.fill(255,0,0);
  pg0.rect(30, 30, 100, 100);
  pg0.endDraw();   
}

void draw() 
{
  background(255);
  // uncomment the following line to get NullPointeException  
  //tint(255,t0); 
  image(pg0, 0, 0);    
}

void mousePressed() 
{
  t0 = random(20, 255);
  println(t0);
}
```

JakubValtar:
This fixes the bug for now, but uncovers a deeper issue.

PImage.modify flag is now used for two purposes and should be split into two variables:
- internally by renderers to signal that pixels[] changed and need to be drawn
- externally to signal other renderers that this image changed in some way and they should update their cache

When a PGraphics is drawn onto other PGraphics, both of these uses collide and may cause problems. This needs to be looked at and resolved later, probably by adding another variable (a flush counter which could be compared with couter of the cached image to tell if cache needs to be updated).

Reverts 51dff7689968c5da7344b61d53e90359eaf8329f

Fixes #5040

### DIFF
--- a/core/src/processing/awt/PGraphicsJava2D.java
+++ b/core/src/processing/awt/PGraphicsJava2D.java
@@ -2797,7 +2797,7 @@ public class PGraphicsJava2D extends PGraphics {
     if (pixels != null) {
       getRaster().setDataElements(0, 0, pixelWidth, pixelHeight, pixels);
     }
-    modified = false;
+    modified = true;
   }
 
 


JakubValtar:
Sorry 'bout that. Hope it's the last problem caused by my pixelDensity update.

benfry:
Plenty to go around… you're talking to the guy who broke `keyPressed` in the last release.

PR #5050, opened 2017-05-04
Fix keyPressed for multiple keys

### LINKED ISSUE #5049: problem when pressing and holding key
I am using the latest version of processing which is 3.3.3.

just add this code to processing and when you run the code, hold any key for (5 sec) and release it.
IT WOULD NOT TURN BACK TO (FALSE)

```processing
void draw() {
  if (keyPressed) {
    println(true);
  } else {
    println(false);
  }
}
```

JakubValtar:
Java2D and FX2D send multiple PRESSED and only one RELEASE event (at
least on Windows). Therefore we have to keep track of what is pressed
and what not. Most keyboards do not support pressing more than ~10 keys
simultaneously, so this should not cause any performance problems.

See focusLost() for possible problem when user overrides it and does not call super. It's there only to clear keys when ALT+TAB is pressed so the chance somebody will hit this is low.

Fixes #5049

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -658,7 +658,7 @@ public class PApplet implements PConstants {
    * @see PApplet#keyReleased()
    */
   public boolean keyPressed;
-  int keyPressedCount;
+  List<Long> pressedKeys = new ArrayList<>(6);
 
   /**
    * The last KeyEvent object passed into a mouse function.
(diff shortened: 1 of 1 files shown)


GoToLoop:
Perhaps a _**HashSet**_&lt;Long> would be a better fit than an _**ArrayList**_&lt;Long> for field _pressedKeys_ here.

Using a _**Set**_ instead, statement `if (!pressedKeys.contains(hash))  pressedKeys.add(hash);`
can be replaced by merely `pressedKeys.add(hash);`, given a _**Set**_ can't have duplicated elements after all.

Due to its hashable nature searching, a _**Set**_ should be more performant when using methods such as **contains(Object)** & **remove(Object)** than a _**List**_.

GoToLoop:
On the other hand, if the # of elements in the container _pressedKeys_ is always small, I guess a _**List**_ might be indeed faster than a _**Set**_ in this case.  O_o

PR #5055, opened 2017-05-06
Consider $SUDO_USER on Linux for locating the sketchbook folder

gohai:
This is #5054 but also overwrites `getSettingsFolder` taking `$HOME` into account, so that `sudo -E` works. Since now the root user is now potentially writing into the settings folder, the patch tries to make those files world-writable.

Compared to #5054, this also handles these situations:
* more than one user running `sudo -E processing` (previously this created settings in the root users home directory pointing the sketch directory to the first that called)
* user having a custom path for the sketch folder set
* the root user having `.processing` & `sketchbook` herself (this is still possible to make use of by running `sudo` without the `-E`)

Modifying all files that get created by P5 to be world-writable is probably a futile attempt, but if this is the way to go I would consider adding this at least for:
* creating the sketchbook and .processing folders (so that `sudo -E` without them being already present wouldn't create folders that can later not be written into)
* Processing's `save` family of functions

### DIFF
--- a/app/src/processing/app/platform/LinuxPlatform.java
+++ b/app/src/processing/app/platform/LinuxPlatform.java
@@ -77,6 +77,16 @@ public class LinuxPlatform extends DefaultPlatform {
   // Attempt to work around this in the least invasive manner,
   // so that "sudo -E processing" or "sudo -E processing-java"
   // will pick up the invoking user's sketchbook folder instead.
+  public File getSettingsFolder() throws Exception {
+    String sysEnvHome = System.getenv("HOME");
+    if (sysEnvHome != null && 0 < sysEnvHome.length()) {
+      return new File(sysEnvHome, ".processing");
+    } else {
+      return super.getSettingsFolder();
+    }
+  }
+
+
   public File getDefaultSketchbookFolder() throws Exception {
     String sysEnvHome = System.getenv("HOME");
     if (sysEnvHome != null && 0 < sysEnvHome.length()) {
--- a/app/src/processing/app/platform/LinuxPlatform.java
+++ b/app/src/processing/app/platform/LinuxPlatform.java
@@ -22,7 +22,10 @@
 
 package processing.app.platform;
 
+import java.io.BufferedReader;
 import java.io.File;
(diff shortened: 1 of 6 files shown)


benfry:
Doesn't `-H` handle this with `sudo`?

gohai:
It's a bit odd `sudo -H` doesn't seem to work on the Pi (Debian derivative), because of some settings in `/etc/sudoers` that sanitize the environment. `sudo -E` does work, probably because it _forces_ the environment to be preserved.

I think I found a way for doing this even with plain `sudo`: `$SUDO_USER` seems to be always set to the original user - one could get to this user's home directory with `echo ~$SUDO_USER`.

@benfry Do you think this is worth it? Or should I just have the reader make a symlink from their sketchbook folder to `/root/sketchbook instead?

gohai:
I like the way this looks and functions now. Believe all that is missing would be to sprinkle some `setWritable(true, false)` at places...

app/src/processing/app/platform/LinuxPlatform.java
```diff
@@ -67,6 +73,53 @@ public void initBase(Base base) {
...
+      try {
+        homeDir = getHomeDir(sudoUser);
+      } catch (Exception e) {}
+    }
+
+    return homeDir;
+  }
+
+
+  public static String getHomeDir(String user) throws IOException {
```
benfry: `PApplet.createReader()` here will save a lot of lines and these funky indents.

benfry:
Started review but decided easier to just fix it up… Fixed placement of a comment, catching `Exception` instead of `IOException` (could cause an NPE or others), other shortening misc.

benfry:
I'm a little wary of the tacit encouragement this gives people to run the environment using `sudo`, but I guess the alternative is worse, that people would simply run as `root`.

gohai:
Thanks @benfry. I'll look for your edits, to pick up a trick or two! Agree about not wanting to promote `sudo`

gohai:
(with `createReader` this is indeed way nicer!)

PR #5077, opened 2017-05-20
Prompt the user to give the L suffix on numbers

### LINKED ISSUE #4878: Warn user to use L when creating a number constant that won't fit into an int
Processing's support for the 'long' data type is mostly useless. The statement like:
long c = 55666777888 is not allowed..?? It would be nice if the next edition would
support the use of 'long' in functions.

GKFX:
Fixes #4878.
`long l = 121238081273;` gives "The type int can't handle numbers that big. Try 121238081273L to upgrade to long."
Anything else gives "Even the type *(e.g.)* long can't handle 8237418923078232103L. Research big numbers in Java."

### DIFF
--- a/java/src/processing/mode/java/Compiler.java
+++ b/java/src/processing/mode/java/Compiler.java
@@ -166,6 +166,8 @@ public class Compiler {
           exception = new SketchException(errorMessage);
         }
 
+        String[] parts = null;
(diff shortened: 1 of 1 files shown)


benfry:
Great, thanks!

PR #5082, opened 2017-05-22
Build out shell a bit more

gohai:


benfry:
Responding to your email…

`shellQuoted()` was only added b/c it was necessary—I think we may need to debug why it wasn't working for you.

And the concern I'd have with just using `/bin/sh` is that folks might be disappointed when their init or config stuff isn't available—isn't that part of the point of having `shell()` as a separate command?

gohai:
I'll put in the code that emulates the interactive shell then!

gohai:
@benfry With the latest commit:

    StringList out = new StringList();
    StringList err = new StringList();
    shell(out, err, "echo $PATH");
    println(out);

Gives me the same the same PATH from inside Processing compared to when query this in the terminal. (Tested on macOS.)

gohai:
Also tested on latest Raspbian.

benfry:
Also the null/non-null thing is going to be a problem in code: I think you're gonna need to cast the null to a `StringList` object, because a null next to `String...` cannot be distinguished by the compiler. Please double-check. 

Do we really need an additional version with just stdout? I lean toward just keeping it simple… either you want the output or not, and my general experience has been that programs are inconsistent enough about what goes in err and what goes in out…

gohai:
Double-checked the situation with the variable  number of arguments - seems to work 👍

Thinking a bit about my use cases: stderr is certainly useful for user-facing errors, but in many situations I'd forgo the extra line that constructs its `StringList` object because I'll anyway do a regex on the stdout - and if the number of matches isn't expected, _this_ would be my error condition.

You bring up a good point that stdout vs stderr isn't really so consistent anyway. Is there a possibility of combining both streams in a way that keeps the ordering of lines perhaps? (On UNIX one could alternatively do `2>&1` in the `cmd`, but this should better be parsed, since people might want to do their own tricks with re-directions inside the argument.) What would you think about folding both into a single `StringList` - vs. having variants with one or two `StringList` arguments?

PR #5096, opened 2017-05-27
Removed useless comment

Maharal:


GKFX:
Unfortunately you can't just do that. Those lines are there for a reason; the javadoc is generated from XML files and if you really want to get rid of them, you'd have to redesign the script that does it: https://github.com/processing/processing-docs/blob/master/java_generate/GenerateDescriptions.py.

Maharal:
Thanks for the information

PR #5097, opened 2017-05-28
Suggestions for change in the PVector class

Maharal:
I believe that I improved some parts of the code


benfry:
Most of those are slower than the current implementations (changing assignments to method calls and then assignments) or unnecessary (setting x, y, z to 0 in the empty constructor).

PR #5100, opened 2017-05-28
Suggestion of the function lerpColor for a set of colors

Maharal:
I overwritten the _lerpColor_ method for an implementation that supports many colors.

### DIFF
--- a/core/src/processing/core/PGraphics.java
+++ b/core/src/processing/core/PGraphics.java
@@ -7996,7 +7996,25 @@ public class PGraphics extends PImage implements PConstants {
   public int lerpColor(int c1, int c2, float amt) {  // ignore
     return lerpColor(c1, c2, amt, colorMode);
   }
-
+  
+  
+   /**
+   * Generalization of the function lerpColor for many colors that are passed 
+   * by parameter as a vector of colors
+   *
+   * @webref color:creating_reading
+   * @usage web_application
+   * @param c[] interpolate to this set of color
+   * @param amt between 0.0 and 1.0
+   * @see PImage#blendColor(int, int, int)
+   * @see PGraphics#color(float, float, float, float)
+   * @see PApplet#lerp(float, float, float)
+   */
+  int lerpColor(int c[], float amt) {
+    int index = int(c.length * amt);
+    return lerpColor(c[index], c[index + 1 < c.length ? index + 1 : c.length-1], amt * c.length - index);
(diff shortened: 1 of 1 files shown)


core/src/processing/core/PGraphics.java
```diff
@@ -7996,7 +7996,25 @@ public final float brightness(int rgb) {
...
+   * @webref color:creating_reading
+   * @usage web_application
+   * @param c[] interpolate to this set of color
+   * @param amt between 0.0 and 1.0
+   * @see PImage#blendColor(int, int, int)
+   * @see PGraphics#color(float, float, float, float)
+   * @see PApplet#lerp(float, float, float)
+   */
+  int lerpColor(int c[], float amt) {
+    int index = int(c.length * amt);
```
GKFX: That's a syntax error outside of the PDE. You meant `(int)(c.length * amt)`. Also, index is now (c.length - 1) for all of the final (e.g.) ⅓ of a three-color gradient, giving the result shown:
wrong
So what you really need is
```java
int index = (int)((c.length-1) * amt);
```
Then you need to check the boundaries with if blocks so if index is not out of range (i.e. `index < 0` or `index > c.length-1`) you set it to within the range.

core/src/processing/core/PGraphics.java
```diff
@@ -7996,7 +7996,25 @@ public final float brightness(int rgb) {
...
+    return lerpColor(c[index], c[index + 1 < c.length ? index + 1 : c.length-1], amt * c.length - index);
```
GKFX: Finally, `amt * c.length - index` needs to be changed to `amt * (c.length-1) - index` to match the change made above. Result:
right

core/src/processing/core/PGraphics.java
```diff
@@ -7996,7 +7996,25 @@ public final float brightness(int rgb) {
...
+   *
+   * @webref color:creating_reading
+   * @usage web_application
+   * @param c[] interpolate to this set of color
+   * @param amt between 0.0 and 1.0
+   * @see PImage#blendColor(int, int, int)
+   * @see PGraphics#color(float, float, float, float)
+   * @see PApplet#lerp(float, float, float)
+   */
+  int lerpColor(int c[], float amt) {
```
Maharal: Thanks for the comment. I did not know the signature _lerpColor(float, int ...)_.

PR #5101, opened 2017-05-28
Add matrix mult in a vector

Maharal:


kritixilithos:
```
@brief Multiply a vector by a scalar
```

Should "scalar" here instead be "matrix"?

benfry:
Closing since we have other versions of this covered by `PMatrix3D` (which seems a more appropriate place for it).

PR #5106, opened 2017-06-02
add install / uninstall scripts for linux a-la-Arduino + correct mime types for IDE

dcuartielles:
Yo!
we had to celebrate that Carlos finished his thesis and decided to go in a hacking session ... and guess what ... we thought that it would be super-cool if the Processing IDE that we adore had a proper install script the same as the Arduino one.
We modified ant's build.xml file to add the files, as well as added the proper application files, install scripts, etc.
Using this patch, you will be able of installing the right icons in linux and let your users click directly on *.pde files for them to open automagically.
The last thing ... you should allow Processing to continue talking in 3rd person about itself ... it is such a nice IDE that we love and respect ...

Hugs from Sweden,

/Carlos and David

PS. you should revise the metadata for this to properly represent the Processing Foundation, I am not sure you will like how we wrote it ... we tried our best :-)

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -1,13 +1,13 @@
 <?xml version="1.0"?>
 <project name="Processing" default="build">
-  
+
   <!-- Sets properties for macosx/windows/linux depending on current system -->
   <condition property="platform" value="macosx">
-    <os family="mac" /> 
+    <os family="mac" />
   </condition>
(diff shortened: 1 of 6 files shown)


benfry:
Awesome! Thanks David & Carlos!

PR #5115, opened 2017-06-07
Add console hiding button

### LINKED ISSUE #970: add ability to show/hide the console for low-resolution screens
_Original author: brendiee...@yahoo.com (December 12, 2011 23:11:49)_

<b>If possible, please attach the smallest possible sketch that reproduces</b>
your problem. First make a tiny example, then use Tools -&gt; Archive Sketch
<b>to create a zip file, and then attach it to this bug report.</b>

<b>What steps will reproduce the problem?</b>
1. Start Program
2. Try to use at on low-resolution screen.

<b>What is the expected output? What do you see instead?</b>
Program usability is expected.
Commandline window uses one forth of screen real estate.

<b>What version of the product are you using? On what operating system?</b>
Current Release/Affects all.

<b>Please provide any additional information below.</b>

Hi,

The console window is much too large in that it use around 1/4 of the vertical space on 768 pixel screen. I would like the command-key and menu option to toggle it on or off entirely for coding or turn it on for debugging. In this way, we can see more of the program structure, allowing us to write better programs.

Once more, the attractive lower light-blue bar block (sans-text) seems to have no purpose (Dunsel).

Toggling and allowing for a greater console resize are solutions that are respectful to the users needs and the reality of smaller screens.

Thank you,
BrendaEM

_Original issue: http://code.google.com/p/processing/issues/detail?id=932_

GKFX:
Fixes #970. Gives a bit more room to code on a smaller screen. I haven't automatically unhidden it when the sketch starts since that's just bothering the user and the status bar still changes colour on exceptions. Also, I'm not leaving space for the `RIGHT_MARGIN` in EditorStatus; should I be? (N.B. you have push access to my PR branches, feel free to make trivial changes to constants/text etc. rather than describing them to me when that's convenient).
Also (mouseExited listener) fixed a bug where if you moved your mouse off an URL in the status bar vertically it would stay the hovered-over color because it only cares about `mouseX`.
image
image

### DIFF
--- a/app/src/processing/app/ui/Editor.java
+++ b/app/src/processing/app/ui/Editor.java
@@ -320,6 +320,17 @@ public abstract class Editor extends JFrame implements RunnerListener {
         status = new EditorStatus(this, Editor.this);
         return status;
       }
+
+
+      @Override
+      public void finishDraggingTo(int location) {
+        super.finishDraggingTo(location);
+        // JSplitPane issue: if you only make the lower component visible at
+        // the last minute, its minmum size is ignored.
+        if (location > splitPane.getMaximumDividerLocation()) {
+          splitPane.setDividerLocation(splitPane.getMaximumDividerLocation());
+        }
+      }
     });
 
     box.add(splitPane);
--- a/app/src/processing/app/ui/EditorStatus.java
+++ b/app/src/processing/app/ui/EditorStatus.java
@@ -79,20 +79,22 @@ public class EditorStatus extends BasicSplitPaneDivider {  //JPanel {
   String url;
   int rightEdge;
   int mouseX;
-  boolean urlRollover;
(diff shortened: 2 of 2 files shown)


benfry:
I think we could probably incorporate this if you tweak it a bit for the just-added “clipboard” icon in the status bar. I'd put this to the right of it (but in a separate `drawString()`, using unicode arrows for the shape, and that'd likely work.

GKFX:
@BenFry Sure, will have a look over the next few days if possible.

GKFX:
I've merged the code for the two buttons so that both display as a Unicode character centred in a square, which can then be highlighted nicely on hover. All seems to work fine.

benfry:
Thanks; giving this a shot.

PR #5126, opened 2017-06-12
BugFix for Duplicate Library Conflicts.

diwi:
Hi, 

_(this is my first PR for processing. So im not sure if i did everything right so far ... pls let me know.)_

This PR tackles the "Duplicate Library"-Conflict. 
The problem occurs, when libraries are based on same 3rd party libraries (same or different version), like Fisica and Box2D for Processing both use jBox2d. 

Fix:
In case of multiple matching libraries ( = possible duplicates) only the
required one is chosen, by making a guess based on the other, guaranteed,
libraries in use.
In a first pass, the guaranteed libraries are cached.
In a second pass, unresolved package-library mappings are generated, based on pass 1.

old Error message:
conflict

fixed:
conflict

Not sure if these verbose print-warnings are necessary, but i kept them for the moment. Most users dont have to deal with possible duplicates anyways i guess.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@
 ._*
 *~
 /build/shared/reference.zip
+/bugfix_info/processing_bugfix_duplicateLibraryConflict_new.png
+/bugfix_info/processing_bugfix_duplicateLibraryConflict_old.png
+/build/ant_run.bat
--- a/app/src/processing/app/Mode.java
+++ b/app/src/processing/app/Mode.java
@@ -30,6 +30,7 @@ import java.awt.image.WritableRaster;
 import java.io.*;
 import java.util.*;
 import java.util.List;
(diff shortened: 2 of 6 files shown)


diwi:
any thoughts, suggestions?

diwi:
For Debugging, I am using this debug sketchbook. 
It contains 3 librarys, all dependent on box2d (different versions), as well as 1 top level box2d library.
To use it, set the sketchbook path to the extracted folder, and run this PR-branch.

Download: [Sketchbook_DebugDuplicateLibraryConflicts.zip](https://github.com/processing/processing/files/1099841/Sketchbook_DebugDuplicateLibraryConflicts.zip)


duplicatelibraryconflict_sketchbookscreen


I could only test on windows so far, but i expect it to work on other systems as well.

PR #5138, opened 2017-06-17
Add defaultFontOrDeath to OpenGL textWidth.

### LINKED ISSUE #5137: textWidth() causes a NullPointerException for P2D and P3D unless a font is explicitly specified
```java
void setup() {
    size(100, 100, P2D); 
    //textFont(createFont("Consolas", 12)); 
    println(textWidth('c'));
}
```

This code seems to be the culprit (line 3485 of PGraphicsOpenGL):

`Object font = this.textFont.getNative();`

GKFX:
Fix #5137.


benfry:
Good catch, thanks!

Issue #2319, opened 2014-01-19
PVector is mutable but has no copy constructor.


GKFX:
Opened by @alexkindel at https://github.com/processing/processing-web/issues/334:

> There is no copy constructor for PVector even though it is mutable.
> 
> Example: `PVector p1 = new PVector(100., -100.);`
> `PVector p2 = new PVector(p1);` // The constructor PVector(PVector) is undefined.
> 
> A vector class is useful because of encapsulation. Having a copy constructor allows users to think of vectors as a discrete object instead of having to think about their makeup.

PR #5146, opened 2017-06-21
Adding copy constructor to the class PVector

cimmic:
I think it is good practice to have a copy constructor in a class.
I have often experienced that it would be easier to write more clear code if PVector had a copy constructor.

### DIFF
--- a/core/src/processing/core/PVector.java
+++ b/core/src/processing/core/PVector.java
@@ -123,6 +123,16 @@ public class PVector implements Serializable {
 
 
   /**
+  * Copy constructor.
+  */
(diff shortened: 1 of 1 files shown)


JakubValtar:
Hi @cimmic, thanks for your pull request!

I'm sorry, but we already had this argument before and the current stance is that Processing don't use any copy constructors. You can use `vec.copy()` to get a copy of your `PVector`.

>  _benfry commented on 22 Jan 2014  https://github.com/processing/processing/issues/2319#issuecomment-32988471_
>
> We don't use copy constructors anywhere else in the API, so it would be out of place.
>

PR #5152, opened 2017-06-24
Handle curly quotes well

### LINKED ISSUE #5133: Detect errors involving curved quotation marks
I was helping some students go through [some online examples](https://docs.google.com/document/d/1NWp53xKKdWX6YT7bAIozGum8iWPctzZfJNDOaMTmnzo/pub). Many of them attempted to copy statements directly from the example, some of which use string literals with curved quotation marks instead of straight ones (like how they would be if typed manually):

```
text(“Stroke size: 3”, x, y)
```

instead of

```
text("Stroke size: 3", x, y)
```

This results in the linting error:

```
Syntax error on tokens, ArgumentList expected instead
```

and compilation error:

```
unexpected char: '\'
```

This might also happen if the source code is put through a word processor.

Would it be possible for Processing to reword these messages? The cause in this case is rather difficult for anyone to spot.

GKFX:
Fix #5133.
The autoformatter now automatically corrects curly quotes when they're used to wrap string or character literals, provided this is safe to do. It is also now less upset by unterminated strings.
On compiling and editing, errors relating to them are reworded.
I've also started prioritising the errors spat out by ECJ since occasionally the first one on the line is not the most relevant one.

### DIFF
--- a/app/src/processing/app/Base.java
+++ b/app/src/processing/app/Base.java
@@ -55,9 +55,9 @@ import processing.data.StringList;
 public class Base {
   // Added accessors for 0218 because the UpdateCheck class was not properly
   // updating the values, due to javac inlining the static final values.
-  static private final int REVISION = 262;
+  static private final int REVISION = 263;
   /** This might be replaced by main() if there's a lib/version.txt file. */
-  static private String VERSION_NAME = "0262"; //$NON-NLS-1$
+  static private String VERSION_NAME = "0263"; //$NON-NLS-1$
   /** Set true if this a proper release rather than a numbered revision. */
 
   /**
--- a/app/src/processing/app/Language.java
+++ b/app/src/processing/app/Language.java
@@ -95,6 +95,7 @@ public class Language {
       "el", // Greek
       "es", // Spanish
       "fr", // French, Français
+      "it", // Italiano, Italian
       "ja", // Japanese
       "ko", // Korean
       "nl", // Dutch, Nederlands
--- a/app/src/processing/app/ui/Editor.java
+++ b/app/src/processing/app/ui/Editor.java
@@ -3120,20 +3120,14 @@ public abstract class Editor extends JFrame implements RunnerListener {
 
   /**
    * @return the Problem for the most relevant error or warning on 'line',
-   *         defaulting to the first.
+   *         defaults to the first error, if there are no errors first warning.
    */
   protected Problem findProblem(int line) {
-    int currentTab = getSketch().getCurrentCodeIndex();
-    return problems.stream()
-        .filter(p -> p.getTabIndex() == currentTab)
-        .filter(p -> {
-          int pStartLine = p.getLineNumber();
-          int pEndOffset = p.getStopOffset();
-          int pEndLine = textarea.getLineOfOffset(pEndOffset);
-          return line >= pStartLine && line <= pEndLine;
-        })
-        .findFirst()
-        .orElse(null);
+    List<Problem> problems = findProblems(line);
+    for (Problem p : problems) {
+      if (p.isError()) return p;
+    }
+    return problems.isEmpty() ? null : problems.get(0);
   }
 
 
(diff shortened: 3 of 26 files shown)


GKFX:
Also I've forced `preproc.substitute_unicode` to false when previously it was the default. Is that likely to be an issue?

JakubValtar:
Looks good. However I think we need to give these priority treatment in the error checker (the same as missing curly braces), otherwise ECJ is going to spit out bunch of nonsense like this:

image

Would you mind if I clone your branch, commit some changes to rectify this, and send you a message so you can merge them back before we accept this PR?

GKFX:
@JakubValtar I don't mind at all. Please could you submit it as a PR to my branch when you're done? Also I feel that that makes my prioritization code a bit redundant, so you could delete that if you think that's best?

GKFX:
Ignore that merge commit, it's got an extra 8. Will fix shortly.

benfry:
@JakubValtar did you make any progress with your plan to clone and clean this up? Or is that merged back in here?

JakubValtar:
It's still on my todo list. Gonna do this by the end of this week.

benfry:
Got it; thanks!

JakubValtar:
@GKFX I sent the pull request, sorry it took me a while :/

Thanks for the original PR, good stuff! I just moved stuff around a bit to make it ready for more custom errors and new build system.

All error checking is now done in PDEX.ErrorChecker and problems are prioritized manually so that user sees only the most severe problems (curly quotes > curly braces > rest), showing other stuff when program structure is messed up does not make much sense and gives lots of false alarms.

I removed the prioritozation code, becasue it was not doing much right now, but if we need to fine tune the priorities in the future we can revive it.

GKFX:
@JakubValtar Looks awesome, thank you!

benfry:
Does that mean this is ready to merge @JakubValtar?

JakubValtar:
@benfry I think it is ready.

Check the error messages whether you like them.

One thing which surfaced is that autoformat might break strings which
contain curly quotes.

For sure it does not format well this:

```
String s = “te“xt”;
```

Not sure if also:

```
String s = "te“xt";
```

JakubValtar:
@benfry I think it is ready.

Check the error messages whether you like them.

One thing which surfaced is that autoformat might break strings which
contain curly quotes.

For sure it does not format well this:

```
String s = “te“xt”;
```

Not sure if also:

```
String s = "te“xt";
```

GKFX:
@JakubValtar 
```java
String s = "te“xt";
```
definitely should be unharmed; I tested strings like that. If the opening quote is straight, the autoformatter doesn't give curly quotes any special meaning. If the user has managed to type
```java
String s = “te“xt”;
```
then I think that all quotes will be straightened (I'm not at my development machine right now and can't check what happens.) But it would be quite hard to distinguish that from something like
```java
String s = “te“.xt(“”);
```
so straightening all quotes that aren't in a real string seems like a reasonable thing to do.

JakubValtar:
@GKFX Thanks, I just tested it and it's fine. Well-formed strings are not affected and in others quotes are straightened. Yesterday I somehow I managed to get a curly quoted string shredded onto multiple lines, but can't reproduce, so it's probably some obscure scenario which will happen to some unlucky soul once in a decade.

@benfry Merge time?

benfry:
Thanks to you both. Now merged.

JakubValtar:
Great! All credit goes to @GKFX for diving into it.

benfry:
This appears to be breaking unicode characters: https://github.com/processing/processing/issues/5413 

Presumably because of this change in `PdePreprocessor.java`:

```
    // Removing all the Unicode characters makes detecting and reporting their
//    if (Preferences.getBoolean("preproc.substitute_unicode")) {
//      program = substituteUnicode(program);
//    }
```

Can you two look into a fix?

benfry:
Just confirmed that it's due to that change; not a problem on macOS, but affects Windows. Haven't checked Linux.

benfry:
(Let's continue discussion over at #5413)

PR #5184, opened 2017-07-20
Fix bugs in line vert shader

### LINKED ISSUE #5181: Lines with perspective have skewed ends based on sketch apect ratio
```processing
void setup() {
  size(1000, 500, P3D);
}

void draw() {
  hint(ENABLE_STROKE_PERSPECTIVE);
  
  background(0);
  stroke(255);
  strokeWeight(20);

  translate(width/2, height/2);
  line(200, -200,  0, -200,  200,  0);
}
```

Actual:
image

Expected:
image

### LINKED ISSUE #5182: Lines with perspective at zero depth have different thickness than without perspective
```processing
void setup() {
  size(1000, 500, P3D);
}

void draw() {
  hint(ENABLE_STROKE_PERSPECTIVE);
  
  background(0);
  stroke(255);
  strokeWeight(20);

  translate(width/2, height/2);
  line(-200, 0,  0, 200, 0,  0);
}
```

Actual:
image

Expected:
image

### LINKED ISSUE #5183: Lines with one end on eye plane disappear
In the code below, the middle line with end at Z==0 disappears.

```processing
void setup() {
  size(500, 500, P3D);
}

void draw() {
  background(0);
  stroke(255);
  
  // set modelview to identity so that
  // our coordinates go straight into the shader
  setMatrix(new PMatrix3D());
  
  for (int i = -10; i <= 10; i++) {
    line(10 * i,  50, i,
         10 * i, -50, -500);
  }
}
```

Actual:
image

Expected:
image

### LINKED ISSUE #5194: Stroke join disappears when line end is behind the camera
```processing

void setup() {
  size(500, 500, P3D);
}

void draw() {
  
  background(0);

  translate(width/2, height/2);
  
  stroke(255);
  strokeWeight(20);
  noFill();
  
  beginShape();
  
  PGraphicsOpenGL pg = (PGraphicsOpenGL) g;
  
  vertex(0, -200);
  vertex(200, 0);
  vertex(-200, 0, pg.cameraZ+1); // add 1 to place it behind the camera
  
  endShape();
}
```

Actual:
image

Expected:
image

JakubValtar:
Fixes #5181
Fixes #5182
Fixes #5183
Fixes #5194

point vert updated with the code from line vert to keep them in sync

### DIFF
--- a/core/src/processing/opengl/shaders/LineVert.glsl
+++ b/core/src/processing/opengl/shaders/LineVert.glsl
@@ -52,7 +52,8 @@ void main() {
   // formula to convert from clip space (range -1..1) to screen space (range 0..[width or height])
   // screen_p = (p.xy/p.w + <1,1>) * 0.5 * viewport.zw
 
-  // prevent dividing by W by transforming the tangent formula
+  // prevent division by W by transforming the tangent formula (div by 0 causes
+  // the line to disappear, see https://github.com/processing/processing/issues/5183)
   // t = screen_q - screen_p
   //
   // tangent is normalized and we don't care which direction it points to (+-)
@@ -66,7 +67,8 @@ void main() {
   // t = +- normalize( ((q.xy*p.w - p.xy*q.w) / (p.w*q.w)) * 0.5 * viewport.zw )
   //
   // remove the common scalar divisor/factor, not needed due to normalize and +-
-  // (keep viewport - can't remove because it has different components for x and y)
+  // (keep viewport - can't remove because it has different components for x and y
+  //  and corrects for aspect ratio, see https://github.com/processing/processing/issues/5181)
   // t = +- normalize( (q.xy*p.w - p.xy*q.w) * viewport.zw )
 
   vec2 tangent = normalize((q.xy*p.w - p.xy*q.w) * viewport.zw);
@@ -79,6 +81,7 @@ void main() {
 
   // Perspective ---
   // convert from world to clip by multiplying with projection scaling factor
+  // to get the right thickness (see https://github.com/processing/processing/issues/5182)
   // invert Y, projections in Processing invert Y
   vec2 perspScale = (projectionMatrix * vec4(1, -1, 0, 0)).xy;
 
--- a/core/src/processing/opengl/shaders/LineVert.glsl
+++ b/core/src/processing/opengl/shaders/LineVert.glsl
@@ -34,54 +34,61 @@ attribute vec4 color;
 attribute vec4 direction;
 
 varying vec4 vertColor;
-
-vec3 clipToWindow(vec4 clip, vec4 viewport) {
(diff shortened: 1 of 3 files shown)


benfry:
@codeanticode want to review before we merge?

benfry:
Oh look, Github now has a reviewer thing… Don't mind me, I'll just be returning to my hole called 2013.

codeanticode:
ha, no worries, I will take a look now!

codeanticode:
This code is excellent, much cleaner with a nice derivation of the tangent vector that factors out the problematic division by w, and without apparent edge cases. I will run a few tests and will report back, but don't see any side effects so far.

core/src/processing/opengl/shaders/LineVert.glsl
```diff
@@ -34,54 +34,64 @@ attribute vec4 color;
...
+  // to get the right thickness (see https://github.com/processing/processing/issues/5182)
+  // invert Y, projections in Processing invert Y
+  vec2 perspScale = (projectionMatrix * vec4(1, -1, 0, 0)).xy;
+
+  // No Perspective ---
+  // multiply by W (to cancel out division by W later in the pipeline) and
+  // convert from screen to clip (derived from clip to screen above)
+  vec2 noPerspScale = p.w / (0.5 * viewport.zw);
+
+  gl_Position.xy = p.xy + offset.xy * mix(noPerspScale, perspScale, perspective > 0);
```
codeanticode: @JakubValtar the glsl compiler on Mac does not automatically convert perspective > 0 into a number, so it gives an error bc does not find a mix(float, float, bool) function. This works:

mix(noPerspScale, perspScale, float(perspective > 0))

JakubValtar: Ah, the overload is in GLSL docs, but looks like only for GLSL 4.50. Try to leave there only `perspective` without anything, it is integer of 0 or 1 and should work on it's own.

JakubValtar: Or `float(perspective)` if the int does not work either.

JakubValtar:
@codeanticode I modified the shaders and checked that they run with `#version 110` which means that they should run on all cards supporting OpenGL 2.0 or newer. I also modified the version check in shader preprocessor to ignore comments (something I've ran into while testing and which we should have in place).

codeanticode:
@JakubValtar thanks for the last tweaks

@benfry this PR is ready to merge

JakubValtar:
@codeanticode Thanks for the review!

JakubValtar:
Found another one which I accidentally fixed.

Fixes #5194

benfry:
Thanks gentlemen…

JakubValtar:
*tips hat*

PR #5189, opened 2017-07-22
Reword another expecting RCURLY error

### LINKED ISSUE #4284: syntax check displays confusing message
have this (this comes from something slightly more elaborate) somewhere:
`public class something
{
  public something()
  {
  }
  
  String thisandthat()
  {
  
  }
    return "string";
}`

this will throw a "Found one too many { characters without a } to match it." message, which somehow just isn't quite right.

GKFX:
Fix #4284.

### DIFF
--- a/java/src/processing/mode/java/JavaBuild.java
+++ b/java/src/processing/mode/java/JavaBuild.java
@@ -301,10 +301,18 @@ public class JavaBuild {
           }
         }
         // If we're still here, there's the right brackets, just not in the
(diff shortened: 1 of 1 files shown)


benfry:
In the majority of cases, the original is correct. This would replace it with a message that doesn't provide a useful message for any scenario (it's a bit like just saying "syntax error").

Agree we can do better here, but I'd rather we replace with a new preprocessor where these things can be sorted out in a broader way.

PR #5201, opened 2017-07-31
Fixed type

libasoles:
Catch method had a typo for variable 'name'

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -1557,7 +1557,7 @@ public class PApplet implements PConstants {
       Method method = c.getMethod(name, new Class[] {});
       meth.add(o, method);
 
-    } catch (NoSuchMethodException nsme) {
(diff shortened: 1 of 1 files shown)


kritixilithos:
It does not appear to be a typo **N**o**S**uch**M**ethod**E**xception => `nsme`. Also the variable `name` is already declared as a parameter in both functions.

Here

[`private void registerNoArgs(String name, Object o) {`](https://github.com/libasoles/processing/blob/04ccf7fdad2c4bf39bba16c5929f3a94ea11e8cd/core/src/processing/core/PApplet.java#L1549)

and here

[`private void registerWithArgs(String name, Object o, Class<?> cargs[]) {`](https://github.com/libasoles/processing/blob/04ccf7fdad2c4bf39bba16c5929f3a94ea11e8cd/core/src/processing/core/PApplet.java#L1570)

benfry:
Yep, @kritixilithos is correct… not a typo (and this would introduce an error).

PR #5202, opened 2017-07-31
Add support for custom icon file for macOS builds

codingjoe:
Currently macOS builds only support custom `Info.plist` files.
It is possible to change the dock icon from there, but not the
bundle icon.
This commit allows putting your own `sketch.icns` file inside
your sketch folder, which overwrites the default.


codingjoe:
Hi @benfry what can I do, to get change up stream? Thx :)

benfry:
@codingjoe Figure out a way to get a couple extra hours of free time in my day? :)

We'll be pursuing a more complete solution here: https://github.com/processing/processing/issues/5225 which will cover this case but also some problems we've had with dock icons on macOS. 

Thanks for looking into it!

Issue #4896, opened 2017-02-12
Welcome screen doesn't size properly for HiDPI screens


benfry:
Even with fixes to the editor in Processing 3.3 (https://github.com/processing/processing/issues/2411, https://github.com/processing/processing/issues/4183), attempts to fix the Welcome screen didn't work out. So… that still needs to be fixed.

cammorales:
Hi all,
I recently downloaded processing to my surface book 2 and the sketch window is very small. Any suggestions on how to fix this? I looked through the threads but i'm not finding a solution. Any help would be greatly appreciated.

benfry:
Fixed for 4.0 alpha 5, hopefully available soon: https://github.com/processing/processing4/releases

PR #5210, opened 2017-08-09
Rewrote Welcome in Swing

### LINKED ISSUE #3911: un-checking "show this message" and closing the welcome window doesn't store the change
You must click "Get Started" after un-checking the box for the preference to take. It's a side effect of it being implemented as a web page. 

Might be time to give up on the web page implementation, since there doesn't seem to be a way around it.

### LINKED ISSUE #3912: click "show this welcome" text to check/uncheck the box
Rather than needing to click the box itself, make it work when users click the text as well. Like #3911, this is a problem with its current implementation as a web page, and something we don't control, unless we rewrite the Welcome window.

whetfire:
Changed the Welcome from a child of WebFrame to a child of JFrame.

This change fixes GUI problems #3911, #3912, and potentially #4896. (If it does not fix 4896, it should make it very easy to - the only changes needed are swapping out the logo image and resizing the window)
Closing the window will change the preferences for showing the welcome screen each time.
All text is now tied to the buttons and check boxes, making it more user friendly.
The Welcome screen should also resize to display well on high resolution displays, but I don't have a computer to test it on.

I also extracted the text out of the GUI. I couldn't find where the translations of the old Welcome screen are, but this will also make it easier to translate the welcome screen. Only a dozen phrases are needed, and the  file format will make it easy to replace the strings in the future.

Closes #3911 
Closes #3912

### DIFF
--- a/app/src/processing/app/ui/Welcome.java
+++ b/app/src/processing/app/ui/Welcome.java
@@ -60,11 +60,28 @@ import processing.app.Preferences;
 import processing.core.PApplet;
 import processing.data.StringDict;
 
+/**
+ * The Welcome class creates a welcome window upon startup
+ * 
+ * It provides links to changes Processing 3
+ *
+ * If the user is migrating from Processing 2, it provides a
+ * prompt asking whether to use the same sketchbook folder
+ * as before, or use a new one.
+ *
+ */
 
 public class Welcome extends JFrame {
   Base base;
-  StringDict dict; // used for submitting form values
+  StringDict dict;  // used for submitting form values
 
+  /**
+   * Constructor for the Welcome window
+   *
+   * @param Base the current Processing Base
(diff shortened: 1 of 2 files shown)


VictorGaiva:
It did not fix #3911. It is still pretty hard to fix this.

benfry:
Thanks for the attempt, though even after putting in some time with it (https://github.com/processing/processing/commit/5b672a1f2b20b79fee9c1a22ab565aabb504afca and https://github.com/processing/processing/commit/ddb81c8cc84871179173752a779767828ec065b8), it's visually still too messy for us to use. 

screen shot 2018-03-18 at 12 41 38 pm

This is why we did the HTML version in the first place—getting UIs to look good in Swing is just really difficult. 

But I've done some other fixes and think we're in good shape for the next release. Thanks for taking a look.

Issue #4266, opened 2016-02-03
Feature - Install Processing from debian repository


Akarshit:
It would be a nice feature if Processing could be directly installed from debian repositories or something similar using commands like `sudo apt-get install processing`, etc.
This would also help users to upgrade processing easily.

@shiffman Do you think this is a good feature to have ?

I would like to implement this but am having some problem as to where to start from. I will keep digging but any help would be highly appreciated.

shiffman:
I'm not enough of a linux user to know, but I can see how it would be useful.

aengelke:
There is support for building .deb packages using `ant deb`.

Akarshit:
Thanks @aengelke

benfry:
The feature is implemented (#114); it's just a matter of having a consistent maintainer.

marado:
A first step would be what I asked in #4230 : without the overhead of maintaining a repository and all of that, just `ant deb` each time there is a new version, and add the resulting .deb next to the other files in the downloads section.

benfry:
Same issue; needs a consistent maintainer. I'm not eager to add another step to the already long build process of releasing and testing on 5 different platforms (OS X, Windows 32 and 64, Linux 32 and 64).

gohai:
I just tested `ant deb` on ARM w/ PR #4300, and it worked well (thanks, @kfeuz)!

I want to work towards getting a .deb into one of these Debian spin-offs for small devices, since that's where my focus is.

benfry:
@gohai So maybe I could convince you to handle the ARM release and also the desktop .deb release? :smile_cat:

gohai:
*Look behind you, a three headed monkey!* \*runs\*

I'll think about it.. I'd do it in a heartbeat if I was running Debian on my own machine, but I unfortunately it's Fedora (.rpm) and OS X.

I agree that it'd be nice to have, but I don't think it's uncommon for projects to only publish tar-balls, plus packaging metadata in the sources, that can be used to automatically generate distribution packages *iff* anyone is sufficiently inclined. But what if the .deb would suddenly not work in some far-out Ubuntu derivative? And what about the distros that don't do .deb?

benfry:
I'm not gonna fall for the three-headed monkey trick again. 

FWIW, the difference for the .deb is that Ubuntu is our only "officially" supported Linux flavor, and it seems like .deb distributions (Ubuntu, Mint, Debian) are most prominent. But obviously, we've also been fine with the tarball to date.

[...]

PR #5228, opened 2017-08-25
Add Linux man pages

GKFX:
Ages ago I told you I'd produce and maintain an official Debian package of Processing (#4266). That didn't really happen, because Debian requires that all of Processing's dependencies would be packaged first, and the Debianized Eclipse dependencies were so out of date that they couldn't handle Java 8. So I sat on my nearly-finished code for a while, but I've decided to send it to you as feature-by-feature pull requests, since that way it's easy to review and it minimizes the work involved in eventually getting a package moving.

Attached are the man pages required by Debian policy, which I've included in the package currently produced by `ant deb`.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -845,6 +845,9 @@
     <copy file="linux/processing-pde.xml" todir="linux/work/lib" />
     <copy file="linux/appdata.xml" todir="linux/work/lib" />
     <copy file="linux/desktop.template" todir="linux/work/lib" />
+    <copy todir="linux/work/lib/man">
+      <fileset dir="linux/man" />
+    </copy>
 
 <!--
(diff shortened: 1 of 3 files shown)


kritixilithos:
It might be useful to specify that the paths for the arguments to `--sketch` and `--output` should be full paths and not relative paths.

GKFX:
@kritixilithos  I'd like to wait for #4912 to get resolved first before rewriting the docs. Also, the man page for processing-java is just `processing-java --help` re-formatted, so it'd be helpful to change both.

benfry:
What does this provide for users that they can't get via running the command to get the help page? I'm concerned about adding one more thing that has to be updated and maintained (with no maintainer).

benfry:
That's what I figured; thanks for checking in.

PR #5230, opened 2017-08-25
Unlock "Create Movie" in Movie Maker after use

### LINKED ISSUE #5168: Movie Maker only works once per session (subsequent uses require restart of PDE)
I recently updated to Processing 3.3.5 and now I can only run the Movie Maker tool once per session - i.e. if I launch processing and run the Movie Maker tool, I have to restart Processing before I use it again. I don't get any error messages, but the 'create movie...' button is greyed out, even though there is a path specified to the frames folder. I've checked a bunch of times and the path to the frames folder is correct. If i restart processing and use the exact same path, it'll work - but just once and then I have to repeat the whole process again. I've test this on the previous version of Processing and there is no issue with Movie Maker. 

macOS 10.12.5 Quicktime 10.4 frames are .tif

GKFX:
Fixes #5168.

### DIFF
--- a/build/shared/tools/MovieMaker/src/processing/app/tools/MovieMaker.java
+++ b/build/shared/tools/MovieMaker/src/processing/app/tools/MovieMaker.java
@@ -606,12 +606,14 @@ public class MovieMaker extends JFrame implements Tool {
         } catch (Exception ex) {
           t = ex;
         }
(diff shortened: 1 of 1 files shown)


benfry:
Great, thanks for tracking this down.

Issue #5052, opened 2017-05-05
resizing a sketch window to zero height throws exception


Fenestron:
The exception is thrown when the size is changed to the minimum width.

![608bf981-5ff1-4943-b7f5-c9e75f8cc2e1](https://cloud.githubusercontent.com/assets/9680621/25752707/bce5a176-31c1-11e7-90a3-68ebeea04580.GIF)

Stacktrace:

```
java.lang.IllegalArgumentException: Width (865) and height (0) cannot be <= 0
	at java.awt.image.DirectColorModel.createCompatibleWritableRaster(DirectColorModel.java:1016)
	at java.awt.GraphicsConfiguration.createCompatibleImage(GraphicsConfiguration.java:186)
	at processing.awt.PGraphicsJava2D.checkImage(PGraphicsJava2D.java:328)
	at processing.awt.PGraphicsJava2D.beginDraw(PGraphicsJava2D.java:338)
	at processing.core.PApplet.handleDraw(PApplet.java:2401)
	at processing.awt.PSurfaceAWT$12.callDraw(PSurfaceAWT.java:1547)
	at processing.core.PSurfaceNone$AnimationThread.run(PSurfaceNone.java:316)

Could not run the sketch (Target VM failed to initialize).
Make sure that you haven't set the maximum available memory too high.
For more information, read revisions.txt and Help ? Troubleshooting.
```

I'm using windows 10, Processing 3.3.3

benfry:
You've resized the window to have zero height… I guess someone could do some extra work to avoid an exception in this case, but I'm not really sure what you'd expect the sketch to do at that point—it doesn't have anywhere to draw.

Fenestron:
What can I do to avoid such an error?

jeremydouglass:
I can see that it isn't reasonable to expect Processing to handle a sketch of height 0 -- but I can also see how an accidental mouse drag could trigger this crash, making resizing a dangerous operation. 

Is there an easy way for a resizable sketch to have a min width / height defined, like 1x1 or 10x10? I haven't looked at the details, but I believe java.awt has `setMinimumSize()` 

https://docs.oracle.com/javase/7/docs/api/java/awt/Component.html#setMinimumSize(java.awt.Dimension)

benfry:
`setMinimumSize()` is already called to set the minimum to the sketch's size, which is obviously overridden by setting the window resizable. Someone would need to dig into how to set a minimum that's actually honored; since on some level it's left up to the OS.

[...]

PR #5234, opened 2017-08-27
A fix for the window resizing problem in Windows

kisarur:
The detailed issue can be found here -> https://github.com/processing/processing/issues/5052. A fix for that is applied by checking if the height being set is zero and if so, resetting it to 1(to make it greater than zero).


benfry:
Thanks for taking the time to look into it. Now incorporated for 3.3.6.

PR #5235, opened 2017-08-29
Make the TextTransform constructor public.

katiewalker:
I'm hoping that leaving the constructor package-private was an oversight. If not: it would be very helpful for it to be accessible. I'm creating a mode that extends JavaMode, and it would be useful to be able to use the TextTransform pipeline to modify code.

Similarly for the Edit class, which looks like it was intentionally left package-private.


benfry:
In general, we try to keep things package-private (or otherwise non-public) unless it's an API we know we want people using, and are committed to supporting/not changing it in future releases. How are you planning to use this class?

katiewalker:
Specifically, I'd like to insert import and other statements into the code that is ran, without any action on the user's part (aside from using my mode).

As to why I want to do this: in order to obtain a Bachelor of Computer Science (Honours), I'm developing a mode that will allow students with visual impairments to use Processing. We're working on a device that will allow the user to feel their sketch. A library provides the code that translates pixel values into commands for the device. We'd use text transform to import and instantiate the library. We're trying to avoid having extra tasks our users must complete, compared to the standard Processing workflow.

JakubValtar:
Hi @katiewalker!

Firstly I think we should not make the classes public, because I might need to change them soon and libraries and modes which would depend on it would break.

To address your problem: I don't think inserting code manually into sketches is the right solution for you. Preprocessing is very messy and error prone and JavaMode is not made to be easily extended in that regard so you will have to fight it (you will need to modify the build process which currently runs on ANTLR and also PreprocessingService which takes care of live error checking).

I think for your use case the best solution would be to use a PApplet subclass (Python Mode does this). 
- when the sketch is run in your mode it will do extra stuff behind the scenes
- you will be able to open any .pde sketch in your mode and enhance it with your library
- sketches created in your mode could be opened and ran in plain JavaMode for other people to use

To get you started:
- Create a PApplet class in the package of your mode, code below
- Create a subclass of PdePreprocessor and override `getCoreImports()` and add your PApplet to the array of core imports. Your import like will take precedence over the wilcard `processing.core.*` and sketches will use your `mymode.PApplet` instead of `processing.code.PApplet`.
- Create a subclass of JavaEditor and override `createPreprocessor()` to return your preprocessor subclass.
- In your Mode class override `createEditor()` and return your subclass of JavaEditor.

```java
package mymode;

// Keep the name PApplet, it is inserted automatically into the sketch code and it's hard to change
public class PApplet extends processing.core.PApplet {

  // keep this private to hide it from users in sketchbook
  // make it protected or create a getter if you want users to access the library too,
  //   but sketches which would access the lib wouldn't compile in plain JavaMode
  //   (the field does not exist in regular PApplet)
  private MyLibrary myLib;
  
  @Override
  public void handleDraw() {
    int prevFrameCount = frameCount;
    super.handleDraw();
    // this is not pretty but it will get the job done
    if (myLib == null && prevFrameCount == 0 && frameCount == 1) {
      // here you can instantiate your library
      // your library can hook up into sketch events (pre, post, draw, input events)
      // see https://github.com/processing/processing/wiki/Library-Basics#library-methods
      myLib = new MyLibrary(this);
    }
  }

}
```

katiewalker:
On 31 August 2017 at 22:14, Jakub Valtar <notifications@github.com> wrote:

> Hi @katiewalker <https://github.com/katiewalker>!
>
Hi! Thanks for taking the time to help me out.

> Firstly I think we should not make the classes public, because I might
> need to change them soon and libraries and modes which would depend on it
> would break.
>
Once the changes are made, would you be willing to open them up to the API?
If this is the case, is there anyway I can assist to speed up the changes?
IMO, there are many use cases for an edits framework for Processing code.
For example, I was surprised that AutoFormat does not make use of it.

> To address your problem: I don't think inserting code manually into
> sketches is the right solution for you. Preprocessing is very messy and
> error prone and JavaMode is not made to be easily extended in that regard
> so you will have to fight it (you will need to modify the build process
> which currently runs on ANTLR and also PreprocessingService which takes
> care of live error checking).
>
I'm not clear on why I'd have to modify the build process (although I agree
about having to fight JavaMode!). In my local copy of Processing, which
includes these changes, I can use TextTransform to intercept and modify the
code (see here
<https://github.com/katiewalker/AudioTactileMode/blob/master/src/audiotactilemode/AudioTactileMode.java#L48>).
It behaves as necessary. The particular library class in use there (which
is in a separate repository
<https://github.com/katiewalker/tactile-graphics-processing/blob/master/src/tactilegraphics/concept/ReadPixelsFromSketch.java>)
is able to read and print the value of each pixel in the sketch. This is
all the final library will do in the way of interacting with Processing.

> I think for your use case the best solution would be to use a PApplet
> subclass (Python Mode does this).
>
>    - when the sketch is run in your mode it will do extra stuff behind
>    the scenes
>    - you will be able to open any .pde sketch in your mode and enhance it
>    with your library
>    - sketches created in your mode could be opened and ran in plain
>    JavaMode for other people to use
>
> All of the above are true of my current implementation. In order to
minimise the maintenance costs of my project, I've tried to avoid extending
components of JavaMode as much as possible. I think my current solution is
cleaner, in this way.

With the big picture in mind, like I said above, an edits framework would
be beneficial, outside of my use case. Even if TextTransform is not made
public, it is trivial to directly manipulate the text before the sketch is
ran. The benefit of TextTransform is that  the finer details of the
Processing language, and awfulness of string manipulation, can be
abstracted away.

JakubValtar:
> Once the changes are made, would you be willing to open them up to the API? If this is the case, is there anyway I can assist to speed up the changes? IMO, there are many use cases for an edits framework for Processing code. For example, I was surprised that AutoFormat does not make use of it.

Maybe there is a misunderstanding, but TextTransform is not exactly an edits framework (although it can do some transformation). It is primarily a tool to record mapping from input text to output text during preprocessing in a form of OffsetMapper. It expects a relatively low number of changes which don't overlap. If you don't need to map positions between input and output then TextTransform is a bit of overkill. SourceUtils are meant to be used at a specific place during the preprocessing. `insertImports` adds import statements at the beginning, basically `"import my.class;\n" + code`. So depending on when you call it it might do a right thing or a wrong thing. There is no clever manipulation happening.

If we make it public, it will probably benefit only people developing Modes or working on JavaMode. Once somebody starts using it, we are stuck with it and can't change it. That makes modifications in JavaMode much more complex than it already is (it is complex).

In a nutshell, making something public is a commitment, we have to document it, maintain it, and we have a responsibility towards users not to break their code with changes. Right now we can't make this commintment, but we will consider it in the future.

> I was surprised that AutoFormat does not make use of it.

The reason is there was no TextTransform when AutoFormat was written. Besides that, it does a different job than what AutoFormat needs to do, and has worse performance if you want to use it for this task anyway.

> I can use TextTransform to intercept and modify the code (see here

You don't need TextTransform for this. Your code does excatly the same thing as:
```java
  private String modifyCode(String code) {
    return "import tactilegraphics.concept.ReadPixelsFromSketch;\n" +
        "ReadPixelsFromSketch writer = new ReadPixelsFromSketch(this);\n" +
        code;
  }
```
just in a slower and more complicated way.

> I'm not clear on why I'd have to modify the build process (although I agree about having to fight JavaMode!).

Although it seems easy to intercept and modify the code like you did it opens another can of worms. In your code you added two lines of code, which means that:
- errors encountered during the build process will be reported on a wrong line
- exceptions thrown when sketch runs will be reported on a wrong line
- debug mode will probably break
- tweak mode will probably break
- if somebody names something "writer" in their code it won't run (and show error on a wrong line)
- if you save the sketch it might include the lines you added, so next time you open it it will be there twice
- setting sketch program to a different string than what is in the editor might break the editor at some places

You see, manipulating sketch code will break things. That's why using a PApplet subclass is a better idea which is robust and easier to maintain.

If you really want to go the way of string manipulation (which I strongly discourage) you should do it in the Preprocessor and manipulate the headerOffset to keep the line mapping working. This should work:

```java
public class MyPreprocessor extends PdePreprocessor {

  // add constructors to match super constructors

  public PreprocessorResult write(Writer out, String program,
                                  StringList codeFolderPackages)
      throws SketchException, RecognitionException, TokenStreamException {

    // create the lib in a block which can run anywhere and doesn't create "writer" identifier
    program = "import tactilegraphics.concept.ReadPixelsFromSketch;\n" +
        "{ new ReadPixelsFromSketch(this); }\n" +
        program;

    PreprocessorResult r = super.write(out, program, codeFolderPackages);

    // add 2 to the header offset for the lines we added above
    return new PreprocessorResult(r.programType, r.headerOffset+2, r.className, r.extraImports);
}
```

Hopefully this helps, let us know if you have problems. Good luck with your project!

benfry:
@katiewalker Does this all make sense? Are we safe to close this PR?

benfry:
No response; closing.

PR #5251, opened 2017-09-08
BezierPoint  

Prince-Polka:
Changing  
 
```
float bezierPoint (float a, float b, float c, float d, float t) {
  float t1 = t-1.0f;
  return t * ( 3*t1*(b*t1-c*t) + d*t*t ) - a*t1*t1*t1;
}

```
to  
```
public float bezierPoint(float a, float b, float c, float d, float t) {
  float t1 = 1.0f - t;
  return a*t1*t1*t1 + 3*b*t*t1*t1 + 3*c*t*t*t1 + d*t*t*t;
}
```

Gives the same result faster, less intuitive code though

### DIFF
--- a/core/src/processing/core/PGraphics.java
+++ b/core/src/processing/core/PGraphics.java
@@ -3211,7 +3211,7 @@ public class PGraphics extends PImage implements PConstants {
    * @see PGraphics#bezierVertex(float, float, float, float, float, float)
    * @see PGraphics#curvePoint(float, float, float, float, float)
    */
-  float bezierPoint (float a, float b, float c, float d, float t) {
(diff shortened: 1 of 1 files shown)


JakubValtar:
Hi @Prince-Polka, thanks for your PR!

Hm, it's true that it is not very intuitive, but I think we can address that.

Would you agree to reshuffle the code this way?

```java
public float bezierPoint(float a, float b, float c, float d, float t) {
  float t1 = 1.0f - t;
  return (a*t1 + 3*b*t)*t1*t1 + (3*c*t1 + d*t)*t*t;
}
```

A bit more readable while keeping the operation count the same.

Prince-Polka:
@JakubValtar that sounds good

PR #5265, opened 2017-09-20
Fix scrub comments for empty block comment /**/

### LINKED ISSUE #5219: /**/ not handled properly by the preprocessor/parser
If `/**/` appears somewhere in the code, something in the error generating code will treat it like it's only a comment opening, and ignore the closing, although the syntax highlighter and compiler treat it like a closed comment and it runs without any problems.

Example code which gives false errors:
```
void setup() {
  /**/
  print("Hello, World!");
}
```

JakubValtar:
Fixes #5219


benfry:
Thanks!

PR #5266, opened 2017-09-20
Make menu separators visible on linux with the GTK look and feel

crazymaster:
Backport from Arduino
https://github.com/arduino/Arduino/commit/f183966296458bc9e157b3d05cadb54aaac090aa

BTW, are there any other problems to set the GTK look & feel by default on Linux?

### DIFF
--- a/app/src/processing/app/platform/GTKLookAndFeelFixer.java
+++ b/app/src/processing/app/platform/GTKLookAndFeelFixer.java
@@ -1,0 +1,151 @@
+/*
+ * This file is part of Arduino.
+ *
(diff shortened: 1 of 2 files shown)


benfry:
We moved away from the GTK Look and Feel because Linux users were complaining about it.

benfry:
In 3.0a6, the look and feel set for Linux in `defaults.txt` (and therefore `preferences.txt`) is Nimbus, because GTK had so many problems like this. Removing `editor.laf` from your `preferences.txt` will reset it for you. 

Or this could be turned into a contributed Tool for folks who want to use GTK instead.

PR #5282, opened 2017-10-15
Remove "Pipe Organ" from exec javadoc

jeremydouglass:
Recent OS X releases no longer come with the "Pipe Organ" voice installed by default. As a result, the exec example fails silently with no console feedback. Removing the -v voice argument causes the example to work on current OS X -- and still work on older OS X.

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -3457,7 +3457,7 @@ public class PApplet implements PConstants {
    * method, which uses the operating system's launcher to open the files.
    * It's always a good idea to use a full path to the executable here.
    * <pre>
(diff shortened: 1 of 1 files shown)


jeremydouglass:
Reference supporting Apple's "voice removal":

https://www.reddit.com/r/MacOS/comments/5708ul/when_did_apple_remove_the_voices_like_good_news/

benfry:
How am I supposed to entertain myself if Apple insists on taking the joy out of esoteric examples? 

Sounds like a better example would be using `exec()` to manually install the Pipe Organ voice, and *then* play the sound directly. That'll show 'em.

benfry:
<sniff>

PR #5283, opened 2017-10-17
Added HSLuv colorspace/colorMode

j-huff:
More info: http://www.hsluv.org/
Used java code for color conversion from https://github.com/hsluv/hsluv-java


benfry:
Thanks, though we can't incorporate code with someone else's copyright. It'd need to be transferred to the Foundation and made available under the same license as the rest of our code. 

You should definitely try making it a library, so that you can release it under the same license as the original code.

PR #5297, opened 2017-10-30
ARM updates

gohai:
We need to use a special shader to workaround register allocation issues with the new driver, see: https://github.com/anholt/mesa/issues/63


gohai:
@benfry Can I merge this PR? Fixes P3D with recent open source graphics drivers on the Pi, and shouldn't break anything else

benfry:
Thanks, sorry for the delay.

PR #5313, opened 2017-11-20
Refactoring Commander.java

christianbender:
I have add an enumeration type 'Task' for describing the tasks. (constructor method)
In addition, I have cleaned up the code a little bit.


benfry:
Thanks, though this seems to do the same thing with more detailed code, and the commented-out portions are mostly things that may be brought back in the future.

PR #5333, opened 2017-12-15
Visibility of methods 

simoncev:
Needs for integration with java internal application


benfry:
Sorry, these aren't methods that we're looking to make public, or rather, deal with the implications of them being public (and developers modifying the event queue) as a broadly available thing.

Issue #5068, opened 2017-05-17
loadTable() should support a separator specification option


tbarnier:
it is impossible to import a csv file where fields are separated by a non standard character (like ';', or '|').

a 'sep' option allowing to specify a field delimiter character would help usability, and be in line with other framework like in R, or in python pandas read_csv()

jeremydouglass:
To support non-standard delimiters (that don't also appear in field data), preprocess the file with search-replace. For previous discussion and this workaround, see: #3039

benfry:
I'll mark this as an enhancement. My hope was to keep things simple by not supporting every CSV variant, but this is coming up frequently enough that the current limitation seems to undermine the usefulness of the `Table` class. 

Also note that OpenOffice `.ods` files also work with `loadTable()`, which provide a more robust format.

Help requested for the delimiter implementation…

TheGuywithTheHat:
I've looked over the code, and I'll try to get this done within a few days. Should I add the customization to saveTable() as well?

benfry:
Yeah… The tricky thing here is how alternate delimiters handle escapes. So if you're using | or ; we need to make sure the escaped versions of the same work properly (and make sure we're aware of the correct escaping systems).

christianbender:
How about a simple class for desired functionality?

christianbender:
I build a CSV class (in Processing itself) to support the requests.

TheGuywithTheHat:
@christianbender I got started on this, but ended up too busy to finish. You can see what changes I made at https://github.com/TheGuywithTheHat/processing/commits/load-table. Feel free to pull from my repo (load-table branch) and build off of what I have, or just use my code as a guideline.

I think what I was stuck on last summer was how to specify the delimiter:

 - if it's part of the options string, we would have to either disallow certain delimiters or implement our own escape sequence parser.
 - if it's a parameter of its own, we would have to somehow handle cases where the user doesn't want to input an options string.

benfry:
Thanks for taking a look at it (and sounds like you're starting to understand the trickiness I was referring to…)

christianbender:
@TheGuywithTheHat  Thanks for response. I see it at the moment.

PR #5334, opened 2017-12-15
CSVFile class Issue #5068

christianbender:
This class implements a data structure for handling of CSV-files.
For issue #5068 

### Overview about methods:

* CSVFile (String path, char seperator)
* CSVFile ( File file, char sperator)
* CSVFIle (char seperator)
* compile (String row, char seperator)
* isPunctuation (ch : char) : boolean
* getElementString(row : int, column : int) : string
* getElementDouble(row : int, column : int) : double
* addRow(row : string) : void
* commit() : void 
* commit(path : String) : void
    * commit(file : File ) : void
* findRow(key : string) : ArrayList<String>
* contains(key : string) : boolean
* getColumn(column : int) : ArrayList<String>
    * getColumn(key : string) : ArrayList<String>
* removeRow(key : string) : void 
    * removeRow(column : int) : void
* updateFile() : void
   * updateFile(file : File) : void
* getNumberOfRows() : int


christianbender:
@JakubValtar  @all 

Why no response on the many pull request? Some PRs a addressed to issues with label "help wanted".

benfry:
Hi Christian, we're working in our spare time and are often unable to address PRs, especially if they're not ready to go.

We can't accept this PR as-is. Rather than implementing delimiter support, it's a complete rewrite of the `Table` code that doesn't follow any of the conventions of the rest of the Processing source, adds a new class, and the implementation is significantly slower and less memory efficient than what's currently in the `Table` class. I know that code is complex, but it's that way in order to be more performant and also use memory in a way that allows us to load very large data sets.

christianbender:
@benfry  
Ok, I understand. I will tried extended the Table class.
Thanks for your feedback.

benfry:
Appreciate your understanding, and thank you for the help.

PR #5340, opened 2017-12-20
cursor() sets the cursor visible, fixes #5330

### LINKED ISSUE #5330: void cursor() don't work after void noCursor() in P2D and P3D
## Description

void cursor() don't work after void noCursor() in P2D and P3D

When the method `noCursor()` is used in P2D or P3D, it's not possible to change the statement of cursor, that's stay in `noCursor()` mode. May be the bug have a same root  https://github.com/processing/processing/issues/4687

## Expected Behavior

switch between different method `cursor()`

that's work, in all mode JAVA2D, P2D, P3D... :)

```
void setup() {
  size(400,400,P2D); // don't work
  // size(400,400,P3D); // don't work
  // size(400,400,JAVA2D); // ok
  // size(400,400); // ok
}

void draw() {
  background(0);
  if(display) {
    println("display cursor");
    cursor(CROSS);
  } else {
    println("hide cursor");
    noCursor() ;
  }
}

boolean display = true;
void keyPressed() {
  if(display) display = false ; else display = true ;
}
```

## Your Environment
* Processing version:3.3.6
* Operating System and OS version: OS Sierra
* Other information: MacOS 2012

alexanderghurst:


codeanticode:
@benfry It's good @alexanderghurst thanks for this PR!

benfry:
Great, thanks folks.

Issue #5271, opened 2017-09-26
Allow editor errors and status messages to be selected/copied


sdaau:
## Description

If you have a Java compile error in Processing IDE, it may show up in a red "top bar" as on this image:

![scrproc3](https://i.stack.imgur.com/Ym96q.png)

In such a case, the error message (the `Error on "VariableDeclarators"` in the case shown on the image) is not copy-pastable (hovering the mouse over the red top bar and clicking Ctrl-C does nothing). This can end up being a problem, as the only obvious way to copy/paste that message is to actually re-type it, which gets complicated for more verbose errors (for example, "`The type com.google.api.client.googleapis.services.json.AbstractGoogleJsonClient$Builder cannot be resolved. It is indirectly referenced from required .class files`")

## Expected Behavior

It should be possible to copy/paste the red top bar error message directly from the Processing IDE GUI.

## Current Behavior

Currently, the error message in the red top bar (if it appears) in Processing IDE is not copy/pastable.

However, if the sketch is ran from the command line; say for a sketch in `~/sketchbook/proctest/proctest.pde`:

    $ mkdir ~/sketchbook/proctest
    $ echo "blahblaha" > ~/sketchbook/proctest/proctest.pde
    $ ./processing-java --sketch=`echo ~/sketchbook/proctest` --run
    proctest.pde:1:2:1:2: unexpected token: blahblaha

... we will get some error messages, though they are not exactly the same as the ones shown in the IDE (here `unexpected token: blahblaha`, vs. the red top bar `Error on "VariableDeclarators"`)

## Steps to Reproduce

1. Open a new sketch in Processing IDE
2. Type a random string which is not valid Java
3. Wait for the string to be detected as error, so it is underlined and an error shows in the red top bar
4. Try to copy/paste the error message in the red top bar from the Proccessing IDE - you will not be able to

## Your Environment

 

* Processing version: Processing 3.3.6
* Operating System and OS version: Ubuntu 14.04.5 LTS
* Other information: [none]

## Possible Causes / Solutions

Possible (UI) solutions - either:

* Draw an indication (i.e. an outline in different color) when the top bar is "selected" or "focused", and make Ctrl-C copy the error message text to clipboard; or:
* Have a button in the red top bar saying "Copy" or "Copy errors", and have the the error message text copied to clipboard upon clicking of this button

benfry:
Good idea… The status bar dates from before StackOverflow and Google existed, so copy/paste wasn't really a priority.

[...]

PR #5345, opened 2017-12-23
Issue #5271

anuran-Chakraborty:
There was a problem copying the errors. However with this commit on clicking the status bar the error will get copied to clipboard.


benfry:
Thanks for taking a look. Wound up going with a more comprehensive solution for Processing 3.4.

PR #5378, opened 2018-01-22
Fixes for Net library

### LINKED ISSUE #3970: clientEvent / serverEvent
It would be useful to be able to process the incoming messages outside the draw loop in both "client" and "server" applications. 
This already works "client side" with the clientEvent() method: the client can parse messages coming from the server in this event-like method.
The server on the other hand doesn't have a similar method: the only way to process incoming messages  from the connected clients is trough the draw loop and Client.available()…  
It would maybe make more sense to have a new serverEvent(Server s, Client c) method which gets fired when a client writes to the server. For clarity the actual serverEvent(Server s, Client c) method could eventually be renamed to connectEvent() as this is the only moment where the method gets called (cfr. disconnectEvent()).

Or 

Both server and client could have a mirror function like messageEvent(Server, Client) fired when a client or a server send a message with write().

### LINKED ISSUE #4419: Error in Examples -> Network -> SharedCanvasServer and Client  (Mac OS X)
I am trying to run the code of the SharedCanvasServer in Mac OS X and I am getting the error
ArrayIndexOutOfBoundsException

I have the 2.07 version but after download the most recent version I am still getting the same error.
I have tried in Windows 8.1 and the example is running without problems.
Thanks

### LINKED ISSUE #5360: processing.net.client not working properly after a few hours
## Description
I have a use case, where i read binary scanner data from the network using the processing.net.client library. After i few hours it throws is a memory overflow error.
 
The reason is here:
https://github.com/processing/processing/blob/0e9174a94e277c354851c798da12c5a7843f4f31/java/libraries/net/src/processing/net/Client.java#L234

Every usage of the library without completely clearing the buffer will fail because of an buffer overflow.

## Possible Solutions
Unfortunately i am not able to improve this code sequence.
It would be very helpful to have an note in the docs ("Clear the buffer regularly using client.clear();"), because this behaviour is very hard to understand (like all bugs which occur only every few hours).

JakubValtar:
Fixes
- Fix synchronization (probably fixes #4419)
- Compact client buffer before making it bigger (fixes #5360)
- Set limit to the size of client buffer (128 MB - might be too much)

Performance, style
- Read blocks of data instead of byte by byte
- Remove sleep in the server loop (probably there to 'fix' some threading issues)
- Code cleanup (creating varargs arrays manually, missing Override annotations)

Features
- Server now calls `clientEvent(Client c)` if it receives data from any client (fixes #3970)
- Unwrap exceptions thrown from the user code before printing them to the console

### DIFF
--- a/java/libraries/net/src/processing/net/Client.java
+++ b/java/libraries/net/src/processing/net/Client.java
@@ -95,16 +95,14 @@ public class Client implements Runnable {
       // which would be called each time an event comes in
       try {
         clientEventMethod =
-          parent.getClass().getMethod("clientEvent",
-                                      new Class[] { Client.class });
+          parent.getClass().getMethod("clientEvent", Client.class);
       } catch (Exception e) {
         // no such method, or an error.. which is fine, just ignore
       }
       // do the same for disconnectEvent(Client c);
       try {
         disconnectEventMethod =
-          parent.getClass().getMethod("disconnectEvent",
-                                      new Class[] { Client.class });
+          parent.getClass().getMethod("disconnectEvent", Client.class);
       } catch (Exception e) {
         // no such method, or an error.. which is fine, just ignore
       }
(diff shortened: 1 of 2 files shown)


benfry:
Many thanks for all these long-overdue updates.

JakubValtar:
The pleasure is mine.

Issue #5214, opened 2017-08-11
nullPointer on close button with P3D and noLoop


hype:
```
void setup() {
 size(600,600,P3D);
 noLoop();
}

void draw() { 
}
```

OSX Sierra 10.12.6 / Processing 3.3.5

if you press the esc key = no problem
if you mouse click "x" close button = 
```
java.lang.NullPointerException
	at processing.core.PApplet.dispose(PApplet.java:3706)
	at processing.core.PApplet.exit(PApplet.java:3653)
	at processing.opengl.PSurfaceJOGL$NEWTWindowListener.windowDestroyNotify(PSurfaceJOGL.java:911)
	at jogamp.newt.WindowImpl.consumeWindowEvent(WindowImpl.java:4389)
	at jogamp.newt.WindowImpl.sendWindowEvent(WindowImpl.java:4317)
	at jogamp.newt.WindowImpl.windowDestroyNotify(WindowImpl.java:4677)
```
ONLY occurs when `noLoop()` is present
ONLY occurs in P3D

christianbender:
I don't use OSX Sierra 10.12.6 but I use Linux Mint 18.2 and Processing 3.3.6.
If I close the window with 'ESC' or click of 'X'.  I get the message, below.
![screenshot1](https://user-images.githubusercontent.com/23243382/33137780-2acf056c-cfa9-11e7-9f91-676e81ca9513.png)

gwygonik:
I have the same issue on OSX (Sierra 10.12.6) with Processing **3.3.6** but with _either_ P2D or P3D.

> java.lang.NullPointerException
> 	at processing.core.PApplet.dispose(PApplet.java:3718)
> 	at processing.core.PApplet.exit(PApplet.java:3665)
> 	at processing.opengl.PSurfaceJOGL$NEWTWindowListener.windowDestroyNotify(PSurfaceJOGL.java:921)
> 	at jogamp.newt.WindowImpl.consumeWindowEvent(WindowImpl.java:4389)
> 	at jogamp.newt.WindowImpl.sendWindowEvent(WindowImpl.java:4317)
> 	at jogamp.newt.WindowImpl.windowDestroyNotify(WindowImpl.java:4677)
>

benfry:
Fixed by @hazmatsuitor for 3.3.7 or 3.4 (whatever we call the next release. 

@christianbender That looks like a different issue that can be filed separately. Thanks!

PR #5384, opened 2018-01-27
fix NPE when closing sketch window on Mac (#5214)

hazmatsuitor:
This fixes the first error mentioned in the issue, but not the other one mentioned by @christianbender. Tested on OSX High Sierra (10.13.2)


benfry:
Thanks for tracking it down!

PR #5389, opened 2018-02-02
Fix Net library buffer sizes

JakubValtar:
Because I don't know how to count...

### DIFF
--- a/java/libraries/net/src/processing/net/Client.java
+++ b/java/libraries/net/src/processing/net/Client.java
@@ -46,7 +46,7 @@ import java.net.*;
  */
 public class Client implements Runnable {
 
-  protected static final int MAX_BUFFER_SIZE = 2 << 27; // 128 MB
+  protected static final int MAX_BUFFER_SIZE = 1 << 27; // 128 MB
 
   PApplet parent;
   Method clientEventMethod;
@@ -222,7 +222,7 @@ public class Client implements Runnable {
     byte[] readBuffer;
     { // make the read buffer same size as socket receive buffer so that
       // we don't waste cycles calling listeners when there is more data waiting
-      int readBufferSize = 2 << 16; // 64 KB (default socket receive buffer size)
+      int readBufferSize = 1 << 16; // 64 KB (default socket receive buffer size)
       try {
         readBufferSize = socket.getReceiveBufferSize();
       } catch (SocketException ignore) { }


benfry:
Thanks for the catch!

JakubValtar:
What do you think would be a reasonable max size for the buffer? If it's too much it might take a while before the buffer fills up and the user is notified about a problem. Right now it is 128 MB per client, which is maybe too much.

I'm thinking about scenarios where small buffer size would be a problem and the buffer would fill up:
- user code is not reading anything or it's leaving leftovers (their mistake and they should be notified)
```processing
// right way to do it
void clientEvent(Client c) {
  while (c.available() > 0) { // read fully
    int b = c.read();
    // do your thing
  }
}
```
- they are using `readBytesUntil()` or `readStringUntil()` and there is too much data without the specified character - how much data is reasonable to send this way? linear search is performed every call...
- they are using `readBytesUntil()` or `readStringUntil()` and multiple sequences come in a single callback... we should maybe mention that safe way to use them is like this:
```processing
// right way to do it
void clientEvent(Client c) {
  String msg;
  while ((msg = c.readStringUntil('\n')) != null) {
    // do your thing
  }
}
```
- if the user polls the `available()` for new bytes and the buffer fills up between the calls... how much data could come between frames?

benfry:
Do we need to set a limit, strictly speaking? Why not just double the array when more is needed? It might cause the buffer to blow up (with an `OutOfMemoryError`) if the code is bad, but that seems less problematic than setting an arbitrary bound.

JakubValtar:
We don't need to, it was without a limit until now. However, if the user is doing something wrong, they will get `OOME` aka `my installation slows down after several hours and later crashes and I don't know why`, which is not a terribly useful error.

By limiting the buffer size, we can print a more useful error message sooner and disconnect, letting the system recover by reconnecting.

Maybe it would be better to print a message when some watermark is reached, but let the buffer grow?

benfry:
Yeah, I think printing a message for each allocation at 1MB and higher is probably appropriate. It won't eliminate all of the "my installation slows down" posts from people who ignore the messages, but I think it's a better balance of teaching people how to fix their issue rather than just providing as much rope as possible to hang themselves (continuous resizing) or an arbitrary cutoff that will break in some odd but otherwise legitimate edge case.

PR #5418, opened 2018-02-20
Updated comment

TobiPristupin:
Comment said function throws RunnerException, but function throws SketchException which isn't related to RunnerException


benfry:
Serves me right for putting any documentation comments in at all. ;-)

PR #5451, opened 2018-03-25
Added Russian Translation

AKonia:
I've added PDE_ru.properties and modified languages.txt. Is there anything else that i should do for integration of russian translation into Processing ?


benfry:
Awesome; thank you! That should do the trick.

PR #5452, opened 2018-03-26
made `context` public in `PGraphicsFX2D`

liquid600pgm:
`g2` is public in `PGraphicsJava2D`, and `context` is its FX2D equivalent.
Change made for consistency.


benfry:
That's what `getNative()` is for. (`g2` is only available for backwards compatibility)

PR #5454, opened 2018-03-27
implemented clip() and noClip() for FX2D

liquid600pgm:
it works, but it is not perfect. the problems with this solution:

 - after calling noClip() the settings like the fill and stroke color will reset to the ones set when calling clip()
 - it can cause major visual glitches and even crash the sketch if noClip() isn't called at the end of the frame

the second issue can be fixed by calling noClip() at the end of the frame automatically, but I couldn't find the code for that


liquid600pgm:
very poorly implemented, going to re-do this and re-submit the PR

PR #5475, opened 2018-04-18
ARM Mali upbringing

gohai:
This makes P3D work on Linux SBCs using ARM Mali graphics and their (closed-source) OpenGL ES 3.1 driver.

I don't have matching hardware, so all testing was done by user seongwook in the Processing Forums, who was very generous with his time in this effort. He tested with an ODROID-XU4 board, which reports a "Mali-T628" model and the following driver string "OpenGL ES 3.1 v1.r17p0-01rel0.a881d28363cdb20f0017ed13c980967e".

Re review: I specifically don't know if my code in d351070d30afe4861e04ec04c55af4a69770570f makes sense in this way.


benfry:
Thanks!

PR #5485, opened 2018-04-24
ARM Mali: Don't use GL_MULTISAMPLE or GL_POLYGON_SMOOTH on ES

gohai:
@codeanticode @JakubValtar This patch should not be controversial really, but in doing so I came across something that looked a bit strange to me: if you search the tree for `POLYGON_SMOOTH`, you'll find a hit in `restoreGL` and one in `setGLSettings` - but one of them is done unconditionally, and one just when `1 <= smooth` ... bug?


JakubValtar:
Good catch. As far as I know we don't want `GL_POLYGON_SMOOTH`, so `restoreGL()` should be modified to match `setGLSettings()`.

gohai:
Thanks @JakubValtar - added to PR.

Issue #5523, opened 2018-06-03
weirdly displaced vertices in SVG


liquid600pgm:
## Description

I am trying to draw [this file](https://gist.github.com/liquid600pgm/305ab9f0f4d003bb32f552b8689d0ed4) onto the screen, but the vertices are weirdly displaced.

## Expected Behavior

The vertices should be drawn correctly, as seen in the Gist.

## Current Behavior

Look at the shape on the right side. This happens in all circumstances, no matter if I scale it, translate it, apply a custom fill, it just always happens.
![Image](https://i.imgur.com/DROvmQB.png)

P2D and P3D crash when trying to draw the SVG, with a tesselation error "out of memory".
FX2D doesn't handle the SVG at all, and it looks like garbage:
![FX2D](https://i.imgur.com/kQsCP4N.png)

I have also tried drawing [this SVG](https://gist.github.com/liquid600pgm/2f93eb7970bf593c63196f89bea7be01), and it just crashes my sketch.

### Update
I tried to split the shape in half, and the sketch just crashes with the following stack trace:
```
java.lang.ArrayIndexOutOfBoundsException: 1
	at processing.core.PShapeSVG.parsePoly(PShapeSVG.java:461)
	at processing.core.PShapeSVG.parseChild(PShapeSVG.java:335)
	at processing.core.PShapeSVG.parseChildren(PShapeSVG.java:289)
	at processing.core.PShapeSVG.<init>(PShapeSVG.java:209)
	at processing.core.PShapeSVG.<init>(PShapeSVG.java:116)
	at processing.awt.PShapeJava2D.<init>(PShapeJava2D.java:52)
	at processing.awt.PGraphicsJava2D.loadShape(PGraphicsJava2D.java:1839)
	at processing.core.PGraphics.loadShape(PGraphics.java:1744)
	at processing.core.PApplet.loadShape(PApplet.java:11526)
	at iliquid.excircuit.IconBank.load(IconBank.java:35)
	at iliquid.excircuit.IconBank.load(IconBank.java:59)
	at iliquid.excircuit.Loader.load_icons(Loader.java:33)
	at iliquid.excircuit.exCircuit.setup(exCircuit.java:185)
	at processing.core.PApplet.handleDraw(PApplet.java:2401)
	at processing.awt.PSurfaceAWT$12.callDraw(PSurfaceAWT.java:1557)
	at processing.core.PSurfaceNone$AnimationThread.run(PSurfaceNone.java:313)
```
[...]

PR #5526, opened 2018-06-05
Fixed a crash occuring while loading certain SVGs exported from Adobe Illustrator

liquid600pgm:
I have already talked about the crash in #5523, and fixed it with my solution. It works like a charm now.

The crash was caused by the fact, that Illustrator exported certain SVGs with a `polygon` element instead of a `path`. The `polygon` element was not interpreted correctly by Processing, which led to a crash.

Here's an example of a previously bad SVG:

```svg
<svg id="Icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><title>Artboard 1</title><polygon points="18.72 4 13.39 20 10.61 20 5.28 4 7.39 4 12 17.84 16.61 4 18.72 4"/></svg>
```

The `polygon` element's `points` string was interpreted by Processing partially properly. Here's the pseudo-code of how the `points` string was processed:
```
if points is not null
  split points into an array with the delimiter ' '
  set vertices to a new array of size vertexCount*2
  for each string in the points array
    split the string to an array with the delimiter ',' (name: coords)
    set vertices[index][0] to the float value of coords[0]
    set vertices[index][1] to the float value of coords[1]
```

This could lead to problems, where the `points` string was "badly formatted", and thus, could lead to an `ArrayIndexOutOfBoundsException` in the for each loop.

I've replaced this with a regex-based solution, which works roughly like this:
```
if points is not null
  create a /([+-]?[\\d]+(\\.[\\d]+)?([eE][+-][\\d]+)?)(,?\\s*)([+-]?[\\d]+(\\.[\\d]+)?([eE][+-][\\d]+)?)/ matcher and set the search string to points
  for each match
    increase vertex count by 1
  reset matcher
  for each match
    set vertices[index][0] to the float value of the match's 1st group
    set vertices[index][1] to the float value of the match's 5th group
```

The regex's first group is the first coordinate (X), and the 5th group is the second coordinate (Y). I have explained the regex in detail in [this comment on #5523](https://github.com/processing/processing/issues/5523#issuecomment-394792948).

Anyways, now, the parser is compliant with the official SVG specification, and Illustrator SVGs are compatible.


benfry:
Great, thank you.

AshTheGoblin:
This issue, which I ran into today, was fixed the day after I downloaded  the pde for the first time. Almost perfect timing with this, thanks!

Issue #1280, opened 2013-02-10
Finish implementation of contains() function in PShape


processing-bugs:
_Original author: Orgi...@gmail.com (September 12, 2012 20:59:55)_

steps to reproduce the problem?
1. Intialize a PShape
2. Apply transformations(translate()/rotate()/scale()) to the PShape
3. Call PShape's contains() method in the draw() loop

Expected output and behaviour:
The contain methods uses the original(untransformed vertex coordinates for the test) only and therefore does not return the correct value when the PShape shape has transformations applied

This is using Processing 1.5.1 on osx 10.6.8

I've also attached a 'fix': storing transformations into a PMatrix2D instance and calling PShape's applyMethod (which btw only works with P3D/OPENGL - what's up with that ?). This way the inverse PMatrix can be used to convert x,y to the &quot;untransformed&quot; coordinate space so contains() can return the expected value. 
More details here: http://stackoverflow.com/questions/12327606/pshapesvg-contains-function-doesnt-work-when-scale-is-applied

<b>Note: Please do not link to locations on other sites, such as your own</b>
<b>image site, video site, blog, etc. Include the relevant information here</b>
<b>with the report.</b>

_Original issue: http://code.google.com/p/processing/issues/detail?id=1242_

JakubValtar:
This may be a nice improvement for PShape, see also #1609

PR #5549, opened 2018-06-19
Finish implementation of PShape.contains()

alexrj:
Transformation matrix is now considered before checking for point inclusion. #1280


benfry:
Ah! Much better, thank you.

PR #5550, opened 2018-06-19
Support PShape.contains() on GROUP objects

alexrj:
Currently, PShape.contains() throws a fatal exception when called on a non-PATH PShape object. This change allows the method to be called also on GROUP objects; in this case, it will loop through children and call `.contains()` on all of them until a match is found. Unsupported children (i.e. non-PATH or non-GROUP objects) will be skipped gracefully and a warning will be generated, but no fatal error will be emitted.

### DIFF
--- a/core/src/processing/core/PShape.java
+++ b/core/src/processing/core/PShape.java
@@ -2906,6 +2906,18 @@ public class PShape implements PConstants {
         }
       }
       return c;
+    } else if (family == GROUP) {
+      // If this is a group, loop through children until we find one that 
+      // contains the supplied coordinates. If a child does not support contains()
+      // just throw a warning and continue.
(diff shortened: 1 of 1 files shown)


alexrj:
This change should be reflected in documentation (pointers welcome).

gohai:
Looks good to me, thank you for your contribution @alexrj

benfry:
Thanks @alexrj for the fix and @gohai for the review.

hkiel:
Instead of try/catch shouldn't you just use children[i].getFamily() to check for valid children?

benfry:
It should actually just fall through to the `IllegalArgumentException`, since that's how the function behaves. Changed with https://github.com/processing/processing/commit/623e35c8d24ce78280936519db30475ad5951934

PR #5551, opened 2018-06-19
Added a rotate(angle, x, y) shortcut for rotating around a given point

alexrj:
Currently, Processing only offers a `rotate(angle)` function which rotates around 0,0. Rotation around a custom point is cumbersome and requires a combination of rotations and translations, as well as a `pushMatrix()`. While many advanced users are familiar with this logic, it overcomplicates the learning curve for students and beginners because it requires understanding of matrices.
I'm adding a `rotate(angle, x, y)` shortcut that simplifies this transformation.

I'm not sure I did everything right; let me know if this commit does everything we need for introducing a new instruction.

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -13285,6 +13285,7 @@ public class PApplet implements PConstants {
    * @param angle angle of rotation specified in radians
    * @see PGraphics#popMatrix()
    * @see PGraphics#pushMatrix()
+   * @see PGraphics#rotate(float, float, float)
    * @see PGraphics#rotateX(float)
    * @see PGraphics#rotateY(float)
    * @see PGraphics#rotateZ(float)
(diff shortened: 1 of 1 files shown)


jeremydouglass:
Questions (IANA core developer):

-  For consistency, how would this interoperate with the API for rotateX, rotateY, rotateZ? Would you propose that they should each have equivalent 3D (angle, x, y, z) forms?

-  Would users also expect this type of method signature with other transforms? For example, `scale(ratio, x, y)` (or the very-rarely-used `shear`)

alexrj:
@jeremydouglass both of your suggestions are look reasonable to me. Of course this means adding more signatures and I don't know whether this conflicts with the core design choices, but all those shortcuts definitely ease the learning curve.

AKonia:
Is it so useful ? I suppose it much better for Processing API don't add such methods, because the same thing can be done via rotate and translate.

benfry:
Thanks, though unfortunately, this would be inconsistent with the already-included `rotate(angle, x, y, z)`. (Inherited from [`glRotate()`](https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glRotate.xml))

Issue #5317, opened 2017-11-25
Characters like the "+" aren't recognised when using FX2D (italian keyboard)


Kryohi:
Ther was already a [bug report](https://github.com/processing/processing/issues/5231) on keycodes being different between P2D and FX2D, but I didn't find explanations on why something like 
`if (key=='+') {
...
}`
doesn't work.
Other keys do work though (like "."), maybe it's a problem of italian localization?

Happens both on windows and Linux with the latest release.

benfry:
Ouch, looks like that's the case for a lot of characters… It's because of the shift key. 

The short-term workaround is to use `keyTyped()` instead of `keyPressed()` if that'll work for your project. 

The issue is about how `key` and `keyCode` interact with key presses that are actually multiple keys at once.

benfry:
For anyone wanting to look into this, it's a matter of rewriting `getKeyChar()` inside `PSurfaceFX`. (And while you're at it, remove usage of that deprecated key char method because it's been removed in Java 9.)

christianbender:
@benfry  

hello, 

is this problem already fixed?

This code below works on Processing 3.3.6 on Linux:

```
void setup()
{
  //size(200,200, FX2D);
  //size(200,200, P2D);
  size(200,200);
}

void draw()
{
  if (key == '+')
   {
      //println("catch + inside draw"); 
   }
}

void keyPressed()
{
   if (key == '+')
   {
      println("catch + inside keyPressed"); 
   }
}
```

benfry:
It's not fixed. If you switch to `FX2D`, you won't see the message inside `keyPressed()` (confirmed on Mac and Windows; I've not checked Linux).

christianbender:
OK. On Linux exists the bug, too

> it's a matter of rewriting getKeyChar()

Have you more informations for me?
I have already download the file PSurfaceFX.java . But I don't know how I test it later?
I have installed Processing.

benfry:
@christianbender I think you've already found your way through this, but the build info is here: https://github.com/processing/processing/wiki/Build-Instructions

christianbender:
@benfry @ALL 
I debug the code above. I found that If the ```+``` key is trigged the function ```keyPressed``` will be active. But the value of the variable ```key``` is not the ```+``` sign but it is a special character. 
Where is the code for reading the key-value and put it in to the variable ```key``` ?

benfry:
[...]

PR #5563, opened 2018-06-28
Handle characters with shift modifiers in FX2D (#5317)

christianbender:
I fixed the issue #5317 .  And test it with the various modes.  (FX2D, P2D, none)  
```java
void setup()
{
  size(200,200, FX2D);
  //size(200,200, P2D);
  //size(200,200);
}

void draw()
{
  if (key == '+')
   {
     // println("catch + inside draw"); 
   }
}

void keyPressed()
{
   //println("key pressed! --> " + key);
   if (key == '+')
   {
      println("catch + inside keyPressed"); 
   }
}
```
Now the ```println("catch + inside keyPressed"); ``` in the function ```keyPressed()``` will correct fired.  
Thank @benfry for the help! Without you I had not found the proper place to fix.   

***NOTE:*** I added the shift characters ```* # < > , ! " § $ % & / ( ) = . : ; - _```

### DIFF
--- a/core/src/processing/javafx/PSurfaceFX.java
+++ b/core/src/processing/javafx/PSurfaceFX.java
@@ -1016,9 +1016,16 @@ public class PSurfaceFX implements PSurface {
       case MULTIPLY:
         return '*';
       case SUBTRACT:
+      case MINUS:
+        if (fxEvent.isShiftDown()) {
+          return '_';
+        }
         return '-';
       case ADD:
       case PLUS:
+        if (fxEvent.isShiftDown()) {
+          return '*';
+        }
         return '+';
       case NUMPAD0:
         return '0';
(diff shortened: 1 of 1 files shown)


gohai:
Thank you @christianbender - I've heard Ben is currently vacationing with family, but he'll sure have a look at your patches for upcoming releases

christianbender:
@gohai Thank you for the feedback.

liquid600pgm:
Just wanted to note that this is a very badly done fix, as it detects characters for one specific (UK) layout of the keyboard. I (and many people) have a US layout keyboard and the shift-keys are different here.

benfry:
Nuts, @liquid600pgm is correct. Let's move the discussion over to the original report.

Issue #4646, opened 2016-09-05
Parser error on very special case-y erroneous statement.


tillnagel:
While running

`boolean b = (boolean) 1;`

gives the correct error message of "Cannot cast from int to boolean" the following erroneous statement:

`boolean c = (boolean) 1 || (boolean) 1;`

throws an unhandled exception in the editor (with the above error message), and on running a weird "23)" message.

doms3:
I looked into this a little bit and I'm getting some strange behavior on version 3.3.6. 

When I first type in the line `boolean c = (boolean) 1 || (boolean) 1` I get the following error.

![image](https://user-images.githubusercontent.com/30886124/34311113-d31e33a6-e718-11e7-9fff-fd5275b82244.png)

However, if I change the code I get some different errors. For example, if I change both the ones to `true` and then back to ones I get this error:

![image](https://user-images.githubusercontent.com/30886124/34311184-525a1ef0-e719-11e7-823b-a39a276355c0.png)

Doing the same action but changing the trues back to ones in the other order I get the expected behavior for the second invalid statement.

![image](https://user-images.githubusercontent.com/30886124/34311233-af9d227e-e719-11e7-8ea1-070b5bcbb8ff.png)

Another error I saw:

![image](https://user-images.githubusercontent.com/30886124/34311293-01212b72-e71a-11e7-85ad-64b04749c396.png)


Similar behavior is observed for the `&&`, `|` and `&` operators, but `==` appears to work correctly.

christianbender:
@benfry @JakubValtar 
I will look into this. Have you further informations for me?

christianbender:
@benfry @JakubValtar @gohai 
The error message ```23)``` comes from 
```
org.eclipse.jdt.internal.compiler.impl.Constant.booleanValue(Constant.java:23)
```
In file ```Constant.java``` line 23 (above)

PR #5568, opened 2018-07-01
Add a simple extended error message (#4646)

christianbender:
I fixed the issue #4646.
Now we get for the code the following error message 
```processing
void setup() {
  size(200, 200, FX2D);
}

void draw() {
  boolean c = (boolean) 1 || (boolean) 1;
}
```

produces the error message
screenshot

instead of mere ```23)``` 

I put in a chunk of code that check the size of the error message. If the size to small (mere 3 characters) then the error message will be extended.

### DIFF
--- a/java/src/processing/mode/java/Compiler.java
+++ b/java/src/processing/mode/java/Compiler.java
@@ -62,6 +62,8 @@ public class Compiler {
     SketchException exception = null;
     boolean success = false;
 
+    // System.out.printf("DEBUG: class Compiler -- BEGIN compile(...)\n"); //DEBUG
+
     String baseCommand[] = new String[] {
       "-g",
       "-Xemacs",
@@ -78,6 +80,8 @@ public class Compiler {
     String[] sourceFiles = Util.listFiles(build.getSrcFolder(), false, ".java");
     String[] command = PApplet.concat(baseCommand, sourceFiles);
     //PApplet.println(command);
+
+    // System.out.printf("DEBUG: class Compiler -- BEGIN  -- compile(...) exception=%s\n",exception); // DEBUG 
 
     try {
       // Load errors into a local StringBuilder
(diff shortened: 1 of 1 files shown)


gohai:
That's clearly better than what it said before - but just curious: could this be made to return just "int constant cannot be casted into boolean", in a general way that would not have to be a special-case?

christianbender:
@gohai 
I can catch the case ```23)``` and then print ```int constant cannot be casted into boolean```. for this case. 
The advantage of my solution is that we can catch such error messages like ```23)``` complete with a extended message-text.

christianbender:
@gohai I changed the mechanism. Now we can catch right away the case ```23)```. In other cases the error message will be extended. 
screenshot

gohai:
You might want to remove the debug `println()` calls again in your commit. Also the `<= 3` seems a bit arbitrary - thinking that it might make sense to compare the full `line` of an error that behaves as expected, with the `line` that you get for `int constant cannot be casted into boolean`, and modify the `errorFormat` regular expression accordingly. Guessing that the regex is simply not taking every possible output into account.

Otherwise, as a band-aid, your current patch is clearly already an improvement (without the `println`s). Thanks!

christianbender:
@gohai 
* I removed my ```println(...)``` statments.  
* I reworte the meachnism, using regex. Checks whether the error message only comprise of words.

gohai:
Hi @christianbender - I see you created another regex. My suggest was rather to modify [this one](https://github.com/christianbender/processing/blob/6aeaec5328ff8f3e005dc45996fd25a7398c495b/java/src/processing/mode/java/Compiler.java#L141) so that it would work with whatever is going on with `23)` also.

If you look [here](https://github.com/processing/processing/pull/5568/files) you see that your pull requests still has some random changes that come with it, like commented-out `println()` statements. My git tricks, in the hope that they are useful for you also: I stage changes with `git add -p`, where I am being asked block by block whether I want to commit this particular change to the next commit [y] or not [n].

When I learn that I made a mistake in a previous commit I fix with `git rebase -i HEAD~1`. This lets you mark the commits you want to edit by placing an "e" at the beginning of each line. Now you can make retroactive changes to this commit, add them to be committed, and commit with `git commit --amend`, which picks up the commit message you already wrote. (This also works with more then just the recent commit, by changing `HEAD~1` to `HEAD~2` and so on.) And lastly, you can also use `git rebase` to "squash" (merge) different commits into one. This works by placing an "s" of the beginning of the line for the commits that you want to merge with its predecessor (also works with multiple). This way you can have one, logical commit at the end of a process of trial and error - instead of a spew of commits that do and undo various things. Note: whenever you rewrite history (this is what `git rebase` does), you need to push your changes to GitHub using `git push -f` - by default GitHub doesn't like history to be rewritten (it's indeed not a good practice, except for pull-requests branches, so before they are in a tree that other people might depend on).

christianbender:
@gohai 
* I changed the regex.  
* I removed useless ```println``` statments in the code.

Issue #5249, opened 2017-09-05
Hitting ESC in FX2D app on macOS throws IllegalStateException


benfry:
Using ESC to quit an app results in
```
java.lang.IllegalStateException: Not on FX application thread; currentThread = JavaFX Application Thread
	at com.sun.javafx.tk.Toolkit.checkFxUserThread(Toolkit.java:236)
	at com.sun.javafx.tk.quantum.QuantumToolkit.checkFxUserThread(QuantumToolkit.java:423)
	at javafx.stage.Window.setShowing(Window.java:921)
	at javafx.stage.Window.hide(Window.java:947)
	at com.sun.javafx.stage.WindowPeerListener.closed(WindowPeerListener.java:100)
	at com.sun.javafx.tk.quantum.GlassWindowEventHandler.run(GlassWindowEventHandler.java:119)
	at com.sun.javafx.tk.quantum.GlassWindowEventHandler.run(GlassWindowEventHandler.java:40)
	at java.security.AccessController.doPrivileged(Native Method)
	at com.sun.javafx.tk.quantum.GlassWindowEventHandler.lambda$handleWindowEvent$423(GlassWindowEventHandler.java:151)
	at com.sun.javafx.tk.quantum.QuantumToolkit.runWithoutRenderLock(QuantumToolkit.java:389)
	at com.sun.javafx.tk.quantum.GlassWindowEventHandler.handleWindowEvent(GlassWindowEventHandler.java:149)
	at com.sun.glass.ui.Window.handleWindowEvent(Window.java:1266)
	at com.sun.glass.ui.Window.notifyDestroy(Window.java:1183)
```

PR #5574, opened 2018-07-12
Fix crash at FX2D exit (#5249)

hkiel:


benfry:
Sorry, this isn't a usable fixed because it's just hiding the error; not actually fixing the underlying issue. The point of `Platform.exit()` is to call `System.exit()` only once it's safe to do so.

Issue #5573, opened 2018-07-11
color(int,int,int,int) is not thread save


Jesse-McDonald:
## Description
color(int,int,int,int) is not thread save and can cause colors to change in other threads without running fill or stroke

## Expected Behavior
running color(int,int,int,int)  in a separate thread should not effect any other threads, nor should any use of color(int,int,int,int) actually change the color being drawn

## Current Behavior
When ran with multiple threads, one running concurrently with draw, any use of the color(int,int,int,int) function (not fill or stroke) can result in elements being drawn to change to that color for 1 frame

## Steps to Reproduce
1.create thread with color(int,int,int,int) in it
2.use main thread to draw a rectangle
3. rectangle may be (race condition here people) the color from the other thread

## Your Environment
* Processing version: 3.37
* Operating System and OS version: Windows 10.0.17134
* Java 1.8.0_131
* hardware https://pcpartpicker.com/list/9LzptJ

## Possible Causes / Solutions
solution 1. The inputs of color(int r,int g,int b,int a) is bound checked then sent to to g.color(r,g,b,a) and the value is returned, rather than doing that after bound checking return `((a&0xff)<<24)+((r&0xff)<<16)+((g&0xff)<<8)  +(b&0xff)`
Solution 2. dont make color(int,int,int,int) final so as to allow other people to overload it if they wish
Solution 3. in PGraphic, dont have colorCalc set the variable calcColor, rather have it return an int, furthermore rather than stroke() calling strokeFromCalc, have stroke call strokeFromColor(int) and pass it the result from colorCalc rather than ever relying on calcColor
## Code
This code will create a blank sketch where the background flashes red, green, blue, and black
```java
class OtherThread extends Thread{
 public void run(){
   while(true){
     color(255,0,0);//nothing is being done with these color 
     color(0,255,0);
     color(0,0,255);
   }
 }
}
OtherThread thread;
void setup(){
  thread=new OtherThread();
  new Thread(thread).start();//start thread
}
void draw(){
  background(0);//set back ground to black, this should not be influenced by any source 
  //background randomly flashes red, green, and blue rather than being black  
}
```

StanLepunK:
I try to run your code but Processing don't know your class OtherThread...I change just for Thread, but nothing happen...
```
public void run(){
  while(true){
    color(255,0,0);//nothing is being done with these color 
    color(0,255,0);
    color(0,0,255);
  }
}

Thread thread;
void setup(){
  thread=new Thread();
[...]

PR #5575, opened 2018-07-12
Fixed issue #5573 about color(...) thread safety

Jesse-McDonald:
replaced all instances of PGraphics using calcColor with a PColor
all versions of colorCalc now return a PColor instead of void
all `____FromCalc()` methods replaced with `set____(PColor)` methods
changed all calls of `colorCalc(...); ____FromCalc();` to `set______(colorCalc(...));


benfry:
Closing this one since we don't have plans to change the overall behavior. More explanation here: https://github.com/processing/processing4/issues/431#issuecomment-1057979636

PR #5581, opened 2018-07-19
IO: Additional improvements

gohai:
Great to see you back, @benfry!

The following is dealing with minor fallout from 3.3.7.1 - the only noteworthy change would be:

* IO: Make I2C errors more verbose

Best,
G


benfry:
Thanks!

PR #5590, opened 2018-07-26
IO: Add ADS1X15 Analog-to-Digital converter example

gohai:
The Pi has no built-in analogRead, so this adds two very popular I2C ADCs - eg. part of these breakout boards from Adafruit:

https://www.adafruit.com/product/1083
https://www.adafruit.com/product/1085


benfry:
Thanks!

Issue #5481, opened 2018-04-22
Show warning when createFont() cannot find a named font


StanLepunK:
Maybe a print can be helpful for the case where the user try to create a font without extension like `.ttf`or `.otf` from unknown font by the system, and Processing use the default font ? 

```
void setup() {
  size(400,400);
  PFont f = createFont("Unknow_font",100); // don't find this font, Processing use default font
  PFont f = createFont("Unknow_font.ttf",100); // return null pointer exception
  // PFont f = createFont("Garamond",100); // find font
  println(f.getName());
}
```

gohai:
Agree that this could be a helpful feedback to have

benfry:
Not sure it's really possible… It's not that it uses the Processing fallback, it's Java's fallback (which is what Processing uses as the default). The name from `getName()` is platform/OS/Java version-dependent, so it won't consistently tell us that the font doesn't match what was requested.

Closest is to check `getPSName()`, but still, matching it against what the user has requested is difficult.

FlorisVO:
I'm a 'beginner' and I'm trying to fix this issue. I understand the source code and I understand why there is no warning displayed (@benfry you're right, it's not easy because a built in java method is used to get the font). 
After some research I know how to solve it. I have two possible solutions and can't choose which one would be the best.

To solve this issue we need all the installed fonts. Luckily, all installed fonts are stored in the static field `fonts` of `PDFont`. On MacOS the fonts are also stored in a hashmap with the font name - font object pair.

- Solution 1: when creating a new font loop  through the installed `java.awt.Font` which are saved in the static `font` field, get their name and compare it to the passed font name. **Pros** No more RAM needed. **Cons** Slow
- Solution 2: load all the font names in a static String hashset and search if the font exists. (a static hashmap already exists on MacOs platforms) **Pros** faster lookup **Cons** More RAM usage
- Solution 3? (maybe a better way?)

What is the opinion of a experienced developper? 😃 
Sorry for my english...

benfry:
Fixed with patch from @FlorisVO in #5605 for 3.4.1 or 3.5 (whichever is the next release).

PR #5605, opened 2018-08-03
Shows a warning when createFont can't find a named font (#5481)

FlorisVO:
I fixed this 'issue' (enchantment) #5481. createFont(...) shows a warning when a named font (without extension) is not found on the system.

### Example:
**Code**
`void setup() {
  PFont font = createFont("unknown",20);
}`

**Output**
`Warning: font "unknown" is missing or inaccessible on this system. Default font "Dialog.plain" is used.`


benfry:
Thank you!

PR #5622, opened 2018-08-15
Add Collapsible Code Blocks

etjos:
Adding Collapsible code blocks would allow for easier management of code with less files, as well as neater code. Since the Processing IDE was made to target more beginners, this could be set to toggle on, but not as default.


gohai:
The commits from the 3.3.7.x branch have no connection to what this PR is supposedly about.

PR #5626, opened 2018-08-19
Fixed for JEP-223 support

hbs:
This PR modifies the javaPlatform extraction method so it works with the newest java.version format (as per JEP-223).


benfry:
Thanks!

Issue #5625, opened 2018-08-18
pixelDensity(2) breaks the width of a tab ("\t") in JAVA2D


clankill3r:
Processing 3.4
Mac OSX 10.12.6

In the `JAVA2D` renderer, when having `pixelDensity(2)` with a lot of fonts the width of a tab character becomes 0.

```java
String[] fonts = PFont.list();
int i = 0;

void setup() {
   size(200, 200, JAVA2D);
   pixelDensity(2); 
   frameRate(5);
}

void draw() {
  
  String s = fonts[i++];
  textFont(createFont(s, 10));
  if (i == fonts.length) noLoop();
  
  background(0);
  text(s+"\nA\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\tB", 10, 10);    
}
```

(side node, the width of a tab in FX2D is huge compared to P2D for example)

liquid600pgm:
Fixed this issue in #5633.

PR #5633, opened 2018-08-27
Fix for #5625 - Made tab widths consistent across renderers

liquid600pgm:
Fortunately this is a very easy fix, just replace all tabs with spaces (I chose 4 as this is what I use as my tab width in text editors, however this should be subject to change. Maybe make a `tabWidth()` function?)


benfry:
Interesting fix… I'd rather get to the source of the problem, but I think this is rare enough that it's ok to go this route.

liquid600pgm:
I know that this fix is not very accurate, because tabs are supposed to help with making tables etc. but I might create a better fix at some other point in time.

clankill3r:
I am pretty sure you just broke `textWidth`, you need to replace the tabs there as well.

jeremydouglass:
Don't break `textWidth`! I like the idea of this change, but if it comes down to a choice I personally I think supporting dynamic typography logic is more important than cross-renderer consistency. My two cents.

liquid600pgm:
@clankill3r I didn't think about that possibility, I'll make a new version of this, fixing the broken fix that breaks things.

benfry:
Ok, backing out that fix. We'd need to add that hack to 4-5 places in the code, and it still wouldn't cover text from `char[]` arrays and it doesn't actually implement tab support anyway.

Issue #5562, opened 2018-06-28
Find in Reference disabled for various keywords (draw, for, if, catch, while)


KevinWorkman:
Possibly related to #5555

From this forum post: https://discourse.processing.org/t/reference-and-help-tabs-greyed-out/1340

I have this code into my Processing editor:

```java
void draw() {
  for (int i = 0; i < 10; i++) {
    if (i < 5) {
      println(i);
    } else {
      println(i);
    }

    try {
      println(i);
    }
    catch(Exception e) {
      println(i);
    }
    
    do{
     println(i); 
    } while(false);
    
    while(true){
     println(i); 
    }
  }
}
```

If I right-click on various keywords, I'll get a menu, and one of the items on that menu is **Find in Reference**, which takes me to a local copy of the reference page for that keyword.

This works fine for `int`, `else`, `println()`, `try`, `true`, and `false`.

However, for many other keywords I see that option grayed out in the menu. This is the case for `draw()`, `for`, `if`, `catch`, and `while`.

I'm using Java mode, and all of those keywords are in my `keywords.txt` file, and they all have a reference page.

benfry:
Huh, not sure what's happening there. Thanks for the report.

svr8:
Hey! Found the error in [keywords.txt](https://github.com/processing/processing/blob/master/java/keywords.txt) and references.zip. Please review this PR: https://github.com/processing/processing/pull/5642

GRsni:
This issue should probably gets closed too, seeing #5555 was closed already.

benfry:
No, #5555 was closed as a duplicate because this has a better description of the problem.

benfry:
We're on the trail of the problem, but @REAS is digging into it a bit more since it involves rebuilding the reference and `keywords_base.txt`.

PR #5642, opened 2018-09-05
Fixed Reference Issue #5562 and #5737

svr8:
Looking at Mode.java:
https://github.com/processing/processing/blob/38b9f3ff7f96086d184f35df5bd4cb9bbda7e834/app/src/processing/app/Mode.java#L162-L175
It seems the map with 3 parameters are defined for tokens that are followed by opening parentheses `(`. For example, function draw is defined as:
https://github.com/processing/processing/blob/38b9f3ff7f96086d184f35df5bd4cb9bbda7e834/java/keywords.txt#L458
It should have been `draw FUNCTION4 draw_`
The selected/highlighted strings that are followed by `(` are automatically appended by the `_` suffix internally:
https://github.com/processing/processing/blob/38b9f3ff7f96086d184f35df5bd4cb9bbda7e834/app/src/processing/app/ui/Editor.java#L2308-L2310

The references.zip file (in root/java, added to gitignore by default) had inappropriate file names. Similarly, the mapped values in [keywords.txt](https://github.com/processing/processing/blob/master/java/keywords.txt) did not match with file names.

I've fixed the mapped values and filenames for the tokens of type FUNCTION3 and FUNCTION4.

### DIFF
--- a/java/keywords.txt
+++ b/java/keywords.txt
@@ -1,17 +1,17 @@
 # THE TEXT BELOW IS HAND WRITTEN AND FOUND IN THE FILE "keywords_base.txt"
 # IN THE PROCESSING-DOCS REPO. DON'T EDITS THE keywords.txt FILE DIRECTLY.
 
-# For an explanation of these tags, see Token.java 
+# For an explanation of these tags, see Token.java
 # trunk/processing/app/src/processing/app/syntax/Token.java
 
 ADD	LITERAL2	blend_
-ALIGN_CENTER	LITERAL2	
-ALIGN_LEFT	LITERAL2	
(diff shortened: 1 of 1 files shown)


benfry:
Thanks, I'll take a look when I get a chance.

svr8:
MINOR UPDATE: I have added Long keyword in /java/keywords.txt to fix this issue #5737 .

benfry:
Working on this one ([notes here](https://github.com/processing/processing/issues/5562#issuecomment-455827810)), though we can't use this patch since it's modifying the auto-generated file (see the message directly above your changes).

svr8:
Oh okay! I didn't see those messages before.

Issue #5645, opened 2018-09-09
P2D renders images with alpha channels incorrectly if large VLW fonts are used


Sardtok:
## Description
When setting the font to a large VLW font, rendering text and images in a sketch using the P2D renderer, something happens to the transparency mode that makes images render incorrectly.

## Expected Behavior
Transparent pixels in the image should be transparent instead of having the image background color or other colors from elsewhere in the same row.

Using the Java 2D (JAVA2D or unspecified) or JavaFX (FX2D), results in the correct behaviour.
Downgrading to Processing 3.3.7 results in the correct behaviour.

## Current Behavior
Transparent pixels are not rendered as transparent, and the color is not necessarily as expected. It can look a bit like

## Steps to Reproduce
1.Create a font with the font creation tool, e.g. Serif.bold at 72px (seems the threshold is 40px).
2. Include a PNG with an alpha channel (might require 8-bit alpha).
3. Set the font to use the VLW with textFont and loadFont.
4. Make sure you use P2D or P3D renderer.
5. Draw some text and the image in draw.

```
PImage button;

void setup() {
  size(800, 800, P2D);
  button = loadImage("button-green2.png");
  textFont(loadFont("Serif.bold-72.vlw"));
}

void draw() {
  background(0);
  image(button, 200, 200, 400, 400);
  text("TEST", 0, 50);
}
```

## Your Environment
* Processing version: 3.4
* Operating System and OS version: Windows 10 (1803 / 17134.228)
* Other information:
nVidia GeForce GTX 1050 and Intel UHD Graphics 620 (same result on both)
Tested in 4K (pixel density 2) and 1080p (pixel density 1).

Sardtok:
![p2dimagesandsmoothvlw](https://user-images.githubusercontent.com/613667/45266606-0e9e9700-b45e-11e8-913e-e4847d6ec18a.png)

Sardtok:
Hmm, it seems this is related to font size, not whether it is set to smooth or not.

Sardtok:
Works in Processing 3.3.7.

Sardtok:
I am almost certain the problem is the change in how blend modes are set in 3b42f225cb4e2a2e4e6222e8c5b761dfb4cee780

I think the test should read:
```
if (this.blendMode != s.blendMode)
```
And not the current one, which says if the current blend mode is anything but REPLACE, set it to whatever the PStyle's blend mode is.

[...]

PR #5647, opened 2018-09-10
Fixes blend mode not being set correctly, fixing #5645.

Sardtok:
Blend mode can incorrectly be set to REPLACE in P2D when rendering certain bitmap fonts. This breaks rendering of images with alpha channels in the same sketch on 3.4. See #5645 which is fixed by this PR.


benfry:
Thanks!

PR #5653, opened 2018-09-18
Fix #5317: Modifier keys not recognized with FX2D

### LINKED ISSUE #5317: Characters like the "+" aren't recognised when using FX2D (italian keyboard)

liquid600pgm:
This fixes #5317. The fix was rather simple: just removing some default return values, and changing the deprecated `KeyCode#impl_getChar()` method call to `KeyEvent#getText()`.

### DIFF
--- a/core/src/processing/javafx/PSurfaceFX.java
+++ b/core/src/processing/javafx/PSurfaceFX.java
@@ -1011,40 +1011,6 @@ public class PSurfaceFX implements PSurface {
         return PConstants.CODED;
       case ENTER:
         return '\n';
-      case DIVIDE:
-        return '/';
-      case MULTIPLY:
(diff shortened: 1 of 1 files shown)


benfry:
Not quite this simple; this would cause it to return `H` when you press the HOME key, or `F` when you press F1 through F12. 

All these just need to be re-mapped so they behave the same way they do w/ the default renderer.

PR #5673, opened 2018-10-07
Minor changes for russian translation

AKonia:
Check it please @benfry

### DIFF
--- a/build/shared/lib/languages/PDE_ru.properties
+++ b/build/shared/lib/languages/PDE_ru.properties
@@ -59,8 +59,8 @@ menu.library.add_library = Добавить библиотеку...
 menu.library.contributed = Вклад сообщества
 menu.library.no_core_libraries = отсутствуют основные библиотеки
 # ---
-menu.sketch = Эскиз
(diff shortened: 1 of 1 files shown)


benfry:
So this looks like a difference of opinion in how “sketch” is to be translated. Can any other Russian speakers weigh in on what's appropriate here?

benfry:
Ok, will just have to trust you on this one.

cansik:
@benfry Checked it with another Russian speaker (not processing user). She told me that `Эскиз` is something which is more developed than `Набросок `.  Checking yandex image search shows that `Набросок` is maybe more something like a scribble. 

I still don't know which matches better, but just to give a second input.

#### Эскиз
image

#### Набросок
image

#### sketch
image

AKonia:
@cansik Both `Эскиз`and `Набросок` are commonly synonyms, but second is native Russian, first is taken from French and sounds more unusual. I also noticed out that Sony's Android app `Sketch` also was translated as `Набросок`. Finally I'm native Russian speaker and developer so I can surely say that both words are using only in art and has no association with programming at all.

PR #5681, opened 2018-10-13
Add unbinary int signature

jeremydouglass:
Accepts an integer that uses decimal format to represent a sequence of bits, and returns an integer in native format.

For example, the integer 1101 represents binary, and returns 13 (binary 0b1101, or 8+4+0+1 = 13). 

This signature extends unbinary(String value). It is a workaround for lack of parser support for binary literals such as 0b1101.

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -10241,6 +10241,28 @@ public class PApplet implements PConstants {
     return stuff;
   }
 
+  /**
+   * Accepts an integer that uses decimal format
+   * to represent a sequence of bits, and returns
+   * an integer in native format.
+   * 
+   * For example, the integer 1101 represents binary,
+   * and returns 13 (binary 0b1101, or 8+4+0+1 = 13).
+   * 
+   * This signature extends unbinary(String value).
+   * It is a workaround for lack of parser support for
+   * binary literals such as 0b1101.
+   */
+  static final public int unbinary(int value) {    
+    int result = 0;
+    int i = 0;
+    while (value != 0) {
+      result |= (value % 10) << i;
+      i++;
+      value /= 10;
+    }
+    return result;
+  }
 
  /**
    * ( begin auto-generated from unbinary.xml )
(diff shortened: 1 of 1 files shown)


benfry:
How would this be used in actual practice? I've never personally run into a situation where binary values were stored as ints, or that it's necessary to have something higher performance than `unbinary(int(value))`.

jeremydouglass:
Not really performance -- I just thought it made more sense intuitively for learner programmers to have the option to write numbers as numbers in a format they could read, then wrap them in conversion functions -- like for example they do with degrees / radians.

Hmm. Here is an example: writing a channel mask method that filters the a,r,g,b channels, it makes sense for it to take a binary number as user-friendly, readable argument. So instead of writing "13" we would write it like this:

    img.pixels = tintMask(img.pixels, 0b1101); // channel mask ARGB bits: AR_B

...but we can't write binary literals, so we write this instead:

    img.pixels = tintMask(img.pixels, 1101); // channel mask ARGB bits: AR_B

and the method uses unbinary to support a binary argument, like this:

```
  int[] tintMask (int[] pixels, int channelMask) {
    channelMask = unbinary(channelMask);
    if (channelMask == 15) {  // ARGB
      return pixels;
    }
    if (channelMask == 0) {  // XXXX
      return new int[pixels.length];
    }
    // process channels separately
    // ... etc etera
```

jeremydouglass:
To follow up on this -- the idea was to allow learners to directly code:

    unbinary(101011)

instead of

    unbinary(str(101011));

The implementation also happens to be higher performance than calling string on the argument and forwarding it to the other signature, but that isn't the main advantage -- the point is it makes code using unbinary easier to read to write, and that plus better performance seemed like a win-win-win.

dzaima:
> Given the dramatic change is syntax capabilities, is that branch a candidate for Processing 4, or would it become part of 3?
There have been discussions about that, though IIRC nothing's set in stone yet.

> One note, though -- unbinary is specifically a converter, so you don't need to `unbinary(0b10101)`. The only things you need to unbinary are `"10101"` and, potentially, `10101`-- things that are not in the right format and need converting.
I don't see really any use case of specifically interpreting a non-constant base-10 `int` as a base-2 one back to `int`, definitely one not big enough to warrant a builtin. Strings make sense as that is a generic format of anything, but storing a number as base-10 base-2 number is just wrong.

And as code review, I'd really expect such a builtin to complain about an invalid input (e.g. `unbinary(1234)` as otherwise it can lead to very strange & unexpected results, and I wouldn't be surprised if many didn't understand what it does)

jeremydouglass:
Very good point about adding bad input handling.

Given the open possibility of adding 0b notation support, I'll close this PR.

However, I want to respond to:

> storing a number as base-10 base-2 number is just wrong

I disagree on this point, it is no more "wrong" than storing a number as String base-2. The premise of unbinary() is that the input needs converting. The user -- eg a learner typing code -- is writing in a format that is easy for them to read and marking their intent through that conversion. If a user expressing binary values in non-binary was wrong then we would drop unbinary() and force users to encode directly in a more virtuous way.

That doesn't mean this PR is necessary an improvement--it may be unneeded, or logically inconsistent in some way with unhex() etc. But IMO it isn't "just wrong".

PR #5697, opened 2018-11-07
Stop frame rate counter from exaggerating

JakubValtar:
Frame rate counter finally gets its ticket for speeding.

Frame rate is now calculated by averaging frame times instead of averaging frame rates. Rationale is explained in the comment.


benfry:
Cool, thanks.

PR #5698, opened 2018-11-07
Fix vertex buffer initialized with wrong number of components

JakubValtar:
Poly vertex buffer was initialized with 3 components, but has 4.


benfry:
Surprised that even worked…

PR #5699, opened 2018-11-07
Recreate FBOs when offscreen PGraphicsOpenGL is resized

JakubValtar:
Otherwise FBOs keep the old size and things understandably break.


benfry:
Danke.

PR #5719, opened 2018-12-10
add boolean mouseOver property

### LINKED ISSUE #5700: Add a way to tell if the mouse is over the window.
## Description

I want an easy way to tell if the mouse is over the window.

## Expected Behavior

I should be able to call a method or check a boolean to know if the mouse is currently over the window.

## Current Behavior

There is no way that I can find to know if the mouse is over the window. If it is not, mouseX and mouseY will remain in their last values.

## Steps to Reproduce

1. Make some code that does something depending on where the mouse is.
2. Move the mouse off the window.

## Your Environment

 

* Processing version: 3.4
* Operating System and OS version: XUbuntu 18.10
* Other information: Using processing as a library in Eclipse.

## Possible Causes / Solutions

Ideally, a boolean, mouseOverWindow or something like that, updated the same way mouseX and mouseY are.

anubhavp28:
Fixes #5700 
Add a boolean property mouseOver to indicate whether the mouse is over the window or not.

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -560,7 +560,30 @@ public class PApplet implements PConstants {
    * @see PApplet#mouseWheel(MouseEvent)
    */
   public boolean mousePressed;
-
+  
+  /**
+   * ( begin auto-generated from mousePressed_var.xml )
+   *
+   * Variable storing if the mouse is currently over the window. The value of the system
+   * variable <b>mouseOver</b> is true if the mouse is over the window and
+   * false if the mouse is not over the window.
+   *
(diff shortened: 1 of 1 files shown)


jeremydouglass:
For the docs, if this was included I would expect it to then be cross-referenced by all the other mouse entries -- in particular mousePressed, but all the ones that cross-reference all the others.

benfry:
This is better handled by the solution in the original report. `mouseOver` isn't a property we want to maintain—too tricky to figure out how it's handled when dealing with other/newer display devices like touch screens. It's easy enough to do the quick workaround, but a heavier lift on documentation, etc.

PR #5733, opened 2018-12-21
Added method setMinimumSize(int x, int y) : void

Maharal:
When I used the method `setResizable(boolean resizable) : void`, I had trouble resizing the application window. I was able to make windows with 0px height and width, now the programmer can set a minimum window size.

### DIFF
--- a/core/src/processing/awt/PSurfaceAWT.java
+++ b/core/src/processing/awt/PSurfaceAWT.java
@@ -583,6 +583,15 @@ public class PSurfaceAWT extends PSurfaceNone {
     }
   }
 
+    /** Set the minimum size of frame */
+  @Override
+  public void setMinimumSize(int x, int y) {
+    if (frame != null) {
+      sketchWidth = x;
+      sketchHeight = y;
+      frame.setMinimumSize(new Dimension(x, y));
+    }
+  } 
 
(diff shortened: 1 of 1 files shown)


core/src/processing/awt/PSurfaceAWT.java
```diff
@@ -583,6 +583,15 @@ public void setResizable(boolean resizable) {
     }
   }
 
+    /** Set the minimum size of frame */
+  @Override
+  public void setMinimumSize(int x, int y) {
```
Shubham-100: Use unit instead of int

Maharal: unit? What type is this?

Shubham-100: Oops, I meant uint.

Maharal: I was looking at other methods and none uses the type that you propose. Why should I use the type `uint` you propose? Other methods use int, like: setLocation, setCursor, setSize. And a make a research and I don't find this type.

benfry:
This isn't sufficient to handle sizing across the various renderers and platforms; would need more attention to that.

Issue #5746, opened 2019-01-06
Allow Processing to run on OSX using JDK > 1.8 by utilizing the new java.awt.desktop classes


ssilverman:
## Description
`processing.core.ThinkDifferent` should use the `QuitHandler` from the newer `java.awt.desktop` package in addition to `com.apple.eawt.QuitHandler`. Some adapter code could be written that allows this to run on all JDKs, without explicitly removing `com.apple.eawt.QuitHandler` support.

## Expected Behavior
Allow Processing to work on OSX beyond JDK 1.8, while still keeping compatibility.

## Current Behavior
Only works on JDK 1.8 and below.

## Steps to Reproduce
Run any Processing app on JDK > 1.8.

## Your Environment
* Processing version: 3.4
* Operating System and OS version: macOS 10.14.2
* Other information:

## Possible Causes / Solutions
Better cross-platform support was added for "desktop" apps and this removed some of the Apple-specific classes in favour of JDK-provided ones.

ssilverman:
The pull request: #5747.

benfry:
Fixed with https://github.com/processing/processing4/pull/1 and continued development happening at https://github.com/processing/processing4

PR #5747, opened 2019-01-07
Making the ThinkDifferent classes be JDK version-independent. JDK 9…

ssilverman:
… and later have a new java.awt.desktop package that replaces the custom handlers in the com.apple.eawt package. This commit uses java.lang.reflect.Proxy to generate handlers of the correct type, independent of the handler parameter types, which are different across JDK versions.

Specifically, this fixes being able to run a Processing app (externally or otherwise) on OSX using a JDK having a version > 1.8 (i.e. 9 and above).


ssilverman:
Fixes bug: #5746.

app/src/processing/app/platform/ThinkDifferent.java
```diff
@@ -131,12 +131,51 @@ public void handleQuitRequestWith(QuitEvent event, QuitResponse response) {
...
   }
-  
+
 
 //  public ThinkDifferent(Base base) {
 //    this.base = base;
 //  }
 
+  /**
+   * Sets a handler on an instance of {@link Application}, taking into account JVM version
```
ssilverman: Should say "JDK version", but I don't want to make a new commit.

benfry: No, JVM is correct.

ssilverman: My thinking on this is that the JDK includes the JVM plus libraries and other misc. stuff. You could theoretically change package names in the JDK 1.8 class files to JDK11-compliant names (eg. change `com.apple.eawt` to `java.awt.desktop`), and then run them on an older JVM (subject to only non-native code, etc). So my view is that this is still a JDK difference and not a JVM difference. JVM, in my mind, includes bytecodes plus support, and doesn't include the libraries (with some exceptions, eg. invokedynamic implementation).

ssilverman: To play devil's advocate with myself: Originally, "Java" was defined as a JVM plus libraries plus language, so in reality, I think the lines are kind of fuzzy anyway. I won't object to whatever's chosen, "JVM" or "JDK".

benfry:
Fixed with https://github.com/processing/processing4/pull/1 and continued development happening at https://github.com/processing/processing4

Issue #5765, opened 2019-01-22
Control-click on Mac OS X result in all subsequent clicks reported as right-click


mcslee:
## Description
Performing a ctrl-click on Mac OS X results in the click button being reported as RIGHT, which is correct. However, it sets a flag which is never unset, such that *all* subsequent clicks in the operation of the program are reported as right-clicks, for the rest of the program's operation - whether or not the control button is down.

## Expected Behavior
Only clicks performed while the control modifier key is actively down should be reported as a RIGHT click.

## Current Behavior
Every click after the first control-click is reported as button = RIGHT.

## Steps to Reproduce
```
void setup() {
  registerMethod("mouseEvent", this);
}

void draw() {
}

void mouseEvent(MouseEvent mouseEvent) {
  if (mouseEvent.getAction() == MouseEvent.PRESS) {
    println(mouseEvent.getButton());
  } 
}
```

Run the above and observe that after a control-click, on Mac OS, all future events are reported as RIGHT button.

## Your Environment
MacOS Mojave 10.14.2
Processing 3.5.1

## Possible Causes / Solutions
The issue is caused by this line:
https://github.com/processing/processing/commit/2d2b4779013045512cb469e615b0dc376672292a#r32029420

benfry:
Ouch; thanks for finding that.

mcslee:
Sure thing - appreciate the quick merge!

benfry:
Fixed with #5766 for Processing 3.5.2.

PR #5766, opened 2019-01-22
Fix bug with control-click for RIGHT on Mac OS, addresses #5765

mcslee:
When a control-click occurs, the macosxLeftButtonWithCtrlPressed flag
was set, but was never being unset due to accidental comparison with
the button variable instead of action variable. All future mouse
operations reported as right-click on Mac OS. This fixes the comparison
so that only the operations from a mouse press with the control button
down are reported as RIGHT. The flag is cleared after mouse release.

### DIFF
--- a/core/src/processing/core/PApplet.java
+++ b/core/src/processing/core/PApplet.java
@@ -2697,7 +2697,7 @@ public class PApplet implements PConstants {
                                event.getX(), event.getY(),
                                button, event.getCount());
       }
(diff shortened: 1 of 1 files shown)


benfry:
Thanks!

benfry:
Incorporated for 3.5.2.

JakubValtar:
@mcslee Good job spotting this, I made a typo when moving the code. Thanks!

Issue #3055, opened 2015-01-22
Replace ANTLR 2.x with updated ANTLR 4


benfry:
We're using a very old version of ANTLR. @fjenett did some work to move us to ANTLR 4, but there hasn't been enough testing (or someone committed to finishing that work) to include it in 3.x yet. 

See also https://github.com/processing/processing/issues/3054, which is about updating the grammar (the `java.g` file) to support newer Java language features.

benfry:
Link to Florian's earlier work: https://github.com/fjenett/processing-preprocessor-antlr4

Note that because this is so fundamental to what the PDE does, it needs a lot of testing and a committed maintainer (or at least longer than "hey this is working / incorporate it / bye"). We can't incorporate a "mostly working" set of code.

JakubValtar:
I just happen to have a working ANTLR4 prototype branch based on Florian's work with complementary Android branch I worked on in August... it was more of a proof of concept work, but tests are green and I can run Java and Android sketches from PDE. Needs update to current HEAD, new error reporting, better tests. API for other modes will need some thought.

https://github.com/JakubValtar/processing/tree/feature-antlr4
https://github.com/JakubValtar/processing-android/tree/feature-antlr4-android

It's a great responsibility and I have to consider if I will be able to do good enough job and maintain it. However, I can at least make it work in current version in the meantime.

benfry:
Cool, thanks for looking into it. Keep us posted on how it evolves, and/or someone may also be able to help build on your work as you've done w/ Florian's. 

Here's an excerpt from an email from Florian:
```
As a question remains how we should handle errors from the parser. ANTLR-4 comes with a nice listener interface that let's one subscribe to the parser. This would make error handling a lot cleaner and easier for us … but would mean that JavaBuild would need to implement the listener interface:

https://github.com/fjenett/processing/blob/antlr4-preproc/app/src/processing/mode/java/preproc/PdePreprocessor.java#L127

The compilation and runtime error handling needs some rethinking too. The way i generate the .java files leaves the headerOffset solution not functioning anymore as all in-between imports are moved to the beginning of the file. One way of solving this could be that i return a mapping between sketch and java lines so we could do a simple reverse lookup. That would make integration with PDE-X easier too i assume.

As a goodie i am using the parser to fish out the arguments from the size() call. Currently it is only printed to console but could easily be returned through the ProcessorResult …

[...]

PR #5776, opened 2019-01-27
trash solution for color method and give posibility to naming a variable color

StanLepunK:
Sure, it's my first trash pull request. I try to work around color problem.
When we remove the variable color as a `int`, after that we can use `color` for naming variable and method. For me that's can be a solution, because i'm not sur keep `color` like a int is very useful and advantage, because it's easy to understand it's a `int ` and my students understand easily when I write `int c = colot(255,0,0)`
May be a lot of people use var `color` so it's a balancing choice.  May be a little survey / poll to knnow if a lot of people use `color`like a `int `.

Happy to make this first pull request for the community and manage it... very long day to do that !

### DIFF
--- a/java/src/processing/mode/java/preproc/PdeEmitter.java
+++ b/java/src/processing/mode/java/preproc/PdeEmitter.java
@@ -687,11 +687,12 @@ public class PdeEmitter implements PdeTokenTypes {
       break;
 
     // the color datatype is just an alias for int
+    /*
     case LITERAL_color:
       out.print("int");
       dumpHiddenAfter(ast);
       break;
-
+    */
     case WEBCOLOR_LITERAL:
       if (ast.getText().length() != 6) {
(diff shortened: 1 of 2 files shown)


StanLepunK:
example:
```
void setup() {
  background(color());
  int color = 23456;
  background(color);
  color c = 0;
  background(color);
}

int color() {
  return 0;
}
```

benfry:
Nah, the solution isn't to disable `color`. We need a new preprocessor.

StanLepunK:
it was just in case, I figure me you say no !!! But it was a good exercice to introduce me to try to find a solution, fork, clone, run and push. That's take my week-end but now I understnd better how pull request work and console to.
When you write new preprocessor, you talk about totally rewrite this part of code ? `PdeEmitter.java`and `pde.g`?

benfry:
Yeah, more here: https://github.com/processing/processing/issues/3055.

Issue #4057, opened 2015-10-24
mouseX starts at 0, mousY starts at 1


clankill3r:
This seems to be operating specific. Linux doesn't have this. For windows I don't know.
For mac it does.

mouseX can go to 0
mouseY lowest value is 1

(this is also in processing 2 and probably 1.5.1 but that doesn't start on my mac anymore).

````java
void setup() {
  size(800,  800);  
}
 
void draw() {
   frame.setTitle(mouseX+" "+mouseY); 
}
````

iamutkarshtiwari:
I tested this on my linux and both co-ordinates are detected perfectly fine till (0,0). The problem might be that you move your pointer faster than the speed at which it's coordinates are tracked. Try taking the pointer to the top-left corner(origin) very slowly. You will surely get your (0, 0) :)

clankill3r:
No it wont. Only for x, not for y.

GKFX:
If you put `set(0, 0, #FF0000);` into `draw()`, can you see the red pixel? Does Processing respond when you hover over it? It seems likely that the window decoration is overhanging the content area. Also try `fullScreen()` which should remove the window decoration.

clankill3r:
I can see the red pixel. I only can' get the mouseY to 0.

````java
void setup() {
  size(200, 200);  
}
void draw() {
  cursor(CROSS);
  set(0, 0, #FF0000);
  surface.setTitle(""+red(get(mouseX, mouseY)));
}
````

GKFX:
I don't know then!

benfry:
I suspect this is a macOS quirk that's out of our control. 

1. Does it happen with all renderers? If so, then it's almost certainly out of our control. 
2. If it only happens with the default renderer, it might be fixable. 
3. If the y=0 line of pixels is visible, then it's also almost certainly out of our control.

clankill3r:
I tested,
1. Only the default renderer, P3D does not has this bug.
2. ^
3. All pixels are visible.

So it seems like a bug in the default renderer.

JakubValtar:
@clankill3r Do you by chance use display scaling/retina/whatever it is to make windows bigger on hi res display?

clankill3r:
I fixed the issue. When the mouse is reaching 0 in the y direction then the `action` is `MouseEvent.EXIT`, which is not processed by `handleMouseEvent`.

I choose to also improve on the issue, now on a EXIT it will stay on a border.

```java
void setup() {
 size(200, 400); 
}

[...]

PR #5796, opened 2019-02-16
fix mouseY not reaching 0 and general improvement on MouseEvent.EXIT

clankill3r:
https://github.com/processing/processing/issues/4057

This is my first contribution ever. I was not sure where to place the comments.
I didn't want to place them in between the `if` and the `else if` cause it would make the `else if` harder to spot.

Please take a look at https://github.com/processing/processing/issues/4057#issuecomment-464332982 as well.


benfry:
Does it do the right thing at the other edges of the window? (left, right, bottom?)

clankill3r:
Yes it does:

PR #5818, opened 2019-03-15
 ADD : Dialouge after Install/Update/Remove contributions

DeerajTheepshi:
- The user was not being notified when the installation of Contribution was completed.
- This PR involves displaying a dialog , when installation is complete or removed or updated
- Solves Issue : #5822

### DIFF
--- a/app/src/processing/app/contrib/DetailPanel.java
+++ b/app/src/processing/app/contrib/DetailPanel.java
@@ -550,7 +550,7 @@ class DetailPanel extends JPanel {
 
       ContribProgressBar downloadProgress = new ContribProgressBar(installProgressBar) {
         public void finishedAction() {
-          // nothing?
+          JOptionPane.showMessageDialog(null, "Installation Completed");
         }
 
         public void cancelAction() {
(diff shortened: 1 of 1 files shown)


SyamSundarKirubakaran:
> installation of Contribution

Can you make this more clear.!? Also, specify the issue that this PR closes.

DeerajTheepshi:
> > installation of Contribution
> 
> Can you make this more clear.!? Also, specify the issue that this PR closes.

By Contribution , I meant, mode/library/tools/Examples installation

kfrajer:
Hi Deeraj, I checked the PR and it requires more testing. The changes seems to work for the `update` action for the trivial case. However, it does not trigger for `remove` and it does trigger for `install` but only after downloading step and not after completing the installation of the library. I only tested this when installing/updating/removing libraries. I have not tested this for tools, examples or mode installs. 

One suggestion is to place your change at the end of the following function in the same file: `private void finishInstall(boolean error)`. You can test it in the different scenarios and provide details on your tests. This information will make easier to evaluate the PR.

PR #5831, opened 2019-04-02
[enhancement/grammar/spelling] - processing/processing/issue_template.md

tmm88:
[enhancement/grammar/spelling] - processing/processiang/issue_template.md


jeremydouglass:
No, that is not a correct use of "between" -- it would introduce an error.  "Between" requires two points of comparison.

What is the difference between A and B? CORRECT
What is the difference between A? WRONG

Issue #4507, opened 2016-05-23
Case sensitivity with libraries


prisonerjohn:
I know the issue of case-sensitivity was brought up before, but this is regarding loading contributed Libraries.

If we look at RiTa as an example, it downloads with no errors from the Contribution Manager, but will not work when used in a sketch.

![capture](https://cloud.githubusercontent.com/assets/119702/15476409/bff59b60-20dc-11e6-82a4-a2f2074bd4f0.PNG)

It works if we change the folder name to match the package name, but that's now different from the Library name.

![capture2](https://cloud.githubusercontent.com/assets/119702/15476445/ed1265e2-20dc-11e6-8faf-808895515e76.PNG)

And that's only an issue on Linux, because that seems to be the only OS that is case-sensitive.

Possible solutions:
* Add a step when downloading a Contribution to transform the path to lowercase.
* Advise authors to use lowercase for all folders in their hierarchy.
* Something else?

cc @dhowe

PR #5844, opened 2019-04-21
FIX Library never appears installed Issue

DeerajTheepshi:
Solves issues reported with Case sensitive file system of Linux and libraries not listed as downloaded/installed in the CM.

Closes the following issues: 
#4507
[https://github.com/processing/processing-android/issues/300](url) (Processing - Android)


benfry:
This shouldn't be necessary and is hopefully resolved in more recent releases.

PR #5881, opened 2019-06-14
Workaround for issue #4104

ghjansen:
Conditional control of FBO layer, based on environment settings.
Looking forward for feedback on any adjustment required.

More info:
Issue #4104 
[Comment explainig the patch](https://github.com/processing/processing/issues/4104#issuecomment-502246089)

### DIFF
--- a/core/src/processing/core/PSurface.java
+++ b/core/src/processing/core/PSurface.java
@@ -158,4 +158,6 @@ public interface PSurface {
   public boolean stopThread();
 
   public boolean isStopped();
+
+  public boolean isFboAllowed();
 }
--- a/core/src/processing/core/PSurfaceNone.java
+++ b/core/src/processing/core/PSurfaceNone.java
@@ -245,6 +245,10 @@ public class PSurfaceNone implements PSurface {
     //g.setFrameRate(fps);
   }
 
+  @Override
+  public boolean isFboAllowed() {
+    return true;
+  }
 
   public class AnimationThread extends Thread {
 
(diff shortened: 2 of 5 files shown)


bill32bill32-dot:
void setup() 
{
  size(screen.width, screen.height);
}

void draw() 
{
  if (mousePressed) 
  {
    fill(0);
  } 
  else 
  {
    fill(255);
  }
  ellipse(mouseX, mouseY, 80, 80);
}
			[](url)

commonpike:
Without this, my mac powerbooks cant use P2D or P3D. I applied this patch and magically solved the issue without noticeable side effects.

Fwiw, here's a writeup of how I did that
https://stackoverflow.com/questions/57943895/processing-apply-core-patch-and-compile

ghjansen:
I am happy to know that it helped @commonpike . Thank you for sharing!
Please let me know if you notice anything wrong while using it.
I upvoted your solution in SO.

benfry:
We can't make use of `sun.java2d.*` classes because they're non-standard. If this is the only way to do it, these methods would need to be called via reflection, so that it can degrade gracefully if the `sun.java2d` package (or a specific class) is not available.

ghjansen:
@benfry understood. Thank you for your feedback!
I don't know yet a different way of doing this.
ce575e9 now uses reflection to handle `sun.java2d.*`.
Please, let me know if you think it can degrade gracefully enough this way.
Also, I'm available for hire. Is there any interest?

benfry:
Thanks for the fix, really appreciate it. I've incorporated a version of it over here: https://github.com/processing/processing4/commit/269f626f98e9a3edaf70ea67545d7456cbbf02b7 for Processing 4.x. 

@codeanticode will want to fix it up later since there's already FBO enabled/disabled code in there, but at least we have the workaround for now.

ghjansen:
@benfry nice!
@commonpike do you mind to test with the latest changes? (ce575e9)

PR #5900, opened 2019-08-01
Change method name 'sadness' to 'printError', 'loop' to 'enableLooping', 'list' to 'getFonts', 'available' to 'getNextActiveClients', 'paint' to 'setTextSelection', 'empty' to 'getSketchInstance' 

pdhung3012:
The class JAppleMenuBar is used as the starting point for the application. This method named 'sadness' is to print error. Thus, the method name 'printError' is more intuitive than 'sadness'.

The class PApplet is used to represent PApplet.  This method named 'loop' is to enable the looping mode. Thus, the method name 'enableLooping' is more intuitive than 'loop'.

The class PFont is used to represent the fonts in the system.  This method named 'list' is to get the list of the fonts installed on the system. Thus, the method name 'getFonts' is more intuitive than 'list'.

The class Server is used to represent Server.  This method named 'available' is to return the next client in line with a new message. Thus, the method name 'getNextActiveClient' is more intuitive than 'available'.

The class LineHighlight is used to Model/Controller for a highlighted source code line.  This method named 'paint' is to set text as selection. Thus, the method name 'setTextToSelection' is more intuitive than 'paint'.

The class PreprocessedSketch is used to represent PreprocessedSketch.  This method named 'empty' is to build a new PreprocessedSketch. Thus, the method name 'getSketchInstance' is more intuitive than 'empty'.

### DIFF
--- a/core/src/japplemenubar/JAppleMenuBar.java
+++ b/core/src/japplemenubar/JAppleMenuBar.java
@@ -51,19 +51,19 @@ public class JAppleMenuBar {
 	        instance = new JAppleMenuBar();
 
 	      } else {
-	        sadness("Problem saving " + FILENAME + " for full screen use.");
+	        printError("Problem saving " + FILENAME + " for full screen use.");
 	      }
 	    } else {
-        sadness("Could not load " + FILENAME + " from core.jar");
+        printError("Could not load " + FILENAME + " from core.jar");
 	    }
 	  } catch (IOException e) {
-	    sadness("Unknown error, here's the stack trace.");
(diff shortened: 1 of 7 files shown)


Bobby54:
I get what you mean, but the names you suggsted are much longer which is annoying when working with these methods.

benfry:
I'll post something later about method naming, but the short version is 1) keep things brief (as you see throughout the entire API, 2) keep things in line with the rest of the API… These suggestions contradict those rules that we've used (more or less) consistently.

PR #5944, opened 2019-10-27
Add new strings to Ukrainian translation

rzats:
Updates my 2016 Ukrainian translation file to add in some of the newer localization strings.
The various `keystroke` strings are copied to this file, despite being identical to the originals. Let me know if these should be removed outright!


benfry:
Thanks!

PR #5964, opened 2019-12-07
processing core code cleanup

Minimaximize:
General refactoring of the processing.core package files (focusing particularly on PApplet) with the primary aim of reducing ambiguity and tech debt:
- Replace C/C++ style array declarations with Java style in accordance with style guide
- Merge duplicate branches in code flow where appropriate
- Replace anonymous classes with lambda expressions where appropriate
- Remove unessesary boxing of values
- Remove redundant modifiers (eg: fields in interfaces are implicitly static and final)
- Remove stale comment blocks and commented-out code from PApplet (anything from 2015 or older).
I understand this goes against the style guide, but I think there's a point to be made about how much it impacts peoples ability to comprehend and engage with the project. (I've kept all of those changes to 1 commit, so it can easily be dropped if it's too mcuh)


benfry:
This is a lot more helpful if it's separate commits. The array stuff is an easy merge… But changing methods to/from `public` has a *huge* impact. 

And if you're admitting “I understand this goes against the style guide,” I'm not really sure what you expect me to do. It's not like I enjoy writing style guides, or being pedantic about code, but having put in the time to do so, why openly contradict it? It undermines the rest of the useful work and wastes time for me trying to wade through it.

Minimaximize:
@benfry Thank you for taking the time to review my changes - I will break them into smaller PRs (appologies, I'm used to tfs, which has an easier time cherry picking commits from Pull Requests)

PR #5993, opened 2020-02-11
Changed 2012-19 to 2012-20

Kirishikesan:
Changed the comments that mentioned the lifetime of processing foundation as 2012-19 to 2012-2020


Kirishikesan:
This is my first time contributing for processing foundation and this is my first proper pull request. I would like to be a fellow contributor and would like some guidance. Thank you.

PR #6013, opened 2020-03-25
Fix PGraphicsJava2D.copy() when src is of same class

thobbs:
Using `copy()` a `PGraphicsJava2D` with a src that is also `PGraphicsJava2D` will produce the following error:

`class sun.java2d.SunGraphics2D cannot be cast to class java.awt.Image`

This is due to `PGraphicsJava2D.native()` returning `g2` (a PGraphicsJava2D instance) rather than a `java.awt.Image` instance, which is what the cast here expects.

I've added specific handling for src being `PGraphicsJava2D`, as well as general handling of other non-Image results from `src.getNative()`. I built and tested this solution locally.


thobbs:
Friendly bump after one year - any chance this could be fixed?

PR #6051, opened 2020-05-28
Tint on FX2D now works properly, and improved JavaFX/Scenebuilder integration

SanielDan:
Now when drawing an image the cache trace all different tintColor for the image, 
in order to make the output consistent with the other renderers and avoid old calls of tint() to be overwritten.
I'v also changed the visibility of some members in order to make easier creating sketch that draw inside a standard JavaFX/Scenebuilder application.


12people:
As someone who would like to use Processing inside of a JavaFX project, this PR is very important to me, as it makes Processing's key JavaFX parts public.

@benfry Would you please consider merging it?

benfry:
I can take a look, but it needs to be in the active repo: https://github.com/processing/processing4 there's no 3.x development happening anymore.

benfry:
Ok, this can't be merged as-is: 
* Making all those fields public is a significant API change, and means that we have to support those public fields in the future. A better option is to do a simple subclass of the `FX2D` renderer that adds your own getters/setters for those fields.
* There are a lot of formatting changes in this PR that shouldn't be there. 
* As implemented, this tint cache would cause applications that use many tint colors to run out of memory very quickly. It stores every instance of the image and doesn't do any cleanup for unused colors, nor does it set a maximum number of cached objects.

If anyone is interested in picking up this work, `FX2D` development has moved here: https://github.com/processing/processing4-javafx

PR #6052, opened 2020-05-29
update issue template -- correct forum link

jeremydouglass:
changes link to not rely on subdomain autoforwarding

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,4 +1,4 @@
-<!--- ** For coding questions, please use https://forum.processing.org ** -->
+<!--- ** For coding questions, please use https://discourse.processing.org ** -->
 <!--- ** This form is only for bugs in the software & feature requests ** -->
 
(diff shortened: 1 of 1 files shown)


jeremydouglass:
The follow-on update would be to migrate to their new issue templates, which are now authored with a builder and stored as YAML in .github/ISSUE_TEMPLATE/

I would do it, but the recommended way to generate those files is to use their web wizard directly on the repo -- and I am not a repo member.

https://help.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates

benfry:
Thanks; the builder thing sounds more useful, though we'd do that over in the 4.x repo.

jeremydouglass:
> we'd do that over in the 4.x repo.

Great. If you would like me to set it up on that repo, I will just need to be a repo member at least temporarily in order to use their UI. Just let me know.

benfry:
Thanks, happy to do that—let me know where I need to set access for you, or feel free to reach out via email: fry at processing dot org.

PR #6081, opened 2020-08-03
loadJSONObject/Array() now return null if the given file was not found

fukuchi:
All data loading functions such as loadBytes() and loadXML() return null instead of throwing an exception when the given file was not found. On the other hand, loadJSONObject() and loadJSONArray() throws NullPointerException. This is not consistent.

This patch fixes this problem and gives the developers a consistent error handling way to other loading functions.


benfry:
Thanks; this is now incorporated for 4.0 alpha 3: https://github.com/processing/processing4/commit/2b19ff8d26f90626b5d9d3516d3e552bc575b49c

PR #6092, opened 2020-09-05
getShape() - Type of vertex was wrong for Cubic

hugovalepereira:
When 'detail' is set to 0, it was causing getShape to break with OTF fonts that use cubic bezier paths.
By mistake, a 3D quadratic was being drawn instead of a proper cubic bezier.


benfry:
Ha, how about that. Thanks for the fix.

PR #6132, opened 2020-12-05
Update PFont.java

dhowe:
This PR adds PFont.[textBounds](https://p5js.org/reference/#/p5.Font/textBounds), PFont.[textToPoints](https://p5js.org/reference/#/p5.Font/textToPoints) from p5.js, and PFont.glyphBounds, as shown in the sketch below (there may be more efficient ways to do these). Also, I attempted to use the suggested formatter in Eclipse, but not sure the output is totally correct -- happy to get feedback/improvements...

Screenshot 2020-12-05 at 9 48 09 PM

```

// DEMO SKETCH

import java.awt.geom.*;

String str = "Lorem ipsum.";
float x = 20, y = 250, fsize = 150;
PFont pf;

void setup() {
  size(1000, 400);
  background(240);

  pf = createFont("Georgia", 300);

  fill(0);
  textFont(pf);
  textSize(fsize);
  text(str, x, y);
  noFill();

  Rectangle2D.Float[] r = pf.glyphBounds(str, x, y, fsize);
  for (int i = 0; i < str.length(); i++) { 
    rect(r[i].x, r[i].y, r[i].width, r[i].height);
  }

  Rectangle2D.Float rb = pf.textBounds(str, x, y, fsize);
  rect(rb.x, rb.y, rb.width, rb.height);

  stroke(200,0,0);
  strokeWeight(3);
  
  PVector[] pts = pf.textToPoints(str, x, y, fsize, 0.5);
  for (int i = 0; i < pts.length; i++) {  
    point(pts[i].x, pts[i].y);
  }
}  
```

### DIFF
--- a/core/src/processing/core/PFont.java
+++ b/core/src/processing/core/PFont.java
@@ -961,7 +961,172 @@ public class PFont implements PConstants {
     return font;
   }
 
+  /**
+   * Returns an array of PVector with each representing a point on the path for
+   * the specified text
+   *
+   * @webref typography:loading_displaying
+   * @param text
+   *          a line of text
+   * @param x
+   *          the x-coordinate
+   * @param y
+   *          the y-coordinate
+   * @param textSize
+   *          the text size to use
+   * @param the
+   *          maximum distance that line segments used to approximate the curved
+   *          segments are allowed to deviate from points on the original curve
+   * @see_external java.awt.Shape
+   */
+  public PVector[] textToPoints(String str, float x, float y, float textSize) {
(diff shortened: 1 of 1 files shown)


micycle1:
Maybe worth opening a PR for this on Processing 4.

clankill3r:
Hi, I think this is really nice. I can see this being used a lot by students from our graphic design department.

Some feedback, I think the Rectangle2D class is super unfriendly for the majority of the processing users.
That is specifically this part: `Rectangle2D.Float[]`.
I think it would be the best to just return a float array.

PR #6156, opened 2021-01-27
Oracle JDK 8u281 compatibility (#6155)

### LINKED ISSUE #6155: Processing does not build on Oracle JDK 8u281
## Description

When using the latest JDK from Oracle, `ant build` fails to compile `PSurfaceFX` because of missing methods in `com.sun.glass.ui.Screen`. it builds fine on the previous release (8u271) under the exact same circumstances.

## Expected Behavior

Compilation should succeed.

## Current Behavior
Compilation fails with the [attached error output](https://github.com/processing/processing/files/5882577/processing.log).

## Steps to Reproduce

1. Retrieve the source code of Processing 3.5.4
2. Install or select Oracle JDK 8u281
3. Attempt to build Processing

## Your Environment

 

* Processing version: 3.5.4
* Operating System and OS version: NixOS 20.09 (a GNU/Linux distribution)
* Other information: I am a NixOS contributor and discovered this issue when trying to update [Oracle JDK on our end.](https://github.com/NixOS/nixpkgs/pull/110798)

## Possible Causes / Solutions

The most likely cause is the fact that between the two versions, the (internal use) interface of `com.sun.glass.ui.Screen` changed, removing the `getUIScale()` and `getRenderScale()` methods. This is evidenced in the `javafx-src.zip` file in both distributions.

There does not seem to be a patch in the master branch.

I might take a stab at trying to restore functionality with the new methods. I will also check processing4 because it might already contain a usable patch.

#5059 tackles the issue of version-bumping the jdk dependency. (because of the same class, funnily enough)
#5286 discusses the use of these methods in the long run.

lourkeur:
Backported from https://github.com/processing/processing4/commit/c14695476ed904fba445668c05384b0c5ad9bc24

- [x] compiles
- [x] basic hand testing

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -71,11 +71,11 @@
             value="../../processing-docs/content/examples" />
 
   <property name="jdk.version" value="8" />
-  <property name="jdk.update" value="202" />
-  <property name="jdk.build" value="08" />
+  <property name="jdk.update" value="281" />
+  <property name="jdk.build" value="09" />
   <!-- See https://gist.github.com/P7h/9741922
        or http://stackoverflow.com/q/10268583 -->
-  <property name="jdk.hash" value="1961070e4c9b4e26a04e7f5a083f551e" />
+  <property name="jdk.hash" value="89d678f2be164786b292527658ca1605" />
 
   <!-- these a*holes at Oracle... no longer works without a login
   <property name="jdk.update" value="212" />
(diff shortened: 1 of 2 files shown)


lourkeur:
oh I should probably bump the jdk minimum version

benfry:
Processing 3.5.4 uses that JDK version because of licensing changes by Oracle (we can't distribute it) after that release.

PR #6211, opened 2021-08-23
Fixed typos and grammar in French translation

SableRaf:


dzaima:
You should move this to https://github.com/processing/processing4, as that's where the current development is on.

PR #6247, opened 2022-03-23
 Add Collapsible Code Blocks #5622

Bigsby11:
Adding Collapsible code blocks would allow for easier management of code with less files, as well as neater code. Since the Processing IDE was made to target more beginners, this could be set to toggle


benfry:
This PR was already closed because it's a total mess—it has a ton of unrelated code and it's connected to a Processing release that's many years old. Please don't re-submit things like this.

PR #6295, opened 2024-12-17
Lock Threads fix

Stefterv:
Run at 6, not every minute of the 6th hour

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -2,7 +2,7 @@ name: 'Lock Threads'
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


SableRaf (changes requested):
Hey Stef, looks like there is a small mistake.

"At 6am every day" would be `cron: '0 6 * * *'`

We could also consider "At 6am on Sunday": `cron: '0 6 * * 0'`

Reference: https://crontab.guru/#*_6_*_*_0

Stefterv:
You're right, edited it too quick

SableRaf (approved):

